PR #38, opened 2019-02-12
Correct much part of code to PEP-8

Deloryn:
I tried to correct the code to be more in line with PEP-8. There is still some code to correct


pasuder:
I would like to ask you to rebase on latest master commit e3737f90d76f121b3c5f01111ddca2b0f99e8916.

Issue #15, opened 2019-02-02
Add `yq` choice to `traverse` prompts


PawelLipski:
That would perform the action and then immediately quit without suggesting further actions.

Issue #27, opened 2019-02-02
Get rid of global state in the implementation


PawelLipski:
As for now it should be enough to extract the following classes:
- `GitContext`, as a lazily evaluated cache of config, logs, reflogs, merge-bases etc.
- `MacheteContext`, for storing the branch dependency tree structure
- `CommandLineContext`, for options/command/args
- possibly `SystemContext` for checking/caching things like `isatty`, hook executability etc.

PawelLipski:
Partly addressed by @rifel123 on #151, still some globals left to go

PawelLipski:
Closing, practically no more global state left 🎉

PR #41, opened 2019-02-27
Add 'yq' option in traverse (#15)

PawelLipski:


git_machete/cmd.py
```diff
@@ -283,7 +283,7 @@ def up(b, prompt_if_inferred):
         u = infer_upstream(b)
         if u:
             if prompt_if_inferred:
-                if ask_if("Branch '%s' not found in the tree of branch dependencies; rebase onto the inferred upstream '%s'? [y/n] " % (b, u)):
```
pasuder, line 286: I would like to ask you to start limiting length to 80 chars. It is tricky to review on GitHub without vertical scrollbar which is on the bottom.

git_machete/cmd.py
```diff
@@ -901,61 +901,71 @@ def print_new_line(new_status):
...
                     del annotations[b]
                 save_definition_file()
+                if ans == 'yq':
+                    raise StopTraversal
                 # No need to flush caches since nothing changed in commit/branch structure (only machete-specific changes happened).
-                continue  # No need to sync branch `b` with remote since it just got removed from the tree of dependencies.
+                continue  # No need to sync branch 'b' with remote since it just got removed from the tree of dependencies.
             elif ans in ('q', 'quit'):
-                return
+                raise StopTraversal
```
pasuder: It seems to be a change of logic. Could you split it into separate commit, please?

pasuder (changes requested):
I could see here two changes:

- related to overall format and some renaming variables along with adding option `yq`
- change of `return` logic to `raise StopTraversal` - it is not mentioned in commit message - I would like to see it in separate commit.

It would be great to start using limit of 80 chars (or 120) to have possibility to review it without horizontal scrollbar usage.

git_machete/cmd.py
PawelLipski: Done, I limited all line lengths in `traverse` method.

git_machete/cmd.py
PawelLipski: I tidied up the raise/return logic in `traverse` and dependent functions, now return is used in `traverse` and `raise` in dependent functions, hope it's clearer now.

PawelLipski:
@pasuder fixes applied on a separate commit

git_machete/cmd.py
pasuder: Nice, thx!

git_machete/cmd.py
```diff
@@ -856,17 +856,123 @@ def __init__(self):
...
+            flush()
+        elif can_pick_other_remote and ans in ('o', 'other'):
+            pick_remote(b)
+        elif ans in ('q', 'quit'):
+            raise StopTraversal
+        return
+
+    new_s = get_relation_to_remote_counterpart(b, rb)
+    if new_s == IN_SYNC_WITH_REMOTE:
```
pasuder: Please decompose those if-elif into small functions.

pasuder:
Those if-elif-else aka _forest of Elves_ should be decomposed into functions and instead of if-elif, dict with names and associated functions should be used.

git_machete/cmd.py
PawelLipski: Done

PawelLipski:
@pasuder: refactored `handle_untracked_branch` in a similar manner to what you suggested

git_machete/cmd.py
```diff
@@ -856,17 +856,123 @@ def __init__(self):
         pass
 
 
+def flush():
+    global branch_defs_by_sha_in_reflog, initial_log_shas_cached, remaining_log_shas_cached, remote_tracking_branches_cached, sha_by_revision_cached
```
pasuder, line 860: This should be handled in a bit different way. It looks very bash'y way.

git_machete/cmd.py
```diff
@@ -856,17 +856,123 @@ def __init__(self):
...
+            if ans == 'yq':
+                raise StopTraversal
+            flush()
+        elif can_pick_other_remote and ans in ('o', 'other'):
+            pick_remote(b)
+        elif ans in ('q', 'quit'):
+            raise StopTraversal
+        return
+
+    message = {
```
pasuder, line 904: Nice!

pasuder (dismissed):
Looks good, one comment for future task about those globals.

git_machete/cmd.py
PawelLipski: Yes, I'm aware of how much mess with globals I made in this script :D there's a task to tidy up all that (#27)

Issue #35, opened 2019-02-05
Bump Python to 3.x


PawelLipski:
`DEPRECATION: Python 2.7 will reach the end of its life on January 1st, 2020. Please upgrade your Python as Python 2.7 won't be maintained after that date. A future version of pip will drop support for Python 2.7.`

PawelLipski:
Also remember to update setup.cfg

PawelLipski:
Finally decided to make compatible with both Python 2.7 and 3.

PR #50, opened 2019-05-07
Make codebase compatible with both Python 2.7 and 3 (#35)

PawelLipski:


.travis.yml
```diff
@@ -1,11 +1,9 @@
 language: python
 matrix:
   include:
-    - python: 2.7
-      env: TOXENV=py27
     - python: 3.6
       env: TOXENV=py36
-    - python: 2.7  # TODO: bump to 3.6
```
pasuder, line 8: What happened here? I found it in commit 2fc9aba575af4911c11d3948b6627f9bcea6ca8a and got it :smirk:

.travis.yml
```diff
@@ -1,11 +1,9 @@
...
+    - python: 3.6
```
pasuder, line 10: IMO not needed any more after that change: https://changelog.travis-ci.com/the-default-python-version-for-your-builds-is-now-3-6-97935

Same for L4.

RELEASE_NOTES.md
```diff
@@ -1,5 +1,9 @@
 # Release notes
 
+## New in git-machete 2.9.1
```
pasuder: IMO that change is a bit break in compatibility and should be reflected in at least minor number: `2.10.0` or even in major number: `3.0.0`.

git_machete/cmd.py
```diff
@@ -44,31 +44,35 @@ def underline(s):
     return UNDERLINE + s + ENDC
 
 
+def star(f):  # tuple unpacking in lambdas
```
pasuder, line 47: Nice trick :+1:

pasuder (changes requested):
IMO change of version should touch minor or major version number due to change of python code base. It is a break in compatibility: https://apr.apache.org/versioning.html

RELEASE_NOTES.md
PawelLipski: Actually... I now tried to ensure that it works both under Python 2.7 and Python 3, pls take a look :) Thus, it now merits at most minor version change.

.travis.yml
PawelLipski: Pls take a look at the updated matrix (that still takes 2.7 into account)

.travis.yml
PawelLipski: Obsolete now

git_machete/cmd.py
PawelLipski: Yeah, that's what stackoverflow adviced to emulate tuple unpacking in lambdas under Python 3... since this feature is gone :(

PawelLipski:
@pasuder comments addressed, I tried to make codebase both Python 2 and 3 compatible... at least until next major relase.

.travis.yml
pasuder: `pep8` run under python2.7 and python3.6. I do not know if it is needed, but maybe there are different rules for those versions. It seems to be okey.

pasuder (approved):

.travis.yml
PawelLipski: I put the entire cartesian product of possible envs/versions just to be sure nothing is missed :)

PR #53, opened 2019-05-19
Make the ZSH completion file autoloadable

ilai-deutel:
This simple change in the ZSH completion file allows it to be automatically loaded as long as it is in the `fpath` (for instance by moving it to `/usr/share/zsh/site-functions/`), see the [ZSH doc](http://zsh.sourceforge.net/Doc/Release/Completion-System.html#Autoloaded-files).

This should allow package managers to install the completion without user interaction.

### DIFF
--- a/completion/git-machete.completion.zsh
+++ b/completion/git-machete.completion.zsh
@@ -1,4 +1,4 @@
-#!/usr/bin/env zsh
+#compdef git-machete
 
 _git-machete() {
     local ret=1


PawelLipski (dismissed):

PawelLipski:
@ilai-deutel could you also provide a relevant change to https://github.com/VirtusLab/git-machete/wiki (pls lmk if you don't have edit permissions there)? Not sure if the `oh-my-zsh` should better be rewritten or extended now...

ilai-deutel:
I am not familiar with the local installation of ZSH / OMZSH completions (as opposed to the global installation with a package manager), but the current wiki page looks ok to me (it is similar to [this project](https://github.com/zsh-users/zsh-completions) for instance).

PawelLipski:
@ilai-deutel Could you rebase onto the recent develop?

ilai-deutel:
Done

PawelLipski (approved):

PR #57, opened 2019-08-30
Implement displaying absolute path by `file` command

tdudzik:
Before and after:
WindowsTerminal_LdIalHYfsJ

### DIFF
--- a/git_machete/cmd.py
+++ b/git_machete/cmd.py
@@ -9,7 +9,7 @@ import subprocess
 import sys
 import textwrap
 
-VERSION = '2.10.0'
+VERSION = '2.10.1'
 
 
 # Core utils
@@ -411,17 +411,18 @@ def get_root_dir():
     return root_dir
 
 
-git_dir = None
+abs_git_dir = None
 
 
-def get_git_dir():
-    global git_dir
-    if not git_dir:
+def get_abs_git_dir():
+    global abs_git_dir
+    if not abs_git_dir:
         try:
             git_dir = popen_git("rev-parse", "--git-dir").strip()
+            abs_git_dir = os.path.abspath(git_dir)
         except MacheteException:
             raise MacheteException("Not a git repository")
-    return git_dir
+    return abs_git_dir
 
 
 config_cached = None
@@ -612,7 +613,7 @@ def go(branch):
 
 
 def get_hook_path(hook_name):
-    hook_dir = get_config_or_none("core.hooksPath") or os.path.join(get_git_dir(), "hooks")
+    hook_dir = get_config_or_none("core.hooksPath") or os.path.join(get_abs_git_dir(), "hooks")
     return os.path.join(hook_dir, hook_name)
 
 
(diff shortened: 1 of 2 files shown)


git_machete/cmd.py
```diff
@@ -424,6 +425,16 @@ def get_git_dir():
     return git_dir
 
 
+def get_abs_git_dir():
+    global abs_git_dir
+    if not abs_git_dir:
+        try:
+            abs_git_dir = popen_git("rev-parse", "--absolute-git-dir").strip()
```
PawelLipski: What is the minimum version of git that supports this flag?...

git_machete/cmd.py
```diff
@@ -424,6 +425,16 @@ def get_git_dir():
     return git_dir
 
 
+def get_abs_git_dir():
```
PawelLipski, line 417: Mostl likely `get_git_dir` in the current form won't be necessary anymore now that we have this function

git_machete/cmd.py
tdudzik: It seems that minimum version is `2.13.0`. I will try to implement it differently as we want to support `2.7` version as well.

git_machete/cmd.py
tdudzik: It turned out to be quite useful in a new implementation.

git_machete/cmd.py
```diff
@@ -424,6 +425,20 @@ def get_git_dir():
...
+    global abs_git_dir
+    if not abs_git_dir:
+        try:
+            # git_dir is absolute when we are in a directory of submodule
+            if os.path.isabs(get_git_dir()):
+                abs_git_dir = git_dir
+            else:
+                abs_git_dir = os.path.join(get_root_dir(), get_git_dir())
```
PawelLipski: Hmmm what about `os.path.abspath` (btw pls see if it works under both Python 2.7 and Python 3.6)? It doesn't depend on possibly undocumented assumptions as the current solution does.

git_machete/cmd.py
```diff
@@ -424,6 +425,20 @@ def get_git_dir():
     return git_dir
 
 
+def get_abs_git_dir():
+    global abs_git_dir
+    if not abs_git_dir:
+        try:
+            # git_dir is absolute when we are in a directory of submodule
```
PawelLipski: I'm afraid we can't really rely on this behavior of git... I'm afraid this can silently change even b/w minor versions in future :/

git_machete/cmd.py
```diff
@@ -1902,7 +1917,8 @@ def parse_direction(b, down_pick_mode):
         elif cmd == "file":
             expect_no_param(parse_options(args))
             # No need to read definition file.
-            print(definition_file)
+            abs_definition_file = os.path.abspath(os.path.join(get_abs_git_dir(), os.pardir, definition_file))
```
PawelLipski: Yeah, you call `abspath` here... I can't see a reason why we just can't use `abspath` in `get_abs_git_dir` instead

PawelLipski: Also, I'd rather store the `definition_file` path as absolute from the very beginning... so line 1866 would look like: `definition_file = os.path.join(get_abs_git_dir(), "machete")`

git_machete/cmd.py
PawelLipski: Pls reconsider given the recent comments :)

PawelLipski (changes requested):

PawelLipski (approved):

PR #58, opened 2019-09-05
Add --color flag to `status`

tdudzik:


RELEASE_NOTES.md
```diff
@@ -1,5 +1,8 @@
 # Release notes
 
+## New in git-machete 2.11.0
+- added: --color flag to `status`
```
PawelLipski: Nit: `--color` in backticks

git_machete/cmd.py
```diff
@@ -1684,6 +1700,7 @@ def usage(c=None):
 
             Options:
             -l, --list-commits            Additionally lists the messages of commits introduced on each branch.
+            --color[=WHEN]                Colorize the output; WHEN can be 'always' (default if omitted), 'auto', or 'never';
```
PawelLipski: I'd say `auto` should be the default

git_machete/cmd.py
```diff
@@ -1772,13 +1789,15 @@ def version():
...
     def parse_options(in_args, short_opts="", long_opts=[], gnu=True):
-        global opt_debug, opt_down_fork_point, opt_fork_point, opt_list_commits, opt_onto, opt_roots, opt_stat, opt_verbose
+        global opt_color, opt_debug, opt_down_fork_point, opt_fork_point, opt_list_commits, opt_onto, opt_roots, opt_stat, opt_verbose
 
         fun = getopt.gnu_getopt if gnu else getopt.getopt
         opts, rest = fun(in_args, short_opts + "hv", long_opts + ['debug', 'help', 'verbose', 'version'])
 
         for opt, arg in opts:
-            if opt in ("-d", "--down-fork-point"):
+            if opt == "--color":
```
PawelLipski, line 1818: Pls update bash&zsh completion scripts in `completion/`

git_machete/cmd.py
```diff
@@ -1979,7 +1999,12 @@ def parse_direction(b, down_pick_mode):
             read_definition_file()
             slide_out(params or [current_branch()])
         elif cmd in ("s", "status"):
-            expect_no_param(parse_options(args, "l", ["list-commits"]))
+            expect_no_param(parse_options(args, "l", ["list-commits", "color="]))
+            if opt_color not in ("always", "auto", "never"):
+                raise MacheteException("Invalid argument for: --color. Valid arguments: always, auto, never.")
```
PawelLipski: Keep the format of the message more similar to what's for `list` command above

PawelLipski: Esp. wrt. listing possible values for the arg

git_machete/cmd.py
```diff
@@ -1979,7 +1999,12 @@ def parse_direction(b, down_pick_mode):
...
+            else:
+                global ascii_only
+                ascii_only = False if opt_color == "always" or (opt_color == "auto" and sys.stdout.isatty()) else True
```
PawelLipski: Or just `ascii_only = opt_color == "always" or (opt_color == "auto" and sys.stdout.isatty())`

git_machete/cmd.py
```diff
@@ -1345,11 +1359,13 @@ def print_line_prefix(b_, suffix):
...
                         # fp_branches_cached will already be there thanks to the above call to 'fp_sha'.
                         fp_branches_formatted = " and ".join(map(star(lambda lb, lb_or_rb: lb_or_rb), fp_branches_cached[b]))
-                        fp_suffix = (RED + ' ➔ fork point ???' + ENDC + ' commit ' + short_sha(fp_sha(b)) + ' found in reflog of ' + fp_branches_formatted)
+                        fp_suffix = (colored(" " + right_arrow() + " fork point ???", RED) + " commit " + short_sha(fp_sha(b)) + " found in reflog of " + fp_branches_formatted)
                     else:
                         fp_suffix = ''
-                    print_line_prefix(b, u"│ " + ENDC + dim(msg) + fp_suffix + "\n")
-            print_line_prefix(b, u"└─")
+                    print_line_prefix(b, vertical_bar() + " " + ENDC + dim(msg) + fp_suffix + "\n")
+            node_ascii_only = {DIM: "|-", RED: "x-", GREEN: "o-", YELLOW: "?-"}
```
PawelLipski: What about hmmm... `m-` for DIM? Since it signifies *m*erged branch

PawelLipski: Also, mention (briefly) the ASCII equivalents in help for `status`

PawelLipski (changes requested):

git_machete/cmd.py
PawelLipski: Aaaah ok it's a bit confusing:
* `auto` is the "default" default if `--color` isn't specified BUT
* `always` is the default value for `--color` flag
Pls think how to clarify that... without writing a whole essay here ;)

git_machete/cmd.py
```diff
@@ -1345,11 +1359,13 @@ def print_line_prefix(b_, suffix):
...
+            node = u"└─" if not ascii_only else node_ascii_only[edge_color[b]]
```
PawelLipski: Nit: I'd call it e.g. `elbow`... `node` generally has many meanings in CS, can be confusing here a bit IMO

git_machete/cmd.py
```diff
@@ -32,17 +32,31 @@ def __str__(self):
...
-    return UNDERLINE + s + ENDC
+    return UNDERLINE + s + ENDC if not ascii_only else s + " *"
+
+
+def colored(s, color):
+    return color + s + ENDC if not ascii_only else s
+
 
+def vertical_bar():
+    return u"│" if not ascii_only else "|"
```
PawelLipski, line 55: Pls check if all this stuff (esp. wrt. unicode characters) works as expected under both Python 2.7 and 3.6

git_machete/cmd.py
PawelLipski: Aaaah ok it's a bit confusing:

* `auto` is the "default" default if` --color` isn't specified BUT
* `always` is the default value for `--color` flag

Pls think how to clarify that... without writing a whole essay here ;)

git_machete/cmd.py
```diff
@@ -1979,7 +1999,12 @@ def parse_direction(b, down_pick_mode):
             read_definition_file()
             slide_out(params or [current_branch()])
         elif cmd in ("s", "status"):
-            expect_no_param(parse_options(args, "l", ["list-commits"]))
+            expect_no_param(parse_options(args, "l", ["list-commits", "color="]))
+            if opt_color not in ("always", "auto", "never"):
```
PawelLipski, line 2023: Is it now consistent with the help for `status` which says that value for `--color` is optional?

git_machete/cmd.py
PawelLipski: And `elbow_ascii_only` above

git_machete/cmd.py
```diff
@@ -32,17 +32,31 @@ def __str__(self):
...
-    return DIM + s + ENDC
+    return DIM + s + ENDC if not ascii_only else s
 
 
 def underline(s):
-    return UNDERLINE + s + ENDC
+    return UNDERLINE + s + ENDC if not ascii_only else s + " *"
+
+
+def colored(s, color):
```
PawelLipski, line 50: I think there's still at least one place where a color (`YELLOW` specifically) is used in the old concatenated manner (hook executability check)... pls do a `grep` or whatnot across the file for other possible instances

git_machete/cmd.py
PawelLipski: Btw consider making the argument non-optional as a workaround, for the sake of clarity

completion/git-machete.completion.bash
```diff
@@ -12,7 +12,7 @@ _git_machete() {
     discover_opts="-l --list-commits -r --roots="
     reapply_opts="-f --fork-point="
     slide_out_opts="-d --down-fork-point="
-    status_opts="-l --list-commits"
+    status_opts="-l --list-commits --color="
```
PawelLipski, line 16: Also complete the possible values ;)

completion/git-machete.completion.zsh
```diff
@@ -62,6 +62,7 @@ _git-machete() {
                     ;;
                 (status|traverse)
                     _arguments \
+                        '(--color)'{--color}'[Colorize the output]' \
```
PawelLipski: Same, complete the possible values

git_machete/cmd.py
```diff
@@ -1678,12 +1695,25 @@ def usage(c=None):
             * displays the output of 'machete-status-branch' hook (see help on 'hooks'), if present;
             * optionally lists commits introduced on each branch if '-l'/'--list-commits' is supplied.
 
+            Using colors can be disabled with a '--color' flag set to 'never'. With '--color=always' git machete always emits colors and with '--color=auto' it emits colors only when standard output
```
PawelLipski, line 1710: I'd move it after the following paragraph (`Note: in practice, ...`)

git_machete/cmd.py
```diff
@@ -1678,12 +1695,25 @@ def usage(c=None):
...
+            |
+            m-<branch4> # dim
+
             Note: in practice, both yellow and red edges suggest that the downstream branch should be updated against its upstream.
             Yellow typically indicates that there are/were commits from some other branches on the path between upstream and downstream and that a closer look at the log of the downstream branch might be necessary.
             Grey edge suggests that the downstream branch can be slid out.
 
             Options:
             -l, --list-commits            Additionally lists the messages of commits introduced on each branch.
+            --color[=WHEN]                Colorize the output; WHEN can be 'always', 'auto', or 'never'.
```
PawelLipski: `[=WHEN]` suggests the arg is optional... but the code below indicates it's actually required

PawelLipski (changes requested):

completion/git-machete.completion.zsh
```diff
@@ -60,7 +60,13 @@ _git-machete() {
                         '(-d --down-fork-point)'{-d,--down-fork-point=}'[Fork point commit after which the rebased part of history of the downstream branch is meant to start]: :__git_references' \
                     && ret=0
                     ;;
-                (status|traverse)
+                (status)
+                    _arguments \
+                        '(--color)'--color='[Colorize the output; WHEN can be "always", "auto", or "never".]: :_git_machete_color_modes' \
```
PawelLipski: `argument can be` instead of `WHEN`

PawelLipski (approved):

Issue #72, opened 2020-01-13
Proper handling of operations such as renaming branch just after it is created


micpiotrowski:
# Overview

We have some branch (say master). Then from that point we create new branch (say branch2) and commit on this branch. Next create another branch (say branch3) from branch2 and next we rename the branch3 to something else. Next we create machete file corresponding to our branches and run `git machete status -l`

## Expected behavior/subcommand output

The `git machete status -l` command should print our branches tree with all green edges.

## Actual behavior/subcommand output

The `git machete status -l` command prints branches tree where edge between master and branch2 is yellow and near commit on branch2 we have message like `➔ fork point ??? commit <commit-hash> found in reflog of branch3`. Edge between branch2 and branch3 is green.

## `git` and `git machete` versions

git --version `git version 2.24.1`

git machete --version `git-machete version 2.12.5`

PawelLipski:
Done as of [v2.13.0](https://github.com/VirtusLab/git-machete/releases/tag/v2.13.0)

PR #73, opened 2020-01-13
Sift out all reflog entries with the same commit hash as an entry excluded due to its subject (#72)

PawelLipski:
Thanks to @micpiotrowski for the idea of the solution


PawelLipski:
Done as of v2.13.0

PR #91, opened 2020-06-05
Handle the case of current directory becoming non-existent

PawelLipski:
Kudoz to @radeusgd for reporting this bug!

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -6,6 +6,7 @@
 - added: `addable` category of `list`
 - added: `advance` command
 - fixed: Bash completion for `-y` option in several commands
+- fixed: handling the case of current directory becoming non-existent (e.g. as a result of checkout)
 
 ## New in git-machete 2.14.0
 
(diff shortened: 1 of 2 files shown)


mkondratek (approved):
lgtm

radeusgd:
Cool that you fixed this issue!

One suggestion - the directory may disappear only during checkout and then reappear when checking out back to the original branch - in this case the current solution is enough.

In the rarer case in which `git machete traverse` rebased current branch onto master and new commits on master happened to remove the directory that we were in (very rare situation, but can happen) - the users shell will be left in the dangling directory after calling the command (because since it was permanently deleted, it won't reappear).

AFAIK it's not possible to cd parent process, but maybe it would be a good idea to at least check again at the very end if the original directory exists and if it doesn't, issue another warning that the parent shell is likely in a dangling directory (plus possibly a hint where the user could cd) to fix that. This would be helpful so as not to confuse users that are left in a dangling directory. (Although maybe it's an overkill since you already emit a warning at the top and this looks like an extremely rare situation)

PawelLipski:
@radeusgd done as suggested on `1st round of fixes`, that's likely the optimal UX indeed :)

image

PawelLipski:
Merged to develop.

PR #101, opened 2020-08-13
Suggest sliding out non-existent branches in `status` and other contexts

### LINKED ISSUE #100: Consider allowing to automatically remove missing branches from the machete file
# Overview

When a branch is removed manually (for example with `git branch -d <name>`), but is still present in the machete tree, I get an error when running status for example.

While I guess it is recommended to use `git machete slide-out` for this usecase, it would be good if a normal git command did not break the machete.

It is of course a very low priority thing, because it can easily be solved by editing the machete file, as indicated in the error, but it could be much more useful if the tool asked to remove the branch automatically.

## Expected behavior/subcommand output

```
# git branch -d <name>
# git machete status -L
<name> is not a local branch, perhaps it was deleted.
Do you want machete to remove it from the configuration file?
(You can also edit the definition file manually with git machete edit)
[y/n] y
<name> has been deleted from the tree.
<shows the requested status>
```

## Actual behavior/subcommand output

```
# git branch -d <name>
# git machete status -L
.git/machete, line 2: <name> is not a local branch. Edit the definition file manually with git machete edit
```

```
# git --version
git version 2.17.1

# git machete --version
git-machete version 2.15.3

```

PawelLipski:
As suggested by @radeusgd on #100

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,5 +1,9 @@
 # Release notes
 
+## New in git-machete 2.15.4
+
+- improved: slide-out is suggested in certain contexts (like `status`) in case a non-existent branch is found in .git/machete
+
 ## New in git-machete 2.15.3
 
 - changed: skip verification of managed branches for `anno` and `show`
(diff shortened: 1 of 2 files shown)


mkondratek (approved):
lgtm

micpiotrowski (approved):

PR #102, opened 2020-08-20
Improve README.md

PawelLipski:


README.md
```diff
@@ -6,7 +6,7 @@
 [![Snap](https://snapcraft.io/git-machete/badge.svg)](https://snapcraft.io/git-machete)
 [![Snap build status](https://build.snapcraft.io/badge/VirtusLab/git-machete.svg)](https://build.snapcraft.io/user/VirtusLab/git-machete)
 
-![](https://raw.githubusercontent.com/VirtusLab/git-machete/master/logo.png)
+<img src="https://raw.githubusercontent.com/VirtusLab/git-machete/master/logo.png" width="180" height="180" align="left" />
```
micpiotrowski: Logo can be a little bit bigger to the text do not wrap under it (probably 200x200 will be enough)

README.md
```diff
@@ -30,64 +30,78 @@ We suggest a couple of alternative ways of installation.
```
micpiotrowski, line 41: GIF on line 19 can be centred

micpiotrowski: And information about port into IntelliJ plugin should be reworded to something like `A port to an IntelliJ plugin is available on [VirtusLab/git-machete-intellij-plugin](http://github.com/VirtusLab/git-machete-intellij-plugin/) repository and is constantly being developed`

README.md
```diff
@@ -30,64 +30,78 @@ We suggest a couple of alternative ways of installation.
 
 git-machete works under both Python 2.7 and Python 3.x.
 
-### Using Homebrew (macOS)
+<details><summary><b>Using Homebrew (macOS)</b></summary><br/>
```
micpiotrowski: Theoretically, spoilers can be replaced by headers because there is a little content inside it, but IMHO spoilers are also OK

mkondratek (commented):
image

A not-centered gif is unjustified. :xd: Probably the actual problem (solution) is that only full-width-images look good within the readmes.

README.md
mkondratek: Looking at the https://github.com/isocpp/CppCoreGuidelines as a reference...

It would be nice to have a full-width-logo for our readmes/docs. Centering is rather a temporary solution.

README.md
mkondratek: I do not like spoilers. That may be only my personal experience but for me, both look and feel are bad. I am used to non-interactive readmes. `<details> `tags may be good for logs and other content that we might want hide but not for readmes.

Have you ever seen a github readme with spoilers? :thinking:

README.md
mkondratek: \+ add a fragment (`#cośtam`) to redirect directly to plugin's readme

PawelLipski: All done

mkondratek (approved):
cool :1st_place_medal:

micpiotrowski (approved):

PR #107, opened 2021-02-09
Drop python 2, remove deprecated commands, use string literals

### LINKED ISSUE #26: Remove deprecated commands
Involves bumping the major version.

- `infer` (now replaced with `discover`)
- `prune-branches` (now replaced with `delete-unmanaged`)
- `format` (now only `help format`, standalone command doesn't make sense)

### LINKED ISSUE #76: Drop support for Python 2
This will require a major release (git-machete 3.0); we can also piggyback removal of deprecated commands onto that release (`prune-branches`, `infer`).

mkondratek:


.travis.yml
```diff
@@ -19,8 +19,8 @@ script: bash ci/tox/travis-script.sh
 jobs:
   include:
     - stage: Tests
-      env: PYTHON_VERSION=2.7 GIT_VERSION=1.7.10  # Earliest version of git supported by git-machete
-    - env: PYTHON_VERSION=2.7 GIT_VERSION=2.7.6
+      env: PYTHON_VERSION=3.6 GIT_VERSION=1.7.10  # Earliest version of git supported by git-machete
+    - env: PYTHON_VERSION=3.6 GIT_VERSION=2.7.6
     - env: PYTHON_VERSION=3.6 GIT_VERSION=2.20.2
```
PawelLipski: Probably this one (`3.6`+`2.20.2`) is now redundant... TBH I've only had two cases so far:
* either the earliest version fails on some obscure feature missing (like unknown %-format specifier in `git log --format`)
OR
* well, everything fails

git_machete/cmd.py
```diff
@@ -3126,14 +3113,18 @@ def short_usage():
...
+    print(f"git-machete version {__version__}")
 
 
 def main():
     launch(sys.argv[1:])
 
 
 def launch(orig_args):
+    if sys.version_info[0] == 2:
+        sys.stderr.write("Python 2.x is no longer supported. Please switch to Python 3.\n")
```
PawelLipski: I'd also check that in case of Python 3, minor version is at least 3.6

tox.ini
```diff
@@ -1,5 +1,5 @@
 [tox]
-envlist = pep8,py27,py35,py36,py37,py38,py39
+envlist = pep8,py35,py36,py37,py38,py39
```
PawelLipski: Hmm `py35` no longer supported I guess? 🤔

git_machete/cmd.py
```diff
@@ -228,7 +221,7 @@ def run_cmd(cmd, *args, **kwargs):
     mark_current_directory_as_possibly_non_existent()
 
     if opt_debug and exit_code != 0:
-        sys.stderr.write(dim("<exit code: %i>\n\n" % exit_code))
+        sys.stderr.write(dim(f"<exit code: {exit_code:d}>\n\n"))
```
PawelLipski: Actually... what happens if you skip the `:d` for number-typed values? does it make any difference?

PawelLipski (approved):

git_machete/cmd.py
mkondratek: no diff. redundant, dropped

PR #108, opened 2021-02-11
Use mypy

### LINKED ISSUE #68: Wire up type checking (mypy)
https://mypy.readthedocs.io/en/latest/python2.html

mkondratek:


git_machete/cmd.py
```diff
@@ -1,5 +1,6 @@
 #!/usr/bin/env python
 # -*- coding: utf-8 -*-
+from typing import Any, AnyStr, Callable, Dict, Generator, Iterable, Iterator, List, Match, Optional, Set, Tuple, TypeVar
```
PawelLipski: Whoops that's failing on pep8... but you probably already know

PawelLipski: Oh okay there some other errors so I'm not gonna go through that file in detail yet

tox.ini
```diff
@@ -13,7 +13,10 @@ install_command =
 deps =
   -r{toxinidir}/requirements.txt
   -r{toxinidir}/test-requirements.txt
-commands = stestr run {posargs}
+  mypy
+commands =
+  stestr run {posargs}
+  mypy --config-file mypy.ini git_machete/cmd.py
```
PawelLipski: Hmmm maybe `mypy` before `stestr`? usually you first want the feedback on failed typing, and only then if typing is correct &mdash; on failed tests

mypy.ini
```diff
@@ -0,0 +1,15 @@
...
+disallow_untyped_calls = True
+disallow_untyped_decorators = True
+no_implicit_optional = True
+warn_return_any = True
+warn_unused_configs = True
+disallow_subclassing_any = True
+disallow_any_generics = True
+check_untyped_defs = True
+warn_redundant_casts = True
+warn_unused_ignores = False
```
PawelLipski: Nit: I guess this can be set to `True` eventually?

tox.ini
```diff
@@ -13,7 +13,10 @@ install_command =
 deps =
   -r{toxinidir}/requirements.txt
   -r{toxinidir}/test-requirements.txt
-commands = stestr run {posargs}
+  mypy
```
PawelLipski: Hmm better add to `test-requirements.txt` for consistency

tox.ini
```diff
@@ -13,7 +13,9 @@ install_command =
 deps =
   -r{toxinidir}/requirements.txt
   -r{toxinidir}/test-requirements.txt
-commands = stestr run {posargs}
+commands =
+  mypy --config-file mypy.ini git_machete/cmd.py
```
PawelLipski: Actually... can you already make mypy check the entire package/directory, instead of just a single file? 🤔

PawelLipski:
Whoops remember the `install gcc something something` command that was removed from Dockerfile on the previous PR? It seems that now it needs to be restored (without the surrounding `if`, though)

ci/tox/Dockerfile
```diff
@@ -5,7 +5,7 @@ RUN ln -s /usr/local/bin/python /usr/bin/python
 
 ARG git_version
 RUN set -x \
-    && apk add --no-cache --virtual git-build-deps  alpine-sdk autoconf gettext wget zlib-dev \
```
PawelLipski, line 8: Actually, `git-build-deps` is an arbitrary name of package group (for easier removal in `apk del git-build-deps`) and NOT an actual Alpine package

ci/tox/Dockerfile
```diff
@@ -5,7 +5,7 @@ RUN ln -s /usr/local/bin/python /usr/bin/python
...
+    && apk add --no-cache --virtual git-build-deps alpine-sdk autoconf gettext wget zlib-dev \
```
PawelLipski: Maybe:
```suggestion
    && apk add --no-cache --virtual=git-build-deps alpine-sdk autoconf gettext wget zlib-dev \
```
if it works

ci/tox/Dockerfile
```diff
@@ -25,6 +25,7 @@ RUN set -x \
 ARG python_version
 ENV PYTHON_VERSION=${python_version}
 ENV PYTHON=python${python_version}
+RUN apk add --no-cache gcc
```
PawelLipski: Pls add a comment why this is needed (we'll easily forget it over time)

git_machete/tests/functional/test_machete.py
```diff
@@ -193,25 +193,25 @@ def adapt(s):
...
 
     @staticmethod
-    def launch_command(*args):
+    def launch_command(*args: str) -> str:
         orig_out = sys.stdout
         out = io.StringIO()
         sys.stdout = out
         try:
-            cmd.launch(args)
+            cmd.launch(args)  # type: ignore
```
PawelLipski: Hmm why `ignore` needed here? not that it matters much since it's test code anyway

PawelLipski:
`musl-dev` needed (https://stackoverflow.com/a/30873179/2750812)

Also, you can run the Dockerized build locally, see ci/tox/README.md

git_machete/tests/functional/test_machete.py
mkondratek: given tuple, required list

ci/tox/Dockerfile
```diff
@@ -25,6 +25,7 @@ RUN set -x \
 ARG python_version
 ENV PYTHON_VERSION=${python_version}
 ENV PYTHON=python${python_version}
+RUN apk add --no-cache gcc musl-dev # both packages are required by tox
```
PawelLipski: It's required to install `mypy` right? not for `tox` itself to run (since `tox` is just a tool like `Make` that only installs&runs stuff but doesn't contain much logic on its own)

git_machete/tests/functional/test_machete.py
```diff
@@ -12,65 +12,65 @@
...
             opt = args[1]
-            self.execute('git init %s' % opt)
+            self.execute(f'git init {opt}')
         else:
             self.execute('git init')
         return self
 
-    def new_branch(self, branch_name):
-        self.execute('git checkout -q -b %s' % branch_name)
+    def new_branch(self, branch_name: str) -> 'SandboxSetup':
```
PawelLipski, line 34: Whoops so for class return types... quotes be should used? 😯 as in, you can't write `-> SandboxSetup`?

git_machete/tests/functional/test_machete.py
```diff
@@ -12,65 +12,65 @@
...
         self.file_dir = os.path.dirname(os.path.abspath(__file__))
         self.remote_path = os.popen('mktemp -d').read().strip()
         self.sandbox_path = os.popen('mktemp -d').read().strip()
 
     @staticmethod
-    def execute(command):
+    def execute(command: str) -> None:
         result = os.system(command)
-        assert result == 0, "%s returned %i" % (command, result)
+        assert result == 0, f"{command} returned {result:d}"
```
PawelLipski: `:d` redundant?

git_machete/tests/functional/test_machete.py
mkondratek: yup

mkondratek: https://stackoverflow.com/questions/15853469/putting-current-class-as-return-type-annotation

git_machete/cmd.py
```diff
@@ -3108,27 +3171,30 @@ def usage(c=None):
...
-def launch(orig_args):
+definition_file_path: str = ""
+
+
+def launch(orig_args: List[str]) -> None:
     if sys.version_info.major == 2:
         sys.stderr.write("Python 2.x is no longer supported. Please switch to Python 3.6 or higher.\n")
         sys.exit(1)
     elif sys.version_info.major == 3 and sys.version_info.minor < 6:
         sys.stderr.write("Python 3.x below 3.6 is no longer supported. Please switch to Python 3.6 or higher.\n")
```
PawelLipski: Unrel: maybe instead of `3.x below 3.6` simply include the actual version from `sys.version_info` for clarity

git_machete/tests/functional/test_machete.py
mkondratek: can be resolved with tihs `from __future__ import annotations` too

git_machete/cmd.py
```diff
@@ -3108,27 +3171,30 @@ def usage(c=None):
...
     launch(sys.argv[1:])
 
 
-def launch(orig_args):
+definition_file_path: str = ""
+
+
+def launch(orig_args: List[str]) -> None:
     if sys.version_info.major == 2:
         sys.stderr.write("Python 2.x is no longer supported. Please switch to Python 3.6 or higher.\n")
```
PawelLipski: Same as below

git_machete/cmd.py
```diff
@@ -3478,7 +3544,7 @@ def strip_first_fragment(rb):
             params = parse_options(args, "d:Mn", ["down-fork-point=", "merge", "no-edit-merge", "no-interactive-rebase"])
             read_definition_file()
             expect_no_operation_in_progress()
-            slide_out(params or [current_branch()])
+            slide_out(params or [current_branch()])  # type: ignore
```
PawelLipski: Why ignore? `tuple` vs `list` again?

git_machete/cmd.py
```diff
@@ -2310,11 +2373,11 @@ def prefix_dfs(u_, prefix):
         prefix_dfs(u, prefix=[])
 
     out = io.StringIO()
-    edge_color = {}
-    fp_sha_cached = {}
+    edge_color: Dict[str, str] = {}
+    fp_sha_cached: Dict[str, Optional[str]] = {}  # todo (#): default dict with None
```
PawelLipski: Pls open an issue + add a traditional script that would ban issue-less TODOs (can be on a separate PR ofc)

PawelLipski: (location: `ci/checks` + add a reference from .travis.yml)

git_machete/cmd.py
```diff
@@ -2105,15 +2167,15 @@ def print_new_line(new_status):
...
             # neither rebase nor merge will be suggested in such case anyway.
-            needs_parent_sync = False
+            needs_parent_sync: bool = False
         elif s == DIVERGED_FROM_AND_OLDER_THAN_REMOTE:
             # Avoid unnecessary fork point check if we already know that the branch qualifies for resetting to remote counterpart;
             # neither rebase nor merge will be suggested in such case anyway.
             needs_parent_sync = False
         elif opt_merge:
-            needs_parent_sync = u and not is_ancestor(u, b)
+            needs_parent_sync = u and not is_ancestor(u, b)  # type: ignore
```
PawelLipski: Why ignore? 🤔

git_machete/cmd.py
```diff
@@ -2090,10 +2151,11 @@ def print_new_line(new_status):
         cb = current_branch()
         expect_in_managed_branches(cb)
 
+    b: str
     for b in itertools.dropwhile(lambda x: x != cb, managed_branches):
```
PawelLipski, line 2160: Whoops no syntax for declaring iterator variable type inside of `for`?... like `for b: str in ...`?

git_machete/cmd.py
```diff
@@ -1888,15 +1947,14 @@ def slidable_after(b):
...
 
-    def connected_with_green_edge(bd):
-        return \
-            not is_merged_to_upstream(bd) and \
-            is_ancestor(b, bd) and \
-            (get_overridden_fork_point(bd) or commit_sha_by_revision(b) == fork_point(bd, use_overrides=False))
+    def connected_with_green_edge(bd: str) -> bool:
+        not_merged_and_is_ancestor = not is_merged_to_upstream(bd) and is_ancestor(b, bd)
+        overriden_fp_or_revision_is_fp = get_overridden_fork_point(bd) or commit_sha_by_revision(b) == fork_point(bd, use_overrides=False)
+        return not_merged_and_is_ancestor and overriden_fp_or_revision_is_fp  # type: ignore
```
PawelLipski: Some explicit `bool(...)` cast missing?

git_machete/cmd.py
```diff
@@ -1732,7 +1791,7 @@ def get_fork_point_override_data(b):
     return to_sha, while_descendant_of_sha
 
 
-def get_overridden_fork_point(b):
+def get_overridden_fork_point(b: str) -> Optional[str]:
```
PawelLipski, line 1798: Btw, does mypy warn about potentially None values properly (like Kotlin/Checker Framework does)? or does it give a free pass on dereferencing None values (like Java/Scala)?

git_machete/cmd.py
mkondratek: unfortunately :c that's the only way

git_machete/tests/functional/test_machete.py
PawelLipski: Ok, just asking out of curiosity, leave as is

ci/tox/Dockerfile
PawelLipski: I'd clarify that in the comment

git_machete/cmd.py
mkondratek: https://mypy.readthedocs.io/en/stable/command_line.html#none-and-optional-handling

> `--no-implicit-optional`
This flag causes mypy to stop treating arguments with a None default value as having an implicit Optional type.
>
> For example, by default mypy will assume that the x parameter is of type Optional[int] in the code snippet below since the default parameter is None:
> ```
> def foo(x: int = None) -> None:
>     print(x)
> ```
>If this flag is set, the above snippet will no longer type check: we must now explicitly indicate that the type is Optional[int]:
>
> ```
> def foo(x: Optional[int] = None) -> None:
>     print(x)
> ```

mkondratek: whoops, N/A 
it is already used

mkondratek: however turning `stirct-optional` on results with about 70 errors like that:
```
➜  git-machete git:(refactor/mypy) ✗ mypy --config-file mypy.ini git_machete
git_machete/cmd.py:355: error: Argument 1 to "len" has incompatible type "Optional[str]"; expected "Sized"
git_machete/cmd.py:356: error: Unsupported operand types for * ("None" and "int")
git_machete/cmd.py:356: note: Left operand is of type "Optional[str]"
git_machete/cmd.py:359: error: Item "None" of "Optional[str]" has no attribute "__iter__" (not iterable)
git_machete/cmd.py:430: error: Unsupported operand types for * ("int" and "None")
git_machete/cmd.py:430: note: Right operand is of type "Optional[str]"
...
```

mkondratek: In particular, this is the warning you asked for 
```
git_machete/cmd.py:1775: error: Argument 1 to "commit_sha_by_revision" has incompatible type "Optional[str]"; expected "str"
```

PawelLipski: Whoops okay so pls add an issue for enabling strict-optional one day

git_machete/cmd.py
```diff
@@ -1494,16 +1526,16 @@ def is_merged_to(b, target):
...
 
-def is_merged_to_upstream(b):
+def is_merged_to_upstream(b: str) -> bool:
     if b not in up_branch:
         return False
     return is_merged_to(b, up_branch[b])
 
 
-def infer_upstream(b, condition=lambda u: True, reject_reason_message=""):
+def infer_upstream(b: str, condition: Callable[..., bool] = lambda u: True, reject_reason_message: str = "") -> Optional[str]:
```
PawelLipski: Hmm why `...` ? is it some kind of varargs?

PawelLipski: Shouldn't the Callable's argument type be just `str` here?

git_machete/cmd.py
```diff
@@ -1449,37 +1473,45 @@ def generate_entries():
...
+            for sha_, branch_defs in branch_defs_by_sha_in_reflog.items():  # type: ignore
+
+                def lb_(lb: str, lb_or_rb: str) -> str:
+                    return lb if lb == lb_or_rb else f"{lb_or_rb} (remote counterpart of {lb})"
 
-        def log_result():
-            for sha_, branch_defs in branch_defs_by_sha_in_reflog.items():
-                yield dim("%s => %s" %
-                          (sha_, ", ".join(map(tupled(lambda lb, lb_or_rb: lb if lb == lb_or_rb else f"{lb_or_rb} (remote counterpart of {lb})"), branch_defs))))
+                yield dim("%s => %s" % (sha_, ", ".join(map(tupled(lb_), branch_defs))))
```
PawelLipski: Interpolation?

git_machete/cmd.py
```diff
@@ -1319,10 +1339,12 @@ def get_combined_remote_sync_status(b):
 # Reflog magic
 
 
-reflogs_cached = None
+REFLOG = Tuple[str, str]
```
PawelLipski: Hmm rather `REFLOG_ENTRY` (for `REFLOG`, I'd expect as list of such entries)

git_machete/cmd.py
```diff
@@ -1319,10 +1339,12 @@ def get_combined_remote_sync_status(b):
...
+
+reflogs_cached: Optional[Dict[str, Optional[REFLOG]]] = None
```
PawelLipski: Shouldn't it rather be `Optional[Dict[str, List[REFLOG_ENTRY]]]`? This would explain the `type: ignore`s below

git_machete/cmd.py
```diff
@@ -1285,7 +1305,7 @@ def commits_between(earliest_exclusive, latest_inclusive):
 DIVERGED_FROM_AND_NEWER_THAN_REMOTE = 6
 
 
-def get_relation_to_remote_counterpart(b, rb):
+def get_relation_to_remote_counterpart(b: str, rb: str) -> int:
```
PawelLipski, line 1311: Actually... is there any easy mypy-supported way to define an enum here?

git_machete/cmd.py
```diff
@@ -1255,23 +1271,27 @@ def squash(cb, fork_commit):
...
     run_git("diff", *params)
 
 
-def log(branch):
+def log(branch: str) -> None:
     run_git("log", "^" + fork_point(branch, use_overrides=True), f"refs/heads/{branch}")
 
 
-def commits_between(earliest_exclusive, latest_inclusive):
+Commit = Tuple[str, str, str]
```
PawelLipski: Maybe `Hash_ShortHash_Message` for clarity instead?

git_machete/cmd.py
```diff
@@ -1449,37 +1473,45 @@ def generate_entries():
...
 
         debug(f"match_log_to_filtered_reflogs({b})", "branches containing the given SHA in their filtered reflog: \n%s\n" % "\n".join(log_result()))
 
-    get_second = tupled(lambda lb, lb_or_rb: lb_or_rb)
     for sha in spoonfeed_log_shas(b):
         if sha in branch_defs_by_sha_in_reflog:
             # The entries must be sorted by lb_or_rb to make sure the upstream inference is deterministic
             # (and does not depend on the order in which `generate_entries` iterated through the local branches).
-            containing_branch_defs = sorted(filter(tupled(lambda lb, lb_or_rb: lb != b), branch_defs_by_sha_in_reflog[sha]), key=get_second)
+            branch_defs: List[BRANCH_DEF] = branch_defs_by_sha_in_reflog[sha]  # type: ignore
```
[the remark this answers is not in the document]
mkondratek:
```
➜  git-machete git:(refactor/mypy) ✗ mypy --config-file mypy.ini git_machete
git_machete/cmd.py:1496: error: Incompatible types in assignment (expression has type "Optional[List[str]]", variable has type "List[Tuple[str, str]]")
Found 1 error in 1 file (checked 5 source files)
```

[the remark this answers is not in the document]
mkondratek: Some vars are globally set to `None` (and in `flush_caches`). That why their types are `Optional[smth]`. What about constructing them in the first place and here:
```
def flush_caches() -> None:
    global branch_defs_by_sha_in_reflog, commit_sha_by_revision_cached, config_cached, counterparts_for_fetching_cached, initial_log_shas_cached
    global local_branches_cached, reflogs_cached, remaining_log_shas_cached, remote_branches_cached
    branch_defs_by_sha_in_reflog = None
    commit_sha_by_revision_cached = None
    config_cached = None
    counterparts_for_fetching_cached = None
    initial_log_shas_cached = {}
    local_branches_cached = None
    reflogs_cached = None
    remaining_log_shas_cached = {}
    remote_branches_cached = None
```
???

git_machete/cmd.py
```diff
@@ -936,15 +952,15 @@ def currently_rebased_branch_or_none():
...
+def currently_checked_out_branch_or_none() -> Optional[str]:
     try:
         raw = popen_git("symbolic-ref", "--quiet", "HEAD").strip()
         return re.sub("^refs/heads/", "", raw)
     except MacheteException:
         return None
 
 
-def expect_no_operation_in_progress():
+def expect_no_operation_in_progress() -> Any:
```
PawelLipski: Hmm not `None`?

git_machete/cmd.py
mkondratek: Hmmm.... but the dict can return None in case of `reflog_cached.get(absent_key)` :c

git_machete/cmd.py
```diff
@@ -444,40 +460,40 @@ def down(b, pick_mode):
         return "\n".join(dbs)
 
 
-def first_branch(b):
+def first_branch(b: str) -> Optional[str]:
```
PawelLipski: Hmmm why optional? I can't see an execution path that would lead to `None` getting returned 🤔

git_machete/cmd.py
```diff
@@ -444,40 +460,40 @@ def down(b, pick_mode):
...
     root = root_branch(b, if_unmanaged=PICK_FIRST_ROOT)
     root_dbs = down_branches.get(root)
     return root_dbs[0] if root_dbs else root
 
 
-def last_branch(b):
+def last_branch(b: str) -> Optional[str]:
```
PawelLipski: Same here, why `Optional`?

git_machete/cmd.py
mkondratek: I think it could be handled like this:
https://stackoverflow.com/questions/52624736/type-annotations-for-enum-value

but let's leave it for another PR (since a bunch of renames will be involved)

git_machete/cmd.py
```diff
@@ -266,47 +274,55 @@ def shell_escape(arg):
...
-def run_git(git_cmd, *args, **kwargs):
+def run_git(git_cmd: str, *args: str, **kwargs: Any) -> int:
     exit_code = run_cmd("git", git_cmd, *args, **kwargs)
     if not kwargs.get("allow_non_zero") and exit_code != 0:
         raise MacheteException(f"`{cmd_shell_repr('git', git_cmd, *args, **kwargs)}` returned {exit_code}")
     return exit_code
 
 
-def popen_git(git_cmd, *args, **kwargs):
+def popen_git(git_cmd: str, *args: str, **kwargs: Any) -> str:
```
PawelLipski: I'd enforce `str` for each of the kwarg values if possible

PawelLipski: Here and in other similar functions

git_machete/cmd.py
mkondratek: good point

git_machete/cmd.py
PawelLipski: Oh ok `...` is just _anything_ AFAICS

git_machete/cmd.py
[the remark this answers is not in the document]
PawelLipski: Whoops okay but wait, even disregarding the `Optional`s... is it correct that `List[str]` is assigned to `List[Tuple[str, str]]`? Isn't one of the involved vars mistyped?

git_machete/cmd.py
PawelLipski: Okay... then `Optional[Dict[str, Optional[List[REFLOG_ENTRY]]]]`?

git_machete/cmd.py
mkondratek: Cannot be be more precise in case of these two :c
- `def popen_cmd(cmd: str, *args: str, **kwargs: Any) -> Tuple[int, str, str]:`
- `def run_cmd(cmd: str, *args: str, **kwargs: Any) -> int:`

For the 3 others `Dict[str, str]` has been used

git_machete/cmd.py
mkondratek: Yup (it is called `Ellipsis` IIRC). I tried to be more precise here but I am afraid that it's not possible (yet)

git_machete/cmd.py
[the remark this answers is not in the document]
mkondratek: :facepalm:, you are right

git_machete/cmd.py
mkondratek: created #110

git_machete/cmd.py
mkondratek: created https://github.com/VirtusLab/git-machete/issues/117

git_machete/cmd.py
mkondratek: created #118

git_machete/cmd.py
mkondratek: bool cast

git_machete/cmd.py
mkondratek: wrong type of params and [current_branch()], fixed

git_machete/cmd.py
```diff
@@ -1339,9 +1340,9 @@ def get_combined_remote_sync_status(b: str) -> Tuple[int, Optional[str]]:
 # Reflog magic
 
 
-REFLOG = Tuple[str, str]
+REFLOG_ENTRY = Tuple[str, str]
 
-reflogs_cached: Optional[Dict[str, Optional[REFLOG]]] = None
+reflogs_cached: Optional[Dict[str, Optional[REFLOG_ENTRY]]] = None
```
```suggestion
reflogs_cached: Optional[Dict[str, Optional[List[REFLOG_ENTRY]]]] = None
```

and possibly then you can remove `type: ignore`s below

git_machete/cmd.py
```diff
@@ -1444,7 +1445,7 @@ def get_latest_checkout_timestamps() -> Dict[str, int]:
     return result
 
 
-branch_defs_by_sha_in_reflog: Optional[Dict[str, Optional[List[str]]]] = None
+branch_defs_by_sha_in_reflog: Optional[Dict[str, Optional[Tuple[str, str]]]] = None
```
PawelLipski: Likely
```suggestion
branch_defs_by_sha_in_reflog: Optional[Dict[str, Optional[List[Tuple[str, str]]]]] = None
```

git_machete/cmd.py
```diff
@@ -3415,8 +3416,8 @@ def parse_direction(b: str, allow_current: bool, down_pick_mode: bool) -> Option
             read_definition_file(verify_branches=False)
             b = opt_branch or current_branch()
             expect_in_managed_branches(b)
-            if params:
-                annotate(b, params)
+            if len(params) == 1:
+                annotate(b, params[0])
```
PawelLipski: Whoops... this is almost surely an error, `annotate` accepts a list of strings (see line 607: `annotations[b] = " ".join(words)`)

git_machete/cmd.py
```diff
@@ -3544,7 +3545,7 @@ def strip_first_fragment(rb: str) -> str:
             params = parse_options(args, "d:Mn", ["down-fork-point=", "merge", "no-edit-merge", "no-interactive-rebase"])
             read_definition_file()
             expect_no_operation_in_progress()
-            slide_out(params or [current_branch()])  # type: ignore
+            slide_out(params[0] if len(params) == 1 else current_branch())
```
PawelLipski: Same here, `slide_out` accepts a list

git_machete/cmd.py
```diff
@@ -1285,17 +1285,18 @@ def log(branch: str) -> None:
...
-def commits_between(earliest_exclusive: str, latest_inclusive: str) -> List[Commit]:
+def commits_between(earliest_exclusive: str, latest_inclusive: str) -> List[Hash_ShortHash_Message]:
     # Reverse the list, since `git log` by default returns the commits from the latest to earliest.
     return list(reversed(list(map(
         lambda x: x.split(":", 2),  # type: ignore
         non_empty_lines(popen_git("log", "--format=%H:%h:%s", f"^{earliest_exclusive}", latest_inclusive, "--"))
     ))))
 
 
+# TODO (#117): extract to namespace, use mypy
```
PawelLipski, line 1301: namespace or enum? 🤔 or is "namespace" a word for enum?

git_machete/cmd.py
```diff
@@ -16,36 +17,42 @@
...
 
-def non_empty_lines(s):
+def non_empty_lines(s: str) -> List[str]:
     return list(filter(None, s.split("\n")))
 
 
 # Converts a lambda accepting N arguments to a lambda accepting one argument, an N-element tuple.
 # Name matching Scala's `tupled` on `FunctionX`.
-def tupled(f):
+def tupled(f: Callable[..., T]) -> Callable[..., T]:
```
PawelLipski: IIUC this could be a bit more specific:
```suggestion
def tupled(f: Callable[..., T]) -> Callable[[Any], T]:
```
since the resulting lambda accepts exactly one arg (actually, what's the difference b/w `...` and `Any`? 🤔 )

PawelLipski: Also... is there any umbrella type for `Tuple`, no matter the arity? like some `TupleN`?

git_machete/cmd.py
```diff
@@ -1680,26 +1741,26 @@ def fork_point_and_containing_branch_defs(b, use_overrides):
...
 
 # Also includes config that is incomplete (only one entry out of two) or otherwise invalid.
-def has_any_fork_point_override_config(b):
+def has_any_fork_point_override_config(b: str) -> bool:
     return (get_config_or_none(config_key_for_override_fork_point_to(b)) or
             get_config_or_none(config_key_for_override_fork_point_while_descendant_of(b))) is not None
 
 
-def get_fork_point_override_data(b):
+def get_fork_point_override_data(b: str) -> Optional[BRANCH_DEF]:
```
PawelLipski: Actually... that's not a branch def  that's returned (local+remote branch name), that's a pair of SHAs, see e.g. line ~second partition of Poland~ 1793:
`return to_sha, while_descendant_of_sha`

PawelLipski: It also boils down to two strings... but I'd still extract a separate type here

git_machete/cmd.py
PawelLipski: Generally, would be nice if you eyeballed all other places where the changes included more than just adding type annotations 🤔 (i.e. the places where the logic has potentially changed)

git_machete/cmd.py
PawelLipski: Ok.... so what's the error if that's replaced with `Callable[str, bool]`? Or maybe just try `Callable[Any, bool]`?

git_machete/cmd.py
```diff
@@ -1449,37 +1474,46 @@ def generate_entries():
...
 
-    get_second = tupled(lambda lb, lb_or_rb: lb_or_rb)
     for sha in spoonfeed_log_shas(b):
         if sha in branch_defs_by_sha_in_reflog:
             # The entries must be sorted by lb_or_rb to make sure the upstream inference is deterministic
             # (and does not depend on the order in which `generate_entries` iterated through the local branches).
-            containing_branch_defs = sorted(filter(tupled(lambda lb, lb_or_rb: lb != b), branch_defs_by_sha_in_reflog[sha]), key=get_second)
+            branch_defs: List[BRANCH_DEF] = branch_defs_by_sha_in_reflog[sha]  # type: ignore
+
+            def lb_b(lb: str, lb_or_rb: str) -> bool:
```
```suggestion
            def lb_is_not_b(lb: str, lb_or_rb: str) -> bool:
```

git_machete/cmd.py
```diff
@@ -1449,37 +1474,46 @@ def generate_entries():
...
 
-        def log_result():
-            for sha_, branch_defs in branch_defs_by_sha_in_reflog.items():
-                yield dim("%s => %s" %
-                          (sha_, ", ".join(map(tupled(lambda lb, lb_or_rb: lb if lb == lb_or_rb else f"{lb_or_rb} (remote counterpart of {lb})"), branch_defs))))
+        def log_result() -> Generator[str, None, None]:
+            branch_defs: List[BRANCH_DEF]
+            for sha_, branch_defs in branch_defs_by_sha_in_reflog.items():  # type: ignore
+
+                def lb_(lb: str, lb_or_rb: str) -> str:
```
PawelLipski: Maybe
```suggestion
                def branch_def_to_str(lb: str, lb_or_rb: str) -> str:
```

git_machete/cmd.py
```diff
@@ -1449,37 +1474,46 @@ def generate_entries():
             if sha in branch_defs_by_sha_in_reflog:
                 # The practice shows that it's rather unlikely for a given commit to appear on filtered reflogs of two unrelated branches
                 # ("unrelated" as in, not a local branch and its remote counterpart) but we need to handle this case anyway.
-                branch_defs_by_sha_in_reflog[sha] += [branch_def]
+                branch_defs_by_sha_in_reflog[sha] += [branch_def]  # type: ignore
```
PawelLipski: Yeah, this is likely wrong, comes from missing `List` in `branch_defs_by_sha_in_reflog` type

git_machete/cmd.py
mkondratek: possibly my bad wording. I meant that after the changes it will be something like `SYNC_STATUS(what I meant to be the "namespace").UNTRACKED`

git_machete/cmd.py
PawelLipski: Or just `Tuple[...]`?

mkondratek: For `Tuple[...]`:
```
git_machete/cmd.py:49: error: Unexpected '...'
git_machete/cmd.py:1489: error: Argument 1 to "tupled" has incompatible type "Callable[[str, str], str]"; expected "Callable[[Tuple[Any]], str]"
```
For `Tuple`:
```
git_machete/cmd.py:49: error: Missing type parameters for generic type "Tuple"
git_machete/cmd.py:1489: error: Argument 1 to "tupled" has incompatible type "Callable[[str, str], str]"; expected "Callable[[Tuple[Any, ...]], str]"
```

But the returned type can be more precise indeed

git_machete/cmd.py
```diff
@@ -16,36 +17,42 @@
...
 
 
 # Converts a lambda accepting N arguments to a lambda accepting one argument, an N-element tuple.
 # Name matching Scala's `tupled` on `FunctionX`.
-def tupled(f):
+def tupled(f: Callable[..., T]) -> Callable[[Any], T]:
     return lambda tple: f(*tple)
 
 
+get_second: Callable[[Any], T] = tupled(lambda a, b: b)  # type: ignore
```
PawelLipski: Redefine `get_second` as a regular `def`

PawelLipski: Also, use `Tuple[S, T]` or sth like that for type of input param

git_machete/cmd.py
```diff
@@ -1255,27 +1274,29 @@ def squash(cb, fork_commit):
...
+def log(branch: str) -> None:
     run_git("log", "^" + fork_point(branch, use_overrides=True), f"refs/heads/{branch}")
 
 
-def commits_between(earliest_exclusive, latest_inclusive):
+def commits_between(earliest_exclusive: str, latest_inclusive: str) -> List[Hash_ShortHash_Message]:
     # Reverse the list, since `git log` by default returns the commits from the latest to earliest.
     return list(reversed(list(map(
-        lambda x: x.split(":", 2),
+        lambda x: x.split(":", 2),  # type: ignore
```
PawelLipski: Nit: the error here is:
`git_machete/cmd.py:1294: error: Argument 1 to "map" has incompatible type "Callable[[str], List[str]]"; expected "Callable[[str], Tuple[str, str, str]]"`

So I'd at least convert to lamda's return type to `Tuple` to align it better with the expected type

git_machete/cmd.py
```diff
@@ -252,8 +260,8 @@ def popen_cmd(cmd, *args, **kwargs):
 # Git core
 
 
-def cmd_shell_repr(cmd, *args, **kwargs):
-    def shell_escape(arg):
+def cmd_shell_repr(cmd: str, *args: str, **kwargs: Dict[str, str]) -> str:
```
PawelLipski, line 265: Actually... why `kwargs` are `Dict[str, str]` but `args` are `str` and not `List[Str]`? 🤔

PawelLipski: That seems inconsistent...

git_machete/cmd.py
```diff
@@ -266,47 +274,55 @@ def shell_escape(arg):
     return " ".join(env_repr + [cmd] + list(map(shell_escape, args)))
 
 
-def run_git(git_cmd, *args, **kwargs):
+def run_git(git_cmd: str, *args: str, **kwargs: Dict[str, str]) -> int:
```
PawelLipski, line 279: Whoops I'm afraid this might be passing only coz `kwargs` are never actually provided to `run_git`...

git_machete/cmd.py
```diff
@@ -266,47 +274,55 @@ def shell_escape(arg):
...
     exit_code = run_cmd("git", git_cmd, *args, **kwargs)
     if not kwargs.get("allow_non_zero") and exit_code != 0:
         raise MacheteException(f"`{cmd_shell_repr('git', git_cmd, *args, **kwargs)}` returned {exit_code}")
     return exit_code
 
 
-def popen_git(git_cmd, *args, **kwargs):
+def popen_git(git_cmd: str, *args: str, **kwargs: Dict[str, str]) -> str:
```
PawelLipski, line 286: Same here... I changed one invocation to `popen_git("version", foo="bar")`... and received error:

`git_machete/cmd.py:698: error: Argument "foo" to "popen_git" has incompatible type "str"; expected "Dict[str, str]"`

PawelLipski: So I guess the type of `kwargs` is really supposed to mean "type of each kwarg", not "type of the collection of all kwargs"

git_machete/cmd.py
mkondratek: Results of `mypy` while annotating `kwargs` as `str`:
```
➜  git-machete git:(refactor/mypy) ✗ mypy --config-file mypy.ini git_machete
git_machete/cmd.py:274: error: Need type annotation for 'env'
git_machete/cmd.py:276: error: Item "str" of "Union[str, Dict[Any, Any]]" has no attribute "items"
git_machete/cmd.py:283: error: Argument 4 to "cmd_shell_repr" has incompatible type "**Dict[str, Dict[str, str]]"; expected "str"
git_machete/cmd.py:290: error: Argument 4 to "cmd_shell_repr" has incompatible type "**Dict[str, Dict[str, str]]"; expected "str"
```

Results of `mypy` while annotating `args` as `List[str]`:
```
➜  git-machete git:(refactor/mypy) ✗ mypy --config-file mypy.ini git_machete
git_machete/cmd.py:220: error: Argument 2 to "cmd_shell_repr" has incompatible type "*Tuple[str, ...]"; expected "List[str]"
git_machete/cmd.py:241: error: Argument 2 to "cmd_shell_repr" has incompatible type "*Tuple[str, ...]"; expected "List[str]"
git_machete/cmd.py:277: error: Argument 1 to "map" has incompatible type "Callable[[str], str]"; expected "Callable[[List[str]], str]"
git_machete/cmd.py:283: error: Argument 2 to "cmd_shell_repr" has incompatible type "str"; expected "List[str]"
git_machete/cmd.py:283: error: Argument 3 to "cmd_shell_repr" has incompatible type "*Tuple[str, ...]"; expected "List[str]"
git_machete/cmd.py:290: error: Argument 2 to "cmd_shell_repr" has incompatible type "str"; expected "List[str]"
git_machete/cmd.py:290: error: Argument 3 to "cmd_shell_repr" has incompatible type "*Tuple[str, ...]"; expected "List[str]"
```
---


Specifying the type of `*args` tells us what EACH "arg" is.
If we use `str` then each of them is a string.
If we use `List[str]` then they are all to be a list of strings.

The same for `kwargs`.

In our case, the "kwarg" is not a type of `str`. It is `Dict[str, str]`.
Why? See how we use it (the type of `kwargs.get`)...
```
env: Dict[str, str] = kwargs.get("env", {})
    # ...
    env_repr = [k + "=" + shell_escape(v) for k, v in env.items() if k not in os.environ]
```


--- 

Refs: 
https://www.python.org/dev/peps/pep-0484/#arbitrary-argument-lists-and-default-argument-values
https://stackoverflow.com/questions/37031928/type-annotations-for-args-and-kwargs

git_machete/cmd.py
mkondratek: > Same here... I changed one invocation to `popen_git("version", foo="bar")`... and received error:
> 
> `git_machete/cmd.py:698: error: Argument "foo" to "popen_git" has incompatible type "str"; expected "Dict[str, str]"`

And this is fine since the type is for each of the "kwargs" and we want it to be a `Dict`.
See my comment above and look close to what happens with `kwargs` and what is the type of its values later.

git_machete/cmd.py
mkondratek: strange... 

```
git_machete/cmd.py:2437: error: Argument 1 to "map" has incompatible type "Callable[[Tuple[T, T]], T]"; expected "Callable[[Tuple[str, str]], T]"
Found 1 error in 1 file (checked 5 source files)
```
for `S` and `T` (the `VarType`s) :confused: 

but since all of the invocations reference strings lets use 
```
def get_second(pair: Tuple[str, str]) -> str:
```
(and since it passes mypy)

git_machete/cmd.py
PawelLipski: Aww crap I forgot about this `env` thing 🤦🏻

PawelLipski (approved):

git_machete/cmd.py
PawelLipski: There might be different `str`-typed params in the future (or maybe they were in the past?) but yeah, `Dict[str, str]` FTW as for now

PR #109, opened 2021-02-12
Refactor/drop global state

### LINKED ISSUE #27: Get rid of global state in the implementation

mkondratek:
- [x] CommandLineContext (done should be split b/w the rest)
- [ ] GitContext
- [x] MacheteContext
- [ ] SystemContext

### DIFF
--- a/git_machete/cmd.py
+++ b/git_machete/cmd.py
@@ -117,8 +117,37 @@ def right_arrow() -> str:
     return "->" if ascii_only else u"➔"
 
 
-def ask_if(msg: str, opt_yes_msg: Optional[str], apply_fmt: bool = True) -> str:
-    if opt_yes and opt_yes_msg:
+class CommandLineContext:
+    opt_as_root: bool = False
+    opt_branch: Optional[str] = None
+    opt_checked_out_since: Optional[str] = None
+    opt_color: str = "auto"
+    opt_debug: bool = False
+    opt_down_fork_point: Optional[str] = None
+    opt_fetch: bool = False
+    opt_fork_point: Optional[str] = None
+    opt_inferred: bool = False
+    opt_list_commits: bool = False
+    opt_list_commits_with_hashes: bool = False
+    opt_merge: bool = False
+    opt_n: bool = False
+    opt_no_edit_merge: bool = False
+    opt_no_interactive_rebase: bool = False
+    opt_onto: Optional[str] = None
+    opt_override_to: Optional[str] = None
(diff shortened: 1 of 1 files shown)


git_machete/cmd.py
```diff
@@ -115,8 +115,37 @@ def right_arrow() -> str:
     return "->" if ascii_only else u"➔"
 
 
-def ask_if(msg: str, opt_yes_msg: Optional[str], apply_fmt: bool = True) -> str:
-    if opt_yes and opt_yes_msg:
+class CommandLineContext:
```
PawelLipski, line 120: I think this changeset can be merged as a self-contained PR, and other planned changes can go on separate PRs

git_machete/cmd.py
```diff
@@ -2307,59 +2339,60 @@ def print_new_line(new_status: bool) -> None:
...
+    if cli_ctxt.opt_return_to == "here" or (cli_ctxt.opt_return_to == "nearest-remaining" and nearest_remaining_branch == initial_branch):
         print(f"Returned to the initial branch {bold(initial_branch)}")
-    elif opt_return_to == "nearest-remaining" and nearest_remaining_branch != initial_branch:
+    elif cli_ctxt.opt_return_to == "nearest-remaining" and nearest_remaining_branch != initial_branch:
         print(
             f"The initial branch {bold(initial_branch)} has been slid out. Returned to nearest remaining managed branch {bold(nearest_remaining_branch)}")
 
 
-def status(warn_on_yellow_edges: bool) -> None:
+def status(cli_ctxt: CommandLineContext, warn_on_yellow_edges: bool) -> None:
```
PawelLipski, line 2400: Probably most of these methods will require all other contexts to be passed as arguments as well...

PawelLipski: Maybe wrapping then into single unified `Context` with `cli`/`machete`/`git` etc. fields won't be a bad idea

mkondratek: sounds good but let's leave the more specific name for now since the other context will be added on the separate PR(s)

PawelLipski (approved):

PR #123, opened 2021-03-12
Remove redundant code

mkondratek:


PawelLipski (approved):

mkondratek:
I should take this into account first: 
https://stackoverflow.com/questions/207000/what-is-the-difference-between-class-and-instance-attributes

git_machete/cmd.py
```diff
@@ -118,32 +118,33 @@ def right_arrow() -> str:
...
-    opt_override_to_inferred: bool = False
-    opt_override_to_parent: bool = False
-    opt_return_to: str = "stay"
-    opt_roots: List[str] = list()
-    opt_start_from: str = "here"
-    opt_stat: bool = False
-    opt_unset_override: bool = False
-    opt_verbose: bool = False
-    opt_yes: bool = False
+    def __init__(self) -> None:
```
PawelLipski, line 121: This could be a `@dataclass`... but AFAICS it's only supported since Python 3.7 without adding any extra dependency on some backport library :/

PawelLipski (approved):

git_machete/cmd.py
mkondratek: Yeah, I thought about it but I share your concerns too

PR #124, opened 2021-03-13
Refactor/machete context

mkondratek:


mkondratek:
image

mkondratek:
https://stackoverflow.com/questions/207000/what-is-the-difference-between-class-and-instance-attributes :open_mouth:

git_machete/cmd.py
```diff
@@ -327,22 +327,25 @@ def popen_git(cli_ctxt: CommandLineContext, git_cmd: str, *args: str, **kwargs:
...
+        self.down_branches: Dict[str, List[str]] = {}  # TODO (#110): default dict with []
+        self.up_branch: Dict[str, str] = {}  # TODO (#110): default dict with None
+        self.indent: Optional[str] = None
+        self.roots: List[str] = []
+        self.annotations: Dict[str, str] = {}
 
 
-def expect_in_managed_branches(b: str) -> None:
-    if b not in managed_branches:
+def expect_in_managed_branches(machete_ctxt: MacheteContext, b: str) -> None:
```
PawelLipski, line 342: Hmm would it make sense to make all those functions into methods of `MacheteContext` instead? 🤔

PawelLipski: This may have made less sense with `cli_ctxt`... but here it feels more justified

asford: I just came across this issue as well, and think this would be an ideal solution, 
as it would provide a natural API for machete.

I would imagine a class layout where the functions that require the machete context are implemented as class method over the context object, and this object holds it's own reference to the cli context. Eg:

```python
class CommandLineContext:
    ....

class MacheteContext:
    cli_ctxt: CommandLineContext
    
    managed_branches : List[str]
    # et al context variables
    
    @staticmethod
    def read_definition(cli_ctxt: CommandLineContext, verify_branches: bool = True) -> MacheteContext:
        ...
    
    def render_tree(self) -> List[str]:
        ...
        
    def save_definition_file(self) -> None:
        ...
    
    # et al methods
```

Does that structure make sense to you @PawelLipski?

PawelLipski: Wow that's a surprise comment, thanks :) yes, indeed, this make sense :)

PawelLipski: Mikołaj (the original PR author) is leaving our company soon... I'll either revisit this PR myself, or try to find some Python-savvy person in VirtusLab that could finish up this PR ;)

PawelLipski: Also:

> it would provide a natural API for machete.

Just out of curiosity, do you have a use case for using git-machete programatically? I've got that in the back of my mind, but haven't got a use case/need for that myself so far

asford: We'd like to use `import git_machete` as a library component to implement a few github workflow helper scripts.
For example, opening stacks of PRs for a given machete configuration.
I think, in general, it makes sense for "higher level" tools or utilities to consume a local configuration tool like machete,
as these tools are very likely to be workflow specific and/or target different repo providers (github/gitlab/et al).
Right now machete's light github integration is useful, but it'd be undesirable to extend it much further into the kinds of highly customized, workflow specific features that it could be used for by specific consuming applications.

PawelLipski: Hmmm regardless of providing a programmatic API (which I'm gonna do myself or ask someone soon-ish), I was actually thinking of adding a `git machete hub` command as well, with e.g. the following subcommands:

* `git machete hub anno-prs` - same as `git machete anno --sync-github-prs`, just add annotations with PR numbers/authors to machete,
* `git machete hub create-pr` - create a PR from current branch, using upstream from machete as PR base,
* `git machete hub checkout-pr <pr-number>` - fetch the remotes & check out the branch corresponding to the given PR & add it to machete (potentially with entire the path up to its root, reflecting the PR structure)
* `git machete hub retarget-pr` - change PR base for the current branch so that this PR base matches the upstream from machete

WDYT? do you think it'd be useful? if so, can you see anything that could be included?

PawelLipski: Although I agree that any command like that will remain pretty provider-specific, I'm just more inclined to choose github coz well, it's the most popular one ;)

asford: Agree that it'd be useful.

I would recommend making these thin wrappers around the existing `gh` cli and intentionally hew to that interface.

For instance:

* `git machete gh pr create` -> `gh pr create` with machete-inferred `--base` and `--head` branch, passthrough all other options.
* `git machete gh pr checkout` -> `gh pr checkout`, passthrough args, then add new branch and/or base to machete

One thing that would be very useful right now, for the existing "annotate" feature, would be to just extract the current token from `gh` if it's available via `gh auth status --show-token`.

PawelLipski: > `git machete gh pr create`

Good idea wrt. sticking to `gh` interface, I've got used to `hub` more, didn't use `gh` much... but my proposition (I've put work the done so far on https://github.com/VirtusLab/git-machete/pull/130/files) doesn't stick to `hub`'s conventions either ;p

> I would recommend making these thin wrappers around the existing gh cli 

I'm not sure about that point, I'd rather make the requirements for the user as low as possible... `gh` is a separate piece of software to be installed after all. I'd rather stick to the most light-weight way, via direct access to GH API... although getting the token via `gh auth status --show-token` if possible sounds like a good plan indeed. As for now, I require `GITHUB_TOKEN` to be present in the env.

PawelLipski: FYI @asford &mdash; we'll have a person working full-time on tasks like this one since ~July 15th (my throughput is heavily limited by a commercial project :/)

PawelLipski:
Closing in favor of @rifel123's work on #151 (started from scratch, a rebase of this PR on `develop` would be too complicated)

PR #125, opened 2021-03-14
Provide syncing custom annotations to github PR numbers

PawelLipski:


git_machete/cmd.py
```diff
@@ -2661,21 +2708,28 @@ def usage(c: str = None) -> None:
...
 
+            If invoked with `-H` or `--sync-github-prs`, annotates the branches based on their corresponding GitHub PR numbers and authors.
+            To allow GitHub API access for private repositories (and also to correctly identify the current user, even in case of public repositories),
+            `GITHUB_TOKEN` env var must contain a GitHub API token with `repo` scope, see `https://github.com/settings/tokens`.
+
             In any other case, sets the annotation for the given/current branch to the given argument.
             If multiple arguments are passed to the command, they are concatenated with a single space.
 
-            Note: the same effect can be always achieved by manually editing the definition file.
+            Note: all the effects of `anno` can be always achieved by manually editing the definition file.
```
mkondratek, line 2730: :+1:

git_machete/github.py
```diff
@@ -0,0 +1,96 @@
...
+import os
+import re
+from http.client import HTTPResponse, HTTPSConnection
+from typing import Dict, List, Optional, Any, Tuple
+
+from git_machete.cmd import MacheteException, fmt
+
+
+class GitHubPullRequest(object):
+    def __init__(self, number: int, user: str, base: str, head: str):
```
mkondratek, line 15: Does it pass `mypy`? :thinking: Or maybe `mypy` is not execute for this file? 

Cause IIRC the __init__ shall be typed as ` -> None`...

git_machete/github.py
```diff
@@ -0,0 +1,96 @@
+#!/usr/bin/env python
+
+import http.client
+import json
+import os
+import re
+from http.client import HTTPResponse, HTTPSConnection
```
mkondratek, line 8: it belongs to std lib... cool :rocket:

mkondratek (approved):
LGTM

And one more (silly?) question... is it possible to test this feature?

I assume it would require a lot of setting and the cake's not worth the candle but I am just thinking "how?"...
some github repository with branches, etc, its token passed to CI.

PawelLipski:
> And one more (silly?) question... is it possible to test this feature?
> I assume it would require a lot of setting and the cake's not worth the candle but I am just thinking "how?"...
> some github repository with branches, etc, its token passed to CI.

Ideally, some test docker container with a dummy GitHub API server, so that the tests don't need to rely on any external state like some PRs in GitHub... although I can't find any such ready Docker image. Soo... as a fallback, probably some dumb mock server that would replay pre-recorded responses

git_machete/github.py
PawelLipski: Actually... that's the bad client, the good one is apparently `requests` (https://requests.readthedocs.io/en/master/)... feels so much easier to use...

PawelLipski: But `requests` don't belong to std lib, and I don't want to add any external deps unless absolutely necessary (so far it hasn't been, and let it stay this way)

README.md
```diff
@@ -148,6 +148,14 @@ git machete advance
 
 Useful for merging the child branch to the current branch in a linear fashion (without creating a merge commit).
 
+### GitHub integration
+```shell script
+git machete anno --sync-github-prs
+```
+
+This will automatically annotate the branches with GitHub PR numbers. <br/>
```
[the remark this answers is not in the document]
PawelLipski: Will be overwritten, lemme spec that in the help...

git_machete/github.py
PawelLipski: It is executed, definitely... apparently it doesn't complain about c'tor not being typed? did it complain to you?

mkondratek: hmmm... :thinking: possibly I mistyped the constructor with `'<ClassName>'` instead of `None` and that's from my error came

mkondratek (approved):

PR #128, opened 2021-05-24
Add support for 0-or-more downstream branches in slide_out

asford:
Updates the core `slide_out` interface to add support when the "tail" of the slide-out branch set has any number of downstream branches. 

This provides slide-out support for trees of the structure...

```
main
|
o-slide_me_out
    |
    o-child_a
    |
    o-child_b
```

...as may often occur when multiple feature branches share a common bugfix.

Refactors and expands the functional tests to include this feature.

This does *not* update any of the existing auto-detection of slide-able branches to these cases.

----

Potential finishing items:

- [x] Update docs.
- [x] Update changelog.


PawelLipski:
Nice surprise 🎉 lemme do a superficial review first, I'll do a more in-depth one tomorrow-ish 👁️

git_machete/tests/functional/test_machete.py
```diff
@@ -4,193 +4,74 @@
...
 from git_machete import cmd
 
 
 class SandboxSetup:
-
     def __init__(self) -> None:
         self.file_dir = os.path.dirname(os.path.abspath(__file__))
-        self.remote_path = os.popen('mktemp -d').read().strip()
-        self.sandbox_path = os.popen('mktemp -d').read().strip()
+        self.remote_path = os.popen("mktemp -d").read().strip()
```
PawelLipski, line 18: Actually... any reason for preferring `"` over `'`? 🤔 or just for the sake of consistency?

git_machete/tests/functional/test_machete.py
```diff
@@ -204,14 +85,311 @@ def launch_command(*args: str) -> str:
...
+            .execute('git config user.name "Tester Test"')
+        )
+
     def test_discover_traverse_squash(self) -> None:
-        Setup.setup_sandbox()
-        self.launch_command('discover', '-y', '--roots=develop,master')
-        self.assertEqual(self.launch_command('status'), expected_status_1)
-        self.launch_command('traverse', '-Wy')
-        self.assertEqual(self.launch_command('status', '-l'), expected_status_l_2)
+        (
```
PawelLipski, line 106: Makes sense to avoid `\`, I didn't think of that ;)

git_machete/tests/functional/test_machete.py
```diff
@@ -204,14 +85,311 @@ def launch_command(*args: str) -> str:
...
+                    o-child_d
+                """
+            ),
+        )
+
+        self.launch_command("slide-out", "-n")
+
+        self.assertEqual(
+            self.launch_command("status", "-l"),
+            self.adapt(
```
PawelLipski: Actually! AFAICS this `self.assertEqual(...., self.adapt(...))` pattern recurs a few times across this test suite... maybe it's worth extracting to a separate method while you're here so that multiple `self.adapt` calls aren't needed?

git_machete/tests/functional/test_machete.py
asford: Oh shoot, this is a result of running [`black`](https://github.com/psf/black) over the file while I was editing.
I don't have a strong preference here, though a zero-configuration `black` run is always nice.

git_machete/cmd.py
```diff
@@ -1978,39 +1978,55 @@ def run_post_slide_out_hook(cli_ctxt: CommandLineContext, new_upstream: str, sli
...
-    if cli_ctxt.opt_merge:
-        print(f"Merging {bold(new_upstream)} into {bold(new_downstream)}...")
-        merge(cli_ctxt, new_upstream, new_downstream)
-    else:
-        print(f"Rebasing {bold(new_downstream)} onto {bold(new_upstream)}...")
-        rebase(cli_ctxt, f"refs/heads/{new_upstream}", cli_ctxt.opt_down_fork_point or fork_point(cli_ctxt, new_downstream, use_overrides=True), new_downstream)
+    run_post_slide_out_hook(cli_ctxt, new_upstream, branches_to_slide_out[-1], new_downstreams)
+
+    go(cli_ctxt, new_upstream)
+    for new_downstream in new_downstreams:
```
PawelLipski, line 2022: Hmmm actually... for the UX point of view, I have a concern here: in case the rebase is interactive (so `-n` option is NOT passed), will the user be able to find out which downstream branch is currently getting rebased? 🤔 I'd just want to make sure that we don't leave the user confused with multiple rebases happening one by one

PawelLipski: Possibly making it similar to what happens in `git machete traverse` (so explicit questions like `Rebase branch 'foo' onto 'bar'?`) would be a good choice for slide-outs where >1 downstream is involved?

git_machete/tests/functional/test_machete.py
PawelLipski: Okay, I didn't know that tool myself... do you think it'd make sense to replace pep8 check with a `black --check` in tox config?

PawelLipski:
Btw @asford, how did you get to know about git-machete? Just checking so that we know how to spread the word ;)

git_machete/tests/functional/test_machete.py
asford: I'd be in favor of that, I typically configure `black` and `isort` via [`pre-commit`](https://pre-commit.com/) hooks, then verify the pre-commit runs in `tox`. I'd be happy to open a follow-up PR for that here if you'd like.

PawelLipski:
Also, I'd activated building PRs from forks in Circle CI, they were off until now since we moved from Travis recently... so the next push should trigger a CI build

git_machete/cmd.py
asford: 🤔 The currently experience isn't bad. The `slide_out` function prints an update message before opening the interactive rebase, so there's pretty good messaging before the rebase fires. Eg:

```bash
mdemo$ git machete status
  main
  │
  └─slide_out
    │
    └─child_a
    │
    └─child_b
    
mdemo$ git machete slide-out
Rebasing child_a onto main...
Successfully rebased and updated refs/heads/child_a.
Rebasing child_b onto main...
Successfully rebased and updated refs/heads/child_b.

mdemo$ git machete status
  main
  │
  └─child_a
  │
  └─child_b
```

I don't think you want to prompt during the `slide-out` operation, because at that point the updated machete configuration has already been written. Aborting there would leave the tree in a vaguely inconsistent state that could be hard to back out from. 

I *personally* don't want an interactive prompt during the operation as slide-out as a primitive operation that operates on a single branch, rather than traverse which is *potentially* operating on multiple branches.

If I'm calling from the cli via `git machete slide-out` I've already looked at and selected the branch, and if it's run during a traverse then I've already been prompted to confirm the slide-out operation at a branch level before firing.

asford:
> Btw @asford, how did you get to know about git-machete? Just checking so that we know how to spread the word ;)

I was looking for a tool to help teach folks this kind of rebase-based pull request workflow, and I think I found this via your blog posts or google. Thanks for the hard work on `machete`, it's been a fantastic tool for our development team to help refactor PRs into reviewable units.

git_machete/tests/functional/test_machete.py
asford: Roger. Would you prefer a follow-up or a merge into this PR?

PawelLipski: In the spirit of git-machete... I'd prefer a separate PR ;)

git_machete/cmd.py
```diff
@@ -3050,15 +3062,17 @@ def usage(c: str = None) -> None:
...
             And now let's assume that `adjust-reads-prec` and later `block-cancel-order` were merged to develop.
             After running `git machete slide-out adjust-reads-prec block-cancel-order` the tree will be reduced to:
             <dim>
               develop
                   change-table
                       drop-location-type
+                  add-notification
             </dim>
-            and `change-table` will be rebased onto develop (fork point for this rebase is configurable, see `-d` option below).
+            and `change-table` and `add-notification` will be rebased onto develop (fork point for this rebase is configurable, see `-d` option below).
```
PawelLipski, line 3075: Whooops... `--down-fork-point` option is tricky in case there's more than one downstream branch hmmm

PawelLipski: Of course it can happen that all N downstream branches have the same fork point (e.g. if they're all in sync with their parent)... but in general each of them can have a different one.

PawelLipski: Actually, _market research_: since I'm not an avid user of `slide-out` myself (I'm sliding out during `traverse`, or via GUI from git-machete-intellij-plugin)... do you remember yourself using `--down-fork-point` frequently?

git_machete/cmd.py
```diff
@@ -1978,53 +1978,66 @@ def run_post_slide_out_hook(cli_ctxt: CommandLineContext, new_upstream: str, sli
...
+
+    go(cli_ctxt, new_upstream)
+    for new_downstream in new_downstreams:
+        go(cli_ctxt, new_downstream)
+        if cli_ctxt.opt_merge:
+            print(f"Merging {bold(new_upstream)} into {bold(new_downstream)}...")
+            merge(cli_ctxt, new_upstream, new_downstream)
+        else:
+            print(f"Rebasing {bold(new_downstream)} onto {bold(new_upstream)}...")
+            rebase(cli_ctxt, f"refs/heads/{new_upstream}", cli_ctxt.opt_down_fork_point or fork_point(cli_ctxt, new_downstream, use_overrides=True), new_downstream)
```
PawelLipski, line 2029: See comment about `down_fork_point` below 🤔

git_machete/cmd.py
PawelLipski: Maybe for clarity we could just forbid using `--down-fork-point` when there's >1 downstream branch? 🤔

git_machete/cmd.py
```diff
@@ -2628,7 +2641,7 @@ def usage(c: str = None) -> None:
         "log": "Log the part of history specific to the given branch",
         "reapply": "Rebase the current branch onto its computed fork point",
         "show": "Show name(s) of the branch(es) relative to the position of the current branch, accepts down/first/last/next/root/prev/up argument",
-        "slide-out": "Slide out the current branch and sync its downstream (child) branch with its upstream (parent) branch via rebase or merge",
+        "slide-out": "Slide out the current branch and sync its downstream (child) branches with its upstream (parent) branch via rebase or merge",
```
PawelLipski, line 2644: If you could also reflect this new change of logic in completion scripts! coz as for now, e.g. `develop` doesn't complete in the below scenario:

image

PawelLipski: Also, out of curiosity, do you use the bash/zsh completion scripts? did you have any trouble setting them up?

git_machete/cmd.py
PawelLipski: Okay, you've convinced me :) let's resolve the confusion around `down-fork-point` tho

git_machete/cmd.py
asford: Oh, I've never actually used `--down-fork-point` and wasn't aware that it was an option.
Any time the fork-point isn't properly inferred I've already configured fork-point overrides for the branch.
I'd consider this "best practices", as it's good to verify via `git machete status -L` if there are any questions on fork points.

With this in mind, I think `--down-fork-point` seems vaguely unsafe to me.
If I'd my druthers I'd probably just remove it and enforce that fork point overrides are configured per-branch with the existing `fork-point` command before running slide out.

Given that it's possible for branches to have a common fork-point, I don't _think_ we need to disable `--down-fork-point`. 
Perhaps just print a warning that this is a potentially unsafe operation.
Heck, maybe even just deprecate `--down-fork-point` unless you know anyone who's _actually_ using it?
I could imagine the deprecation message linking back to an issue for folks to chime in if they're using it.

PawelLipski: > Any time the fork-point isn't properly inferred I've already configured fork-point overrides for the branch.

Yeah, you're right, I'm actually doing that myself, `--fork-point` options of `reapply` and `update`, as well as `--down-fork-point` option of `slide-out` are a relic of the times when fork point override wasn't available yet ;)

> With this in mind, I think --down-fork-point seems vaguely unsafe to me.

I agree, even more so now that you point that out 😅

> Given that it's possible for branches to have a common fork-point, I don't think we need to disable --down-fork-point.

Ok, let's leave the option as is, it's obscure but still _theoretically useful_ so I think it can stay as is. 

So maybe so that you don't spend more effort than it's really worth: what if you just rephrase the help for `--down-fork-point` option in `slide-out`'s help so that it reflects the new behavior correctly?

PawelLipski (dismissed):
Minor comments left, approving in general :)

git_machete/cmd.py
asford: Done, updated existing doc and added reference to `git machete fork-point` as an preferred alternative.

PawelLipski (approved):

PawelLipski:
Feel free to fast-forward aka `git machete advance` develop branch to this PR

asford:
Retargeted against `develop`, I think you need to fire the merge as I don't have commit rights here.

PR #130, opened 2021-05-28
Add `hub` command for GitHub operations

### LINKED ISSUE #82: Provide basic Github integration
Just parse `hub list prs` and e.g. set the annotations to PR numbers + check if PR bases are correct

PawelLipski:


asford:
This looks great. Is it ready for user testing?

PawelLipski:
There's still a couple known issues at this point, lemme add comments here on the PR... I'm gonna follow up on that once I'm ready with v3.2.0 (a master build is running already: https://app.circleci.com/pipelines/github/VirtusLab/git-machete/138/workflows/ab506111-5409-415f-870d-41356ef6ed46/jobs/139)

git_machete/cmd.py
```diff
@@ -648,31 +649,128 @@ def annotate(b: str, words: List[str]) -> None:
...
+        try:
+            return open(path).read()
+        except IOError:
+            return ''
+
+    title: str = popen_git(cli_ctxt, "log", "-1", "--format=%s").strip()
+    description_path = get_git_subpath(cli_ctxt, 'info', 'description')
+    description: str = slurp_file_or_empty(description_path)
+
+    pr: GitHubPullRequest = create_pull_request(org, repo, head=head, base=base, title=title, description=description, draft=draft)
```
PawelLipski: If:
* branch is not pushed yet OR
* a PR for the branch already exists,

then the user is left with a cryptic 422 HTTP status, no further details :/

git_machete/cmd.py
```diff
@@ -648,31 +649,128 @@ def annotate(b: str, words: List[str]) -> None:
...
+
+    if current_user:
+        debug(cli_ctxt, f'create_github_pr({head})', f'adding {current_user} as assignee to PR #{pr.number}')
+        add_assignees_to_pull_request(org, repo, pr.number, [current_user])
+
+    reviewers_path = get_git_subpath(cli_ctxt, 'info', 'reviewers')
+    reviewers: List[str] = non_empty_lines(slurp_file_or_empty(reviewers_path))
+    if reviewers:
+        debug(cli_ctxt, f'create_github_pr({head})', f'adding {", ".join(reviewers)} as reviewers to PR #{pr.number}')
+        add_reviewers_to_pull_request(org, repo, pr.number, reviewers)
```
PawelLipski: Displaying progress won't be a bad idea here probably (esp. in case the operation is interrupted in-between, to see which steps already passed)

git_machete/cmd.py
```diff
@@ -648,31 +649,128 @@ def annotate(b: str, words: List[str]) -> None:
...
+    debug(cli_ctxt, f'retarget_github_pr({head})', f'found {pr}')
+
+    new_base: Optional[str] = up_branch.get(head)
+    if not new_base:
+        raise MacheteException(f'Branch `{head}` does not have a parent branch (it is a root) '
+                               f'even though there is an open PR #{pr.number} to `{pr.base}`.\n'
+                               f'Consider modifying the branch definition file (`git machete edit`) so that `{head}` is a child of `{pr.base}`.')
+
+    if pr.base != new_base:
+        set_base_of_pull_request(org, repo, pr.number, base=new_base)
```
PawelLipski: Some message would be useful here

git_machete/cmd.py
```diff
@@ -2718,6 +2810,24 @@ def print_line_prefix(b_: str, suffix: str) -> None:
 
 
 def usage(c: str = None) -> None:
+    aliases = {
```
PawelLipski: Unrel, makes more sense to keep it at the beginning

git_machete/cmd.py
```diff
@@ -3076,6 +3187,38 @@ def usage(c: str = None) -> None:
...
+              Creates a PR for the current branch, using the upstream (parent) branch as the PR base.
+              Once the PR is successfully created, annotates the current branch with the new PR's number.
+
+              If `.git/info/description` file is present, its contents is used as PR description.
+              If `.git/info/milestone` file is present, its contents (a single number - milestone id) is used as milestone.
+              If `.git/info/reviewers` file is present, its contents (one GitHub login per line) are used to set reviewers.
+
+              <b>Options:</b>
+                <b>--draft</b>    Creates the new PR as a draft.
+
```
PawelLipski: Also planned: `checkout-pr` (or whatever new name it should assume to be more `gh`-compatible), but this will be significantly more complex since it ideally should be able to traverse the PR chain upwards, adding parents to the branch layout one by one

git_machete/cmd.py
```diff
@@ -648,31 +649,128 @@ def annotate(b: str, words: List[str]) -> None:
...
+    reviewers_path = get_git_subpath(cli_ctxt, 'info', 'reviewers')
+    reviewers: List[str] = non_empty_lines(slurp_file_or_empty(reviewers_path))
+    if reviewers:
+        debug(cli_ctxt, f'create_github_pr({head})', f'adding {", ".join(reviewers)} as reviewers to PR #{pr.number}')
+        add_reviewers_to_pull_request(org, repo, pr.number, reviewers)
+
+    annotations[head] = f'PR #{pr.number}'
+    save_definition_file()
+
+    print(fmt(f'Created {pr}, see <b>{pr.html_url}</b>'))
```
PawelLipski: This should be displayed regardless of whether setting reviewers/milestone succeeds or not, as long as creating the PR itself succeeds

git_machete/github.py
```diff
@@ -40,19 +46,19 @@ def fire_github_api_get_request(url: str, token: Optional[str]) -> Any:
...
-            return body
+            return parsed_response_body
         else:
-            first_line = fmt(f'GitHub API returned {response.status} HTTP status with error message: `{body.get("message")}`.\n')
+            first_line = fmt(f'GitHub API returned {response.status} HTTP status with error message: `{parsed_response_body.get("message")}`\n')
             if token:
                 raise MacheteException(
                     first_line + fmt(f'Make sure that the token provided in <b>{GITHUB_TOKEN_ENV_VAR}</b> env var is valid '
-                                     f'and allows for access to `GET https://{host}{url}`.'))
+                                     f'and allows for access to {method} https://{host}{url}.'))
```
PawelLipski: This message sometimes doesn't make sense:

```
➜ 19:41 ~/infra-sandbox bar PR #16 $ git machete hub create-pr 
Creating a PR from bar to master... OK, see https://github.com/PawelLipski/infra-sandbox/pull/17
Setting milestone of PR #17 to 1... OK
Adding PawelLipski as assignee to PR #17... OK
Adding git-machete-bot as reviewers to PR #17...
GitHub API returned 422 HTTP status with error message: Reviews may only be requested from collaborators. One or more of the users or teams you specified is not a collaborator of the PawelLipski/infra-sandbox repository.
Make sure that the token provided in GITHUB_TOKEN env var is valid and allows for access to POST https://api.github.com/repos/PawelLipski/infra-sandbox/pulls/17/requested_reviewers.
```

PawelLipski: Probably it should only be displayed for 403 (and 401?) status

git_machete/cmd.py
PawelLipski: Possibly some `traverse`-like interactive behavior for pushing the branch would be useful here

RELEASE_NOTES.md
```diff
@@ -1,5 +1,9 @@
 # Release notes
 
+## New in git-machete 3.3.0
+
+- added: `hub` command with `anno-prs`, `create-pr` and `retarget-pr` subcommands
```
PawelLipski: Also, `Command properties/classification` in CONTRIBUTING.md needs to be updated

asford:
FWIW, `git machete` is currently broken on develop until this PR is merged:

```
$ python git-machete

Usage: git machete [--debug] [-h] [-v|--verbose] [--version] <command> [command-specific options] [command-specific argument]

TL;DR tip

    Get familiar with the help for format, edit, status and update, in this order.

General topics

    file              Display the location of the definition file
    help              Display this overview, or detailed help for a specified command
    hooks             Display docs for the extra hooks added by git machete
    version           Display the version and exit

Build, display and modify the tree of branch dependencies

    add               Add a branch to the tree of branch dependencies
    anno              Manage custom annotations
    discover          Automatically discover tree of branch dependencies
    edit, e           Edit the definition file
  File "git-machete", line 4, in <module>
    cmd.main()
  File "/home/alexford/asford/git-machete/git_machete/cmd.py", line 3410, in main
    launch(sys.argv[1:])
  File "/home/alexford/asford/git-machete/git_machete/cmd.py", line 3559, in launch
    usage()
  File "/home/alexford/asford/git-machete/git_machete/cmd.py", line 3390, in usage
    print("    %s%-18s%s%s" % (BOLD, cm + alias, ENDC, short_docs[cm]))  # bold(...) can't be used here due to the %-18s format specifier
KeyError: 'hub'
```

https://github.com/asford/git-machete/blob/4f8387d778a4e5c87a3f06b4734b62a4fdf05a12/git_machete/cmd.py#L2729

git_machete/cmd.py
```diff
@@ -648,31 +649,131 @@ def annotate(b: str, words: List[str]) -> None:
...
+        print(fmt(ok_str))
+
+    reviewers_path = get_git_subpath(cli_ctxt, 'info', 'reviewers')
+    reviewers: List[str] = non_empty_lines(slurp_file_or_empty(reviewers_path))
+    if reviewers:
+        print(fmt(f'Adding {", ".join(f"`{r}`" for r in reviewers)} as reviewer{"s" if len(reviewers) > 1 else ""} to PR #{pr.number}...'), end='', flush=True)
+        add_reviewers_to_pull_request(org, repo, pr.number, reviewers)
+        print(fmt(ok_str))
+
+    annotations[head] = f'PR #{pr.number}'
```
asford: It would actually be *extremely* useful if this annotation were the full PR url (eg. `https://github.com/VirtusLab/git-machete/pull/130`) rather than just the PR number.

This is then auto-converted into a clickable link is more terminals, which makes for a _super_ slick one-click-from-machete-to-PR experience.

PawelLipski:
> FWIW, `git machete` is currently broken on develop until this PR is merged:
> 
> ```
> $ python git-machete
> 
> Usage: git machete [--debug] [-h] [-v|--verbose] [--version] <command> [command-specific options] [command-specific argument]
> 
> TL;DR tip
> 
>     Get familiar with the help for format, edit, status and update, in this order.
> 
> General topics
> 
>     file              Display the location of the definition file
>     help              Display this overview, or detailed help for a specified command
>     hooks             Display docs for the extra hooks added by git machete
>     version           Display the version and exit
> 
> Build, display and modify the tree of branch dependencies
> 
>     add               Add a branch to the tree of branch dependencies
>     anno              Manage custom annotations
>     discover          Automatically discover tree of branch dependencies
>     edit, e           Edit the definition file
> Traceback (most recent call last):
>   File "git-machete", line 4, in <module>
>     cmd.main()
>   File "/home/alexford/asford/git-machete/git_machete/cmd.py", line 3410, in main
>     launch(sys.argv[1:])
>   File "/home/alexford/asford/git-machete/git_machete/cmd.py", line 3559, in launch
>     usage()
>   File "/home/alexford/asford/git-machete/git_machete/cmd.py", line 3390, in usage
>     print("    %s%-18s%s%s" % (BOLD, cm + alias, ENDC, short_docs[cm]))  # bold(...) can't be used here due to the %-18s format specifier
> KeyError: 'hub'
> ```
> 
> https://github.com/asford/git-machete/blob/4f8387d778a4e5c87a3f06b4734b62a4fdf05a12/git_machete/cmd.py#L2729

Whoops good spot 😅 I'll add a regression test for that after the weekend

git_machete/cmd.py
PawelLipski: Good idea... the only caveat I can see is that the output of `status`/`traverse` may become pretty cluttered :/ I'd make the annotation format controllable by a command-line switch... either a flag (`--url`?) or even a separate subcommand (`anno-pr-nums` vs `anno-pr-urls`?)

PawelLipski:
> > FWIW, `git machete` is currently broken on develop until this PR is merged:
> > ```
> > $ python git-machete
> > 
> > Usage: git machete [--debug] [-h] [-v|--verbose] [--version] <command> [command-specific options] [command-specific argument]
> > 
> > TL;DR tip
> > 
> >     Get familiar with the help for format, edit, status and update, in this order.
> > 
> > General topics
> > 
> >     file              Display the location of the definition file
> >     help              Display this overview, or detailed help for a specified command
> >     hooks             Display docs for the extra hooks added by git machete
> >     version           Display the version and exit
> > 
> > Build, display and modify the tree of branch dependencies
> > 
> >     add               Add a branch to the tree of branch dependencies
> >     anno              Manage custom annotations
> >     discover          Automatically discover tree of branch dependencies
> >     edit, e           Edit the definition file
> > Traceback (most recent call last):
> >   File "git-machete", line 4, in <module>
> >     cmd.main()
> >   File "/home/alexford/asford/git-machete/git_machete/cmd.py", line 3410, in main
> >     launch(sys.argv[1:])
> >   File "/home/alexford/asford/git-machete/git_machete/cmd.py", line 3559, in launch
> >     usage()
> >   File "/home/alexford/asford/git-machete/git_machete/cmd.py", line 3390, in usage
> >     print("    %s%-18s%s%s" % (BOLD, cm + alias, ENDC, short_docs[cm]))  # bold(...) can't be used here due to the %-18s format specifier
> > KeyError: 'hub'
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > https://github.com/asford/git-machete/blob/4f8387d778a4e5c87a3f06b4734b62a4fdf05a12/git_machete/cmd.py#L2729
> 
> Whoops good spot  I'll add a regression test for that after the weekend

Ok, fixed on PR #139

git_machete/cmd.py
```diff
@@ -3093,6 +3189,38 @@ def print_line_prefix(b_: str, suffix: str) -> None:
...
     """,
+    "hub": """
+        <b>Usage: git machete hub <subcommand></b>
+        where <subcommand> is one of: `anno-prs`, `create-pr`, `retarget-pr`.
+
+        Creates, checks out and manages GitHub PRs while keeping them reflected in branch definition file.
+
+        For all subcommands, to allow GitHub API access for private repositories
+        (and also to correctly identify the current user, even in case of public repositories),
+        `GITHUB_TOKEN` env var must contain a GitHub API token with `repo` scope, see `https://github.com/settings/tokens`.
```
PawelLipski: Update it with the changes done on PR #136

git_machete/cmd.py
```diff
@@ -648,31 +649,131 @@ def annotate(b: str, words: List[str]) -> None:
...
+            return ''
+
+    title: str = popen_git(cli_ctxt, "log", "-1", "--format=%s").strip()
+    description_path = get_git_subpath(cli_ctxt, 'info', 'description')
+    description: str = slurp_file_or_empty(description_path)
+
+    ok_str = ' <green><b>OK</b></green>'
+    print(fmt(f'Creating a {"draft " if draft else ""}PR from `{head}` to `{base}`...'), end='', flush=True)
+    pr: GitHubPullRequest = create_pull_request(org, repo, head=head, base=base, title=title, description=description, draft=draft)
+    print(fmt(f'{ok_str}, see <b>{pr.html_url}</b>'))
```
PawelLipski: Also:

```
Creating a PR from chore/tidy-readme-comparison-table to main...
GitHub API returned 307 HTTP status with error message: Moved Permanently
Make sure that the token provided in gh auth status or ~/.config/hub or GITHUB_TOKEN is valid and allows for access to POST https://api.github.com/repos/VirtusLab/akka-safer-serializer/pulls`.
```

for a Github repo that has been renamed... we should follow the redirect in such case rather than fail the entire flow.

asford:
FWIW, with the updates made in #143 I was able to implement the github integration I needed by combing the `gh` cli and `git machete` in simple-ish bash scripts.

I like having the full PR status, base branch and url in the machete annotation, so having flexibility in this part of the process is quite useful to me.

Definitely not saying there isn't a role for `machete` to provide github integration "out of the box", but I think it's worth considering whether it's better to provide it via "integration wrappers" of the type I'm using here rather than deeply integrating into machete proper.

#### `git-machete-pr-create` - opens a pull request via the `gh` cli targeting the machete base branch:

```bash
#!/usr/bin/env bash
#http://redsymbol.net/articles/unofficial-bash-strict-mode/

set -uo pipefail
# shopt -s globstar
# shopt -s extglob
IFS=$'\n\t'

gh pr create --base `git machete show up`

git machete anno $(gh pr view --json state,url,baseRefName --jq '"\(.state) \(.baseRefName) \(.url)"')
```

#### `git-machete-pr-update` - updates `base` pointers and annotations for all machete-tracked branches:
 
```bash
#!/usr/bin/env bash
#http://redsymbol.net/articles/unofficial-bash-strict-mode/

set -uo pipefail
# shopt -s globstar
# shopt -s extglob
IFS=$'\n\t'


for branch in $(git machete list slidable)
do
  base_ref=$(gh pr view $branch --json baseRefName --jq .baseRefName)

  if [ "$base_ref" == "" ]; then
    continue
  fi

  up_branch=$(git machete show up $branch)

  if [ "$base_ref" != "$up_branch" ]; then
    echo gh pr edit $branch --base $up_branch
    gh pr edit $branch --base $up_branch
  fi

  git machete anno --branch=$branch $(gh pr view $branch --json state,url,baseRefName --jq '"\(.state) \(.baseRefName) \(.url)"')
done
```

PawelLipski:
> FWIW, with the updates made in #143 I was able to implement the github integration I needed by combing the `gh` cli and `git machete` in simple-ish bash scripts.

Heh TBH I've done similar wrappers (`hub`-based, tho) myself: https://github.com/PawelLipski/utils/blob/master~30/bashrc-append.sh#L225

Although still, I'd like to provide some default set of most common operations out of the box even for the less GitHub-CLI-savvy users :)

git_machete/cmd.py
rifel123: Current library http.client used for requests seems to support redirections only if it gets status code 302 in response.
Let's consider using urllib instead in this case, which has built in support of redirections for codes 301, 302, 303, or 307.

PawelLipski: Whoops... I knew that `http.client` is a library of _plywood and cardboard_ (aka _z dykty i kartonu_), but didn't know it's that bad 😅 yeah, if you could migrate the existing impl to urllib then... as long as we don't add any external dependencies I'm fine.

CONTRIBUTING.md
```diff
@@ -21,10 +21,10 @@ Deprecated commands are excluded.
...
-| can accept interactive input on stdin                             | `add`, `advance`, `delete-unmanaged`, `discover`, `go`, `traverse`, `update`                       |
-| can display status (and run `machete-status-branch` hook)         | `discover`, `status`, `traverse`                                                                   |
-| can modify the .git/machete file                                  | `add`, `advance`, `anno`, `discover`, `edit`, `slide-out`, `traverse`                              |
-| can modify the git repository (excluding .git/machete)            | `add`, `advance`, `delete-unmanaged`, `go`, `reapply`, `slide-out`, `squash`, `traverse`, `update` |
+| can accept interactive input on stdin                             | `add`, `advance`, `delete-unmanaged`, `discover`, `go`, `traverse`, `update`, `gh` (\*\*\*\*)     |
+| can display status (and run `machete-status-branch` hook)         | `discover`, `status`, `traverse`, `gh` (\*\*\*\*)                                                 |
```
PawelLipski: Nit: pls keep commands in abc order in each row

git_machete/github.py
```diff
@@ -123,6 +123,8 @@ def fire_github_api_get_request(method: str, url: str, token: Optional[str], req
 
         if 200 <= response.status < 300:
             return parsed_response_body
+        elif response.status == 422 and parsed_response_body.get("message") == "Validation Failed":
+            raise MacheteException("Cannot create pull request on untracked branch. \nPlease push your changes before making pull request.")
```
PawelLipski: Hmmm putting this specific message to the generic `fire_github_api_get_request` method is questionable... note that this method is used not only for creating PRs, but also for a few other types of requests.

In general, I'd rather throw a dedicated exception from `fire_github_api_get_request`, to be caught & interpreted by each high-level function.

git_machete/github.py
```diff
@@ -123,6 +123,8 @@ def fire_github_api_get_request(method: str, url: str, token: Optional[str], req
 
         if 200 <= response.status < 300:
             return parsed_response_body
```
PawelLipski, line 124: While you're here... `fire_github_api_get_request` should probably be made private

git_machete/github.py
```diff
@@ -148,8 +150,22 @@ def create_pull_request(org: str, repo: str, head: str, base: str, title: str, d
...
+    prs: List[GitHubPullRequest] = derive_pull_requests(org, repo)
+    to_load: GitHubPullRequest = GitHubPullRequest(1, 'user', base, head, '')
+    if not is_pr_already_created(to_load, prs):
+        pr = fire_github_api_get_request('POST', f'/repos/{org}/{repo}/pulls', token, request_body)
+        return parse_pr_json(pr)
+    else:
+        raise MacheteException('Given Pull Request is already created!')
+
+
+def is_pr_already_created(pull: GitHubPullRequest, pull_requests: List[GitHubPullRequest]) -> bool:
```
PawelLipski: Same here, private (TBD if the prefix should be `_` or `__`, I'm not certain what is the proper convention for module-private functions, as opposed to private class methods)

git_machete/github.py
```diff
@@ -148,8 +150,22 @@ def create_pull_request(org: str, repo: str, head: str, base: str, title: str, d
...
+    if not pull_requests:
+        return False
+    for pr in pull_requests:
+        if all((pull.base == pr.base, pull.head == pr.head)):
```
PawelLipski: Hmmm just
```suggestion
        if pull.base == pr.base and pull.head == pr.head:
```
?

git_machete/github.py
```diff
@@ -148,8 +150,22 @@ def create_pull_request(org: str, repo: str, head: str, base: str, title: str, d
...
+    to_load: GitHubPullRequest = GitHubPullRequest(1, 'user', base, head, '')
+    if not is_pr_already_created(to_load, prs):
+        pr = fire_github_api_get_request('POST', f'/repos/{org}/{repo}/pulls', token, request_body)
+        return parse_pr_json(pr)
+    else:
+        raise MacheteException('Given Pull Request is already created!')
+
+
+def is_pr_already_created(pull: GitHubPullRequest, pull_requests: List[GitHubPullRequest]) -> bool:
+    if not pull_requests:
```
PawelLipski: This check is redundant

git_machete/github.py
```diff
@@ -148,8 +150,22 @@ def create_pull_request(org: str, repo: str, head: str, base: str, title: str, d
...
+        return False
+    for pr in pull_requests:
+        if all((pull.base == pr.base, pull.head == pr.head)):
+            return True
+    return False
```
PawelLipski: Also, the above loop can be simplified to `any` + lambda

git_machete/github.py
```diff
@@ -148,8 +150,22 @@ def create_pull_request(org: str, repo: str, head: str, base: str, title: str, d
...
     }
-    pr = fire_github_api_get_request('POST', f'/repos/{org}/{repo}/pulls', token, request_body)
-    return parse_pr_json(pr)
+    prs: List[GitHubPullRequest] = derive_pull_requests(org, repo)
+    to_load: GitHubPullRequest = GitHubPullRequest(1, 'user', base, head, '')
+    if not is_pr_already_created(to_load, prs):
+        pr = fire_github_api_get_request('POST', f'/repos/{org}/{repo}/pulls', token, request_body)
+        return parse_pr_json(pr)
+    else:
+        raise MacheteException('Given Pull Request is already created!')
```
PawelLipski:
```suggestion
        raise MacheteException('Pull request for branch {head} is already created!')
```
Plus I'd also maybe display a link to the PR in question (so `is_pr_already_created` maybe should return a `GitHubPullRequest` object instead of a bool)

git_machete/cmd.py
```diff
@@ -3861,9 +3936,9 @@ def check_required_param(in_args: List[str], allowed_values_string: str) -> str:
             param = check_optional_param(parse_options(args))
             # No need to read definition file.
             usage(param)
-        elif cmd == "hub":
-            hub_allowed_values = "anno-prs|create-pr|retarget-pr"
-            param = check_required_param(parse_options(args, "", ["draft"]), hub_allowed_values)
+        elif cmd == "gh":
```
PawelLipski: Judging by how `gh`'s CLI looks like (https://cli.github.com/manual/index), I'd rather expect a slightly different set of subcommands (sth like `pr create`, `pr anno`, `pr retarget`)... although that's to be discussed.

`git machete gh pr create` would follow the conventions of the original `gh`, but at the same time require an extra rather redundant argument to the command line... maybe `git machete gh-pr create` would be a good compromise? 🤔

git_machete/cmd.py
```diff
@@ -3242,15 +3313,19 @@ def print_line_prefix(b_: str, suffix: str) -> None:
...
+        <b>Usage: git machete gh <subcommand></b>
         where <subcommand> is one of: `anno-prs`, `create-pr`, `retarget-pr`.
 
         Creates, checks out and manages GitHub PRs while keeping them reflected in branch definition file.
-
+        
         For all subcommands, to allow GitHub API access for private repositories
         (and also to correctly identify the current user, even in case of public repositories),
         `GITHUB_TOKEN` env var must contain a GitHub API token with `repo` scope, see `https://github.com/settings/tokens`.
+        This will be resolved from the first of:
```
PawelLipski: Whoops these lines don't follow well after the line 3324 above

PawelLipski: `GITHUB_TOKEN` is just one of three possible auth methods

git_machete/cmd.py
```diff
@@ -803,6 +806,74 @@ def slurp_file_or_empty(path: str) -> str:
     save_definition_file()
 
 
+def sync_before_pr(cli_ctxt: CommandLineContext) -> None:
```
PawelLipski: Hmmm actually... if you could prioritize rebasing your current changes onto current develop, for two reasons:
1. the more changes in more places you introduce on this PR, the harder it'll be to rebase later
2. once the state of `feature/hub` branch includes both `gh` CLI and your recent refactors, I can [dogfood](https://en.wikipedia.org/wiki/Eating_your_own_dog_food) both at the same time on my commercial project without shuttling b/w `develop` and `feature/hub` and reinstalling

git_machete/cmd.py
```diff
@@ -754,6 +754,9 @@ def create_github_pr(cli_ctxt: CommandLineContext, head: str, draft: bool) -> No
     from git_machete.github import GitHubPullRequest, add_assignees_to_pull_request, add_reviewers_to_pull_request, \
         create_pull_request, derive_current_user_login, set_milestone_of_pull_request
 
+    # first make sure that head branch is synced with remote
+    sync_before_pr(cli_ctxt)
```
PawelLipski: Nit:
```suggestion
    sync_before_creating_pr(cli_ctxt)
```

git_machete/cmd.py
```diff
@@ -803,6 +806,74 @@ def slurp_file_or_empty(path: str) -> str:
...
+
+    if is_merged_to_upstream(cli_ctxt, cb):
+        return
+    s, remote = get_strict_remote_sync_status(cli_ctxt, cb)
+    statuses_to_sync = (UNTRACKED, AHEAD_OF_REMOTE)
+    needs_remote_sync = s in statuses_to_sync
+
+    if needs_remote_sync:
+        print_new_line(False)
+        sys.stdout.write(f"Checking out {bold(cb)}\n")
```
PawelLipski: Hmmm is checkout needed? There's an assumption that `cb` is already the current branch 🤔

git_machete/cmd.py
PawelLipski: Also, maybe actually it'd be most reasonable to ditch following either `hub` or `gh` CLI completely, and name our command like `git machete github` to avoid any association with either existing CLI 🤔 then we'll have zero need to be concerned about stuff is called/organized in the existing tools

README.md
```diff
@@ -149,12 +149,23 @@ git machete advance
 Useful for merging the child branch to the current branch in a linear fashion (without creating a merge commit).
 
 ### GitHub integration
+
+Annotate the branches with GitHub PR numbers: <br/>
+```shell script
+git machete hub anno-prs
```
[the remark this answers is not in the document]
PawelLipski: ...or as discussed somewhere below, don't stick to either `hub` or `gh` at all... TBD

README.md
```diff
@@ -149,11 +149,21 @@ git machete advance
...
+```
+
+Check out PR and reflect its parent (grandparent, etc.) in `.git/machete`: <br/>
```
PawelLipski: Whoops as discussed on video, that's not gonna be implemented on the this PR... actually if you open a separate issue for implementing `checkout-pr` (however it's gonna be called)

completion/git-machete.completion.bash
```diff
@@ -1,11 +1,13 @@
 #!/usr/bin/env bash
 
 _git_machete() {
-  local cmds="add advance anno d delete-unmanaged diff discover e edit file fork-point g go help is-managed l list log reapply s show slide-out squash status traverse update version"
+  local cmds="add advance anno d delete-unmanaged diff discover e edit file fork-point g go help hub is-managed l list log reapply s show slide-out squash status traverse update version"
```
PawelLipski: TODO: the potential change to the subcommand name should be reflected in the completion scripts

git_machete/docs.py
```diff
@@ -299,17 +300,9 @@
         <b>Usage: git machete g[o] <direction></b>
         where <direction> is one of: `d[own]`, `f[irst]`, `l[ast]`, `n[ext]`, `p[rev]`, `r[oot]`, `u[p]`
 
-        Checks out the branch specified by the given direction relative to the current branch:
```
PawelLipski, line 302: Whoops I wouldn't expect any change to `go`'s help on this PR, probably a rebase glitch

completion/git-machete.completion.zsh
```diff
@@ -209,6 +212,16 @@ __git_machete_directions_show() {
   _describe -t directions 'direction' directions "$@"
 }
 
+__git_machete_github_subcommands() {
+  local github_subcommands
+  github_subcommands=(
+    'anno-prs:annotate the branche based on their corresponding GitHub PR numbers and authors'
```
BartoszSiwinski:
```suggestion
    'anno-prs:annotate the branches based on their corresponding GitHub PR numbers and authors'
```

git_machete/cmd.py
PawelLipski: Added as https://github.com/VirtusLab/git-machete/issues/175

CONTRIBUTING.md
```diff
@@ -21,10 +21,10 @@ Deprecated commands are excluded.
...
 | ----------------------------------------------------------------- | -------------------------------------------------------------------------------------------------- |
-| can accept interactive input on stdin                             | `add`, `advance`, `delete-unmanaged`, `discover`, `go`, `traverse`, `update`                       |
-| can display status (and run `machete-status-branch` hook)         | `discover`, `status`, `traverse`                                                                   |
-| can modify the .git/machete file                                  | `add`, `advance`, `anno`, `discover`, `edit`, `slide-out`, `traverse`                              |
-| can modify the git repository (excluding .git/machete)            | `add`, `advance`, `delete-unmanaged`, `go`, `reapply`, `slide-out`, `squash`, `traverse`, `update` |
+| can accept interactive input on stdin                             | `add`, `advance`, `delete-unmanaged`, `discover`,`github` (\*\*\*\*),  `go`, `traverse`, `update`     |
```
PawelLipski:
```suggestion
| can accept interactive input on stdin                             | `add`, `advance`, `delete-unmanaged`, `discover`, `github` (\*\*\*\*),  `go`, `traverse`, `update`     |
```

CONTRIBUTING.md
```diff
@@ -21,10 +21,10 @@ Deprecated commands are excluded.
...
+| can display status (and run `machete-status-branch` hook)         | `discover`, `github` (\*\*\*\*), `status`, `traverse`                                                 |
+| can modify the .git/machete file                                  | `add`, `advance`, `anno`, `discover`, `edit`, `github`, `slide-out`, `traverse`         |
+| can modify the git repository (excluding .git/machete)            | `add`, `advance`, `delete-unmanaged`, `github` `go`, `reapply`, `slide-out`, `squash`, `traverse`, `update` |
```
PawelLipski:
```suggestion
| can modify the git repository (excluding .git/machete)            | `add`, `advance`, `delete-unmanaged`, `github`, `go`, `reapply`, `slide-out`, `squash`, `traverse`, `update` |
```

PawelLipski: Also, does `(****)` apply here as well?

git_machete/docs.py
```diff
@@ -298,19 +299,50 @@
...
+        a GitHub API token with `repo` scope is required, see `https://github.com/settings/tokens`. This will be resolved from the first of:
+        1. `GITHUB_TOKEN` env var,
+        2. current auth token from the `gh` GitHub CLI,
+        3. current auth token from the `hub` GitHub CLI.
+
+        <b>`anno-prs`:</b>
+
+          Annotates the branches based on their corresponding GitHub PR numbers and authors.
+          Any existing annotations are overwritten for the branches that have an opened PR; annotations for the other branches remain untouched.
+          Equivalent to `git machete anno --sync-github-prs`.
```
BartoszSiwinski, line 330: Proposal: Is either `git machete github anno-prs` or `git machete anno --sync-github-prs` recommended? Maybe it would be possible/a good idea to deprecate one of the ways. It could happen for 4.0.0 version, but we could create a file/issue to store things that shall be removed in the future and in version 3.4.0 we could introduce deprecation warning.

CONTRIBUTING.md
```diff
@@ -37,6 +37,8 @@ Deprecated commands are excluded.
 
 (\*\*\*) A stable output is only guaranteed for `fork-point` when invoked without any option or only with `--inferred` option.
 
+(\*\*\*\*) `github` can display status and accept interactive mode only when `create-pr` subcommand is passed and the current branch is in untracked or ahead of origin state.
```
PawelLipski: Nit:
```suggestion
(\*\*\*\*) `github` can display status and accept interactive mode only when `create-pr` subcommand is passed and the current branch is untracked or ahead of its remote.
```

README.md
```diff
@@ -149,11 +149,21 @@ git machete advance
...
+Sets the base of the current branch's PR to upstream (parent) branch, as seen by git machete: <br/>
 ```shell script
-git machete anno --sync-github-prs
+git machete github retarget-pr
 ```
 
-This will automatically annotate the branches with GitHub PR numbers. <br/>
+Create the PR, using the parent branch from `.git/machete` as the base: <br/>
+```shell script
+git machete github create-pr [--draft]
```
PawelLipski: Nit: I'd swap `create-pr` with `retarget-pr` here

completion/git-machete.completion.bash
```diff
@@ -70,10 +79,30 @@ _git_machete() {
...
+                else
+                  COMPREPLY=('')
+                fi ;;
+              help)
+                if [[ $COMP_CWORD -eq 3 ]]; then
+                  __gitcomp "$help_topics"
+                else
+                  COMPREPLY=('')
+                fi ;;
+              github)
```
PawelLipski, line 92: Nit: switch with `help` for abc order

git_machete/docs.py
```diff
@@ -13,6 +13,7 @@
     "fork-point": "Display or override fork point for a branch",
     "format": "Display docs for the format of the definition file",
     "go": "Check out the branch relative to the position of the current branch, accepts down/first/last/next/root/prev/up argument",
+    "github": "Create, retarget and manage GitHub PRs while keeping them reflected in git machete",
```
PawelLipski, line 15: Switch with `go` for abc order

git_machete/docs.py
```diff
@@ -298,19 +299,50 @@
     "go": """
         <b>Usage: git machete g[o] <direction></b>
         where <direction> is one of: `d[own]`, `f[irst]`, `l[ast]`, `n[ext]`, `p[rev]`, `r[oot]`, `u[p]`
-
```
PawelLipski, line 301: Restore the empty lines for better readability/consistency with the help for other commands

git_machete/docs.py
PawelLipski: I'd say `git machete github anno-prs` is recommended... not sure about deprecating `git machete anno --sync-github-prs` tho 🤔

git_machete/docs.py
```diff
@@ -423,9 +455,7 @@
         <b>Usage: git machete show <direction> [<branch>]</b>
         where <direction> is one of: `c[urrent]`, `d[own]`, `f[irst]`, `l[ast]`, `n[ext]`, `p[rev]`, `r[oot]`, `u[p]`
         displayed relative to given <branch>, or the current checked out branch if <branch> is unspecified.
-
```
PawelLipski: Restore

git_machete/github.py
```diff
@@ -90,73 +97,131 @@ def _token_from_hub() -> Optional[str]:
...
-        response: HTTPResponse = conn.getresponse()
-        body: Any = json.loads(response.read().decode())
-
-        if 200 <= response.status < 300:
-            return body
+        with urlopen(http_request) as response:
+            parsed_response_body: Any = json.loads(response.read().decode())
+            return parsed_response_body
+    except HTTPError as err:
+        if err.code == 422:
```
BartoszSiwinski: You could replace magic number 422 with that:
```
import http
http.HTTPStatus.UNPROCESSABLE_ENTITY
<HTTPStatus.UNPROCESSABLE_ENTITY: 422>
```

git_machete/exceptions.py
```diff
@@ -12,3 +12,7 @@ def __str__(self) -> str:
 class StopTraversal(Exception):
     def __init__(self) -> None:
         pass
+
+
+class UnprocessableEntityError(Exception):
```
PawelLipski: I'd specify:
```suggestion
class UnprocessableHttpEntityError(Exception):
```

git_machete/options.py
```diff
@@ -34,3 +34,4 @@ def __init__(self) -> None:
         self.opt_sync_github_prs: bool = False
         self.opt_unset_override: bool = False
         self.opt_yes: bool = False
+        self.opt_draft: bool = False
```
PawelLipski: Move up for abc order

git_machete/github.py
PawelLipski: Nit: does urrlib provide some constant instead, like `UNPROCESSABLE_ENTITY`? I mean, 422 is relatively obscure code compared to e.g. 400 or 404, I can imagine it might get confused one day once this `if` grows

PawelLipski: (oh okay, didn't see the one above)

PawelLipski: Ideally if it were imported from `urllib` to use just one library or everything... or is `urllib` a wrapper over `http` anyway?

git_machete/github.py
```diff
@@ -90,73 +97,131 @@ def _token_from_hub() -> Optional[str]:
...
-def parse_github_remote_url(url: str) -> Optional[Tuple[str, str]]:
-
+def parse_github_remote_url(github_url: str) -> Tuple[str, str]:
     for pattern in GITHUB_REMOTE_PATTERNS:
-        match = re.match(pattern, url)
+        match = re.match(pattern, github_url)
         if match:
             return match.group(1), match.group(2)
-    return None
+    raise MacheteException(f'{github_url} is not a GitHub remote URL')
```
PawelLipski: Hmmm judging by how `parse_github_remote_url` is used currently... should really an exception be thrown here? 🤔 what if a local repository has a few remotes, some of which are GitHub and some aren't?

git_machete/exceptions.py
BartoszSiwinski: Is this error needed? There is in reality no error like that, there is just generic HTTPError that have some error code inside. By extracting this status code as separate exception we're loosing the context of HTTP Error. As it is now it's hard to figure out what this error is about (what is entity? database entry or maybe element of an array?).

I propose following options:
1. Add docstring specifying that this is HTTP Error with status code 422. Ideally also add context to this exception by either putting into the file `http_errors.py` or by storing this exception in some dict (`http_specific_errors`?) and than referring to it by `raise http_specific_errors[http.HTTPStatus.UNPROCESSABLE_ENTITY]`
2. Don't implement this exception and just use HTTP Error instead with appropriate msg etc.

git_machete/github.py
rifel123: unfortunatelly it's a wrapper around http.client

PawelLipski: Okay, then using a constant from `http` is justified

git_machete/exceptions.py
BartoszSiwinski: I would suggest rather not splitting Unprocessable - Entity and Http - Error pairs since when coupled together they have special meaning. So either UnprocessableEntityHttpError or HttpErrorUnprocessableEntity.

git_machete/docs.py
rifel123: lets keep both for now

CONTRIBUTING.md
```diff
@@ -37,6 +37,8 @@ Deprecated commands are excluded.
 
 (\*\*\*) A stable output is only guaranteed for `fork-point` when invoked without any option or only with `--inferred` option.
 
+(\*\*\*\*) `github` can display status and accept interactive mode only when `create-pr` subcommand is passed and the current branch is untracked or ahead of its remote.
```
PawelLipski:
```suggestion
(\*\*\*\*) `github` can only display status, accept interactive mode or modify git repository when `create-pr` subcommand is passed and the current branch is untracked or ahead of its remote.
```

PawelLipski: (at least I think the above holds true!)

README.md
```diff
@@ -149,11 +149,21 @@ git machete advance
...
+```
+
+Create the PR, using the parent branch from `.git/machete` as the base: <br/>
 ```shell script
-git machete anno --sync-github-prs
+git machete github create-pr [--draft]
 ```
 
-This will automatically annotate the branches with GitHub PR numbers. <br/>
+Sets the base of the current branch's PR to upstream (parent) branch, as seen by git machete: <br/>
```
PawelLipski:
```suggestion
Sets the base of the current branch's PR to its upstream (parent) branch, as seen by git machete: <br/>
```

README.md
```diff
@@ -149,11 +149,21 @@ git machete advance
 Useful for merging the child branch to the current branch in a linear fashion (without creating a merge commit).
 
 ### GitHub integration
+
+Annotate the branches with GitHub PR numbers: <br/>
+```shell script
+git machete github anno-prs
+```
+
+Create the PR, using the parent branch from `.git/machete` as the base: <br/>
```
PawelLipski:
```suggestion
Create the PR, using the upstream (parent) branch from `.git/machete` as the base: <br/>
```

PawelLipski: The reason why I still keep this double naming (_up/downstream_, _parent/child_) is that:
* the former works better with "navigation" in `go`/`show` commands (it's more natural to _go up_ rather than _go (to) parent_ BUT
* the latter is more established nomenclature when speaking about trees in theoretical CS/math.

CC @BartoszSiwinski

PawelLipski: Neither is free of naming conflicts tho:
* _upstream_ (although not _downstream_) is used by git CLI to mean _remote for the given branch_, see https://virtuslab.github.io/tips/#git/upstream
* _parent_ (although not _child_) is used for _parent commits_, as in e.g. the output of `git cat-file -p HEAD`.

git_machete/docs.py
```diff
@@ -300,17 +301,49 @@
...
         Checks out the branch specified by the given direction relative to the current branch:
-
         * `down`:    the direct children/downstream branch of the current branch.
         * `first`:   the first downstream of the root branch of the current branch (like `root` followed by `next`), or the root branch itself if the root has no downstream branches.
         * `last`:    the last branch in the definition file that has the same root as the current branch; can be the root branch itself if the root has no downstream branches.
         * `next`:    the direct successor of the current branch in the definition file.
         * `prev`:    the direct predecessor of the current branch in the definition file.
         * `root`:    the root of the tree where the current branch is located. Note: this will typically be something like `develop` or `master`, since all branches are usually meant to be ultimately merged to one of those.
         * `up`:      the direct parent/upstream branch of the current branch.
-
```
PawelLipski, line 311: Restore

git_machete/docs.py
```diff
@@ -425,7 +458,6 @@
         displayed relative to given <branch>, or the current checked out branch if <branch> is unspecified.
 
         Outputs name of the branch (or possibly multiple branches, in case of `down`) that is:
-
```
PawelLipski: Restore

git_machete/exceptions.py
```diff
@@ -12,3 +12,10 @@ def __str__(self) -> str:
 class StopTraversal(Exception):
     def __init__(self) -> None:
         pass
+
+
+class UnprocessableEntityHTTPError(Exception):
+    """This exception is raised when Github API returns HTTP status code 422 - Unprocessable Entity.
+    Such a situation occurs when trying to do something not allowed by github, egz. assign as a reviewer someone from outside organization or create a pull request which is already created.
```
PawelLipski:
```suggestion
    Such a situation occurs when trying to do something not allowed by GitHub,
    e.g. assigning someone from outside organization as a reviewer
    or creating a pull request for a branch that already has a PR.
```

git_machete/exceptions.py
```diff
@@ -12,3 +12,10 @@ def __str__(self) -> str:
...
+    """
+    pass
```
PawelLipski: `def __init__` ? not sure if there's any difference for exception classes whether the c'tor is present or not

PawelLipski: (either way, pls keep consistent across all exception classes)

git_machete/github.py
```diff
@@ -6,24 +6,32 @@
 import shutil
 import subprocess
 # Deliberately NOT using much more convenient `requests` to avoid external dependencies
-from http.client import HTTPResponse, HTTPSConnection
+from urllib.request import Request, urlopen
+from urllib.error import HTTPError
+import http
```
PawelLipski: Nit: keep imports in abc order (probably https://github.com/VirtusLab/git-machete/issues/140 will keep them sorted anyway)

git_machete/github.py
```diff
@@ -6,24 +6,32 @@
...
         self.user = user
         self.base = base
         self.head = head
+        self.html_url = html_url
 
     def __repr__(self) -> str:
         return f"PR #{self.number} by {self.user}: {self.head} -> {self.base}"
 
 
+def parse_pr_json(pr_json: Any) -> GitHubPullRequest:
```
PawelLipski: Private?

git_machete/cmd.py
```diff
@@ -1570,15 +1573,174 @@ def flush_caches(self) -> None:
...
+                f'Branch {head} does not have a parent branch (it is a root), base branch for the PR cannot be established')
+
+        org: str
+        repo: str
+        _, (org, repo) = self.__derive_remote_and_github_org_and_repo()
+        current_user: Optional[str] = derive_current_user_login()
+        debug(f'create_github_pr({head})', f'organization is {org}, repository is {repo}')
+        debug(f'create_github_pr({head})', 'current GitHub user is ' + (current_user or '<none>'))
+
+        def slurp_file_or_empty(path: str) -> str:
```
PawelLipski: Maybe move to `utils` + also I think `read()` call should be wrapped into `with` (so that `close()` is always called, even in case of an error)

git_machete/github.py
```diff
@@ -90,71 +98,129 @@ def _token_from_hub() -> Optional[str]:
...
+        'draft': draft
+    }
+    prs: List[GitHubPullRequest] = derive_pull_requests(org, repo)
+    to_load: GitHubPullRequest = GitHubPullRequest(1, 'user', base, head, '')
+    pr_found: Optional[GitHubPullRequest] = __check_pr_already_created(to_load, prs)
+    if not pr_found:
+        pr = __fire_github_api_request('POST', f'/repos/{org}/{repo}/pulls', token, request_body)
+        return parse_pr_json(pr)
+    else:
+        raise MacheteException(f'Pull request for branch {head} is already created under link {pr_found.html_url}! PR details:\n{pr_found}')
```
PawelLipski: Hmmm actually I looked at the output here:

image

I'd change two things:
1. I don't think there's a need to print the status if no push was needed,
2. PR details are probably redundant here (or at least `\n` should go BEFORE `PR details`)

git_machete/github.py
PawelLipski: \+ other methods in this module

git_machete/cmd.py
```diff
@@ -1570,15 +1573,174 @@ def flush_caches(self) -> None:
...
+        self.save_definition_file()
+
+    def __sync_before_creating_pr(self) -> None:
+
+        self.expect_at_least_one_managed_branch()
+        self.__empty_line_status = True
+
+        cb = self.__git.current_branch()
+        try:
+            self.expect_in_managed_branches(cb)
```
PawelLipski: Hmmm there's a call to `machete_client.expect_in_managed_branches(cb)` already in the main function... so IIUC we won't even get to this method unless current branch is listed in .git/machete aka _managed_

git_machete/cmd.py
```diff
@@ -1570,15 +1573,174 @@ def flush_caches(self) -> None:
...
+        except MacheteException:
+            self.add(cb)
+
+        if self.__is_merged_to_upstream(cb):
+            return
+
+        s, remote = self.__git.get_strict_remote_sync_status(cb)
+        statuses_to_sync = (UNTRACKED, AHEAD_OF_REMOTE)
```
PawelLipski: Know what, you were right... I'm using this new features in my commercial project, and a force-push in case of `DIVERGED_FROM_AND_NEWER_THAN_REMOTE` would be clearly useful

PawelLipski: I'm not sure what to with `BEHIND_REMOTE` (`traverse` suggests `git pull`) or `DIVERGED_FROM_AND_OLDER_THAN_REMOTE` (`traverse` suggests `git reset --keep`)... probably displaying a warning to the user and proceeding with PR creation will be the most adequate action (?)

git_machete/exceptions.py
rifel123: __init__ is redundant here, because the whole logic is inherited from Exception. For the consistency, let's pass message to c'tor explicitly.

git_machete/cmd.py
rifel123: whoops, good spot :)

rifel123: Anyway, i'll remove this check from the main function. IMO it's better to ask user to add new branch automatically rather than stop here and tell user to do it manually.

git_machete/cmd.py
```diff
@@ -632,75 +632,20 @@ def traverse(self) -> None:
...
-                        self.handle_untracked_branch(rmt, b)
-                    elif len(rems) == 1:
-                        self.handle_untracked_branch(rems[0], b)
-                    elif "origin" in rems:
-                        self.handle_untracked_branch("origin", b)
-                    else:
-                        # We know that there is at least 1 remote, otherwise 's' would be 'NO_REMOTES'
-                        print(fmt(f"Branch `{bold(b)}` is untracked and there's no `{bold('origin')}` repository."))
-                        self.__pick_remote(b)
+                try:
```
PawelLipski, line 635: image

git_machete/cmd.py
```diff
@@ -1676,6 +1619,90 @@ def slurp_file_or_empty(path: str) -> str:
         self.__annotations[head] = f'PR #{pr.number}'
         self.save_definition_file()
 
+    def __handle_diverged_and_newer_state(self, cb: str, remote: str, abort: bool = True) -> None:
```
PawelLipski: Hmmm maybe the param should be called `abort_on_quit` for clarity?

git_machete/cmd.py
```diff
@@ -1676,6 +1619,90 @@ def slurp_file_or_empty(path: str) -> str:
...
+            f"Push {bold(cb)} with force-with-lease to {bold(remote)}?" + pretty_choices('y', 'N', 'q', 'yq'),
+            f"Branch {bold(cb)} diverged from (and has newer commits than) its remote counterpart {bold(rb)}.\n"
+            f"Pushing {bold(cb)} with force-with-lease to {bold(remote)}...",
+            override_answer=None if self.__cli_opts.opt_push_tracked else "N")
+        if ans in ('y', 'yes', 'yq'):
+            self.__git.push(remote, cb, force_with_lease=True)
+            if ans == 'yq':
+                if not abort:
+                    return
+                raise MacheteException('Process interrupted.')
```
PawelLipski: Hmm I think displaying such a message to the user might sound a bit too... _stiff_? technical?

PawelLipski: Consider using `StopTraversal` exception, it's been created for the purpose of returning a few methods when quitting `traverse`

git_machete/cmd.py
```diff
@@ -1691,46 +1718,35 @@ def __sync_before_creating_pr(self) -> None:
...
-                elif "origin" in rems:
-                    self.handle_untracked_branch("origin", cb)
-                else:
-                    # We know that there is at least 1 remote, otherwise 's' would be 'NO_REMOTES'
-                    print(fmt(f"Branch `{bold(cb)}` is untracked and there's no `{bold('origin')}` repository."))
-                    self.__pick_remote(cb)
-
-        self.__print_new_line(False)
-        self.status(warn_on_yellow_edges=True)
+                ans = self.ask_if("Proceed with Pull Request creation?" + pretty_choices('y', 'N', 'q', 'yq'),
```
PawelLipski: I'd rather use lower-case `pull request` everywhere

git_machete/cmd.py
rifel123: I agree, this is too technical, but on the other hand, would be perfect if the message here fits both contexts: `traverse` and `__sync_before_creating_pr`.  
BTW don't you think that wee need specialized exception here as issue #164 says?

PawelLipski: > I agree, this is too technical, but on the other hand, would be perfect if the message here fits both contexts: traverse and __sync_before_creating_pr.

Okay, but how about just displaying NO message? I think that would also fit both contexts 🤔

PawelLipski: > BTW don't you think that wee need specialized exception here as issue #164 says?

As in, an exception different than `StopTraversal`, but with the same effect on the execution? 🤔

git_machete/cmd.py
```diff
@@ -1631,14 +1630,14 @@ def __handle_diverged_and_newer_state(self, cb: str, remote: str, abort: bool =
         if ans in ('y', 'yes', 'yq'):
             self.__git.push(remote, cb, force_with_lease=True)
             if ans == 'yq':
-                if not abort:
+                if not abort_on_quit:
```
PawelLipski: Maybe to make the logic here a bit clearer:
```
if abort_on_quit:
  raise StopTraversal
```

PawelLipski: and similarly in other places

completion/git-machete.completion.zsh
```diff
@@ -67,6 +67,9 @@ _git-machete() {
         (help)
           _arguments '1:: :__git_machete_help_topics' && ret=0
           ;;
+        (github)
```
PawelLipski, line 67: Move up fo abc order

git_machete/cmd.py
PawelLipski: Ok! resolved?

git_machete/github.py
PawelLipski: Hmmm okay so is there any special handling of 401/403 status now? 🤔

git_machete/exceptions.py
PawelLipski: (resolved?)

git_machete/cmd.py
rifel123: yes, by removing `machete_client.expect_in_managed_branches(cb)` right before `machete_client.create_github_pr(cb, draft=cli_opts.opt_draft)`

git_machete/cmd.py
PawelLipski: (`return` will be probably unneeded then?)

git_machete/exceptions.py
rifel123: Since context has been added to this exception, i think so, yes. @BartoszSiwinski, Could you confirm?

git_machete/github.py
rifel123: No, there isn't. But is there something we can still do in case of 401/403 Status?

PawelLipski: I'd put it in a different way: the existing error messages (`Make sure that the token provided ...` and `This repository might be private. Provide a GitHub API token...`) only make sense in case of 401/403 status.

PawelLipski: In case of a different status &mdash; it's hard to provide a ready response, it might be better to just display the details with http status + response, and even maybe a link to create an issue against git-machete?

git_machete/cmd.py
PawelLipski: @rifel123 could you take a look? this can be probably extracted to a separate issue

git_machete/cmd.py
```diff
@@ -1570,15 +1517,241 @@ def flush_caches(self) -> None:
...
+                warn(f"Branch {cb} is in <b>BEHIND_REMOTE</b> state.\nConsider using 'git pull'.\n")
+                self.__print_new_line(False)
+                ans = self.ask_if("Proceed with pull request creation?" + pretty_choices('y', 'N', 'q', 'yq'),
+                                  "Proceeding with pull request creation...")
+            elif s == DIVERGED_FROM_AND_OLDER_THAN_REMOTE:
+                warn(f"Branch {cb} is in <b>DIVERGED_FROM_AND_OLDER_THAN_REMOTE</b> state.\nConsider using 'git reset --keep'.\n")
+                self.__print_new_line(False)
+                ans = self.ask_if("Proceed with pull request creation?" + pretty_choices('y', 'N', 'q', 'yq'),
+                                  "Proceeding with pull request creation...")
+            if ans in ('y', 'yes', 'yq'):
```
PawelLipski, line 1744:
```$ git machete github create-pr 
  File "/usr/local/bin/git-machete", line 4, in <module>
    __import__('pkg_resources').run_script('git-machete==3.4.0', 'git-machete')
  File "/usr/lib/python3.7/site-packages/pkg_resources/__init__.py", line 666, in run_script
    self.require(requires)[0].run_script(script_name, ns)
  File "/usr/lib/python3.7/site-packages/pkg_resources/__init__.py", line 1469, in run_script
    exec(script_code, namespace, namespace)
  File "/usr/local/lib/python3.7/site-packages/git_machete-3.4.0-py3.7.egg/EGG-INFO/scripts/git-machete", line 4, in <module>
    __import__('pkg_resources').run_script('git-machete==3.4.0', 'git-machete')
  File "/usr/local/lib/python3.7/site-packages/git_machete-3.4.0-py3.7.egg/git_machete/cmd.py", line 1812, in main
  File "/usr/local/lib/python3.7/site-packages/git_machete-3.4.0-py3.7.egg/git_machete/cmd.py", line 2075, in launch
  File "/usr/local/lib/python3.7/site-packages/git_machete-3.4.0-py3.7.egg/git_machete/cmd.py", line 1570, in create_github_pr
  File "/usr/local/lib/python3.7/site-packages/git_machete-3.4.0-py3.7.egg/git_machete/cmd.py", line 1741, in __sync_before_creating_pr
UnboundLocalError: local variable 'ans' referenced before assignment
```

git_machete/github.py
rifel123: The message to open an issue will also be included in UnprocessableEntity error, as it's considered as bug.

git_machete/exceptions.py
```diff
@@ -21,7 +21,7 @@ class UnprocessableEntityHTTPError(Exception):
     or creating a pull request for a branch that already has a PR.
     """
     def __init__(self, msg: str) -> None:
-        self.msg: str = msg
+        self.msg: str = msg + "\nPlease open an issue regaring this topic under link: https://github.com/VirtusLab/git-machete/issues/new"
```
PawelLipski: `regarding`

git_machete/github.py
```diff
@@ -125,14 +125,17 @@ def __fire_github_api_request(method: str, url: str, token: Optional[str], reque
...
             first_line = fmt(f'GitHub API returned {err.code} HTTP status with error message: `{err.reason}`\n')
             if token:
                 raise MacheteException(first_line + fmt(f'Make sure that the token provided in `gh auth status` or `~/.config/hub` or <b>{GITHUB_TOKEN_ENV_VAR}</b> is valid and allows for access to `{method.upper()}` https://{host}{url}`.'))
             else:
                 raise MacheteException(
                     first_line + fmt(f'This repository might be private. Provide a GitHub API token with `repo` access via `gh` or `hub` or <b>{GITHUB_TOKEN_ENV_VAR}</b> env var.\n'
                                      'Visit `https://github.com/settings/tokens` to generate a new one.'))
+        else:
+            first_line = fmt(f'GitHub API returned {err.code} HTTP status with error message: `{err.reason}`\n')
+            raise MacheteException(first_line + "Please open an issue regaring this topic under link: https://github.com/VirtusLab/git-machete/issues/new")
```
PawelLipski: `regarding`

git_machete/cmd.py
```diff
@@ -1570,15 +1517,246 @@ def flush_caches(self) -> None:
...
+
+        new_base: Optional[str] = self.up_branch.get(head)
+        if not new_base:
+            raise MacheteException(f'Branch `{head}` does not have a parent branch (it is a root) '
+                                   f'even though there is an open PR #{pr.number} to `{pr.base}`.\n'
+                                   f'Consider modifying the branch definition file (`git machete edit`) so that `{head}` is a child of `{pr.base}`.')
+
+        if pr.base != new_base:
+            set_base_of_pull_request(org, repo, pr.number, base=new_base)
+            print(fmt(f'The base branch of PR #{pr.number} is switched to `{new_base}`'))
```
```suggestion
            print(fmt(f'The base branch of PR #{pr.number} has been switched to `{new_base}`'))
```

git_machete/cmd.py
```diff
@@ -1889,6 +2069,21 @@ def check_required_param(in_args: List[str], allowed_values_string: str) -> str:
...
+                machete_client.sync_annotations_to_github_prs()
+            elif param == "create-pr":
+                cb = git.current_branch()
+                machete_client.create_github_pr(cb, draft=cli_opts.opt_draft)
+            elif param == "retarget-pr":
+                cb = git.current_branch()
+                machete_client.expect_in_managed_branches(cb)
+                machete_client.retarget_github_pr(cb)
+            else:
+                raise MacheteException(f"`gh` requires a subcommand: one of `{github_allowed_subcommands}`")
```
PawelLipski: `s/gh/github`

git_machete/cmd.py
```diff
@@ -1570,15 +1517,246 @@ def flush_caches(self) -> None:
...
+                                  "Proceeding with pull request creation...")
+            elif s == NO_REMOTES:
+                raise MacheteException("Could not create pull request - there is no remote repository!")
+            else:
+                ans = 'y'  # only IN SYNC status is left
+
+            if ans in ('y', 'yes', 'yq'):
+                return
+            elif ans in ('N', 'n', 'q', 'quit'):
+                raise MacheteException('Pull request creation interrupted.')
```
PawelLipski, line 1747: Maybe just `raise StopTraversal`, as in other places?

git_machete/exceptions.py
PawelLipski: Hmmm but wait... isn't 422 a perfectly predicted case (when `assigning someone from outside organization as a reviewer or creating a pull request for a branch that already has a PR`)? should a user create an issue in such case?

git_machete/cmd.py
```diff
@@ -1570,15 +1517,246 @@ def flush_caches(self) -> None:
...
+                self.__print_new_line(False)
+                ans = self.ask_if("Proceed with pull request creation?" + pretty_choices('y', 'N', 'q', 'yq'),
+                                  "Proceeding with pull request creation...")
+            elif s == DIVERGED_FROM_AND_OLDER_THAN_REMOTE:
+                warn(f"Branch {cb} is in <b>DIVERGED_FROM_AND_OLDER_THAN_REMOTE</b> state.\nConsider using 'git reset --keep'.\n")
+                self.__print_new_line(False)
+                ans = self.ask_if("Proceed with pull request creation?" + pretty_choices('y', 'N', 'q', 'yq'),
+                                  "Proceeding with pull request creation...")
+            elif s == NO_REMOTES:
+                raise MacheteException("Could not create pull request - there is no remote repository!")
```
PawelLipski: Nit:
```suggestion
                raise MacheteException("Could not create pull request - there are no remote repositories!")
```

git_machete/cmd.py
```diff
@@ -1570,15 +1517,246 @@ def flush_caches(self) -> None:
...
+        ans = self.ask_if(f"Branch {bold(b)} is behind its remote counterpart {bold(rb)}.\n"
+                          f"Pull {bold(b)} (fast-forward only) from {bold(remote)}?" + pretty_choices('y', 'N', 'q', 'yq'),
+                          f"Branch {bold(b)} is behind its remote counterpart {bold(rb)}.\nPulling {bold(b)} (fast-forward only) from {bold(remote)}...")
+        if ans in ('y', 'yes', 'yq'):
+            self.__git.pull_ff_only(remote, rb)
+            if ans == 'yq':
+                raise StopTraversal
+            self.flush_caches()
+            print("")
+        elif ans in ('n', 'N', 'q', 'quit'):
```
PawelLipski: `ask_if` returns a lowercase string (see the impl), pls remove upper case letters from the similar checks everywhere

git_machete/cmd.py
```diff
@@ -1570,15 +1517,246 @@ def flush_caches(self) -> None:
...
+
+    def __sync_before_creating_pr(self) -> None:
+
+        self.expect_at_least_one_managed_branch()
+        self.__empty_line_status = True
+
+        cb = self.__git.current_branch()
+        try:
+            self.expect_in_managed_branches(cb)
+        except MacheteException:
```
PawelLipski: Maybe just check if `cb in managed_branches` instead of calling `expect_in_managed_branches`?

git_machete/exceptions.py
rifel123: Yes, then should be ticket opened. Error 422 tells that operation is not allowed by GitHub for some reason. `assigning someone from outside organization as a reviewer or creating a pull request for a branch that already has a PR` are just known possible reasons to get 422. I expect to be more of them, especially in functions which use `PATCH` method. AFAICS all those errors could be avoided by simple checks, like it is right now in case of `creating a pull request for a branch that already has a PR`. `assigning someone from outside organization` could also be avoided by checking in GitHub API who is in organization (I'll open an issue for this as soon as this PR is merged). In conclusion every 422 error is a great base to open a ticket.

rifel123: In case we cannot do anything with 422, it should be caught by higher lever function.

git_machete/cmd.py
rifel123: Hmm🤔, here the exception is not connected to `traverse` plus there has to be a message to user that there is nothing left for git_machete to do.

git_machete/exceptions.py
PawelLipski: Okay! thanks for the above clarification, I've got lost a bit.

PawelLipski (commented):
Haha I can't approve coz it's technically my own PR 😅 good job @rifel123, pls approve & merge, I'll dogfood it myself on another project's PRs (and encourage you and @BartoszSiwinski to do the same on this project) before 3.4.0 is released 🎉

rifel123 (approved):

PR #135, opened 2021-06-16
Fix newly created branches sometimes incorrectly recognized as merged to parent

PawelLipski:
Ad-hoc fix for the bug observed by @micpiotrowski's.
Let's see if that's a good way to address the problem.

Counterpart of https://github.com/VirtusLab/git-machete-intellij-plugin/pull/742


git_machete/cmd.py
```diff
@@ -1122,7 +1122,7 @@ def merge_base(cli_ctxt: CommandLineContext, sha1: str, sha2: str) -> str:
 
 # Note: the 'git rev-parse --verify' validation is not performed in case for either of earlier/later
 # if the corresponding prefix is empty AND the revision is a 40 hex digit hash.
-def is_ancestor(
+def is_ancestor_or_equal(
```
PawelLipski, line 1125: Just a rename to correctly reflect what this method was doing all along

git_machete/cmd.py
```diff
@@ -1690,18 +1690,14 @@ def lb_is_not_b(lb: str, lb_or_rb: str) -> bool:
 # Complex routines/commands
 
 def is_merged_to(cli_ctxt: CommandLineContext, b: str, target: str) -> bool:
-    if commit_sha_by_revision(cli_ctxt, target) == commit_sha_by_revision(cli_ctxt, b):
-        # If branch is equal to the target, we need to distinguish between the
+    if is_ancestor_or_equal(cli_ctxt, b, target):
```
PawelLipski, line 1693: The actual change... it's suspiciously simple, I'm afraid this will backfire somehow that I can't see yet :/

PawelLipski: YOLO

PR #151, opened 2021-07-22
Introduce MacheteClient class

rifel123:


git_machete/cmd.py
```diff
@@ -339,409 +339,1171 @@ def popen_git(cli_ctxt: CommandLineContext, git_cmd: str, *args: str, **kwargs:
...
+    def expect_in_managed_branches(self, b: str) -> None:
+        if b not in self.managed_branches:
+            raise MacheteException(
+                f"Branch `{b}` not found in the tree of branch dependencies.\nUse `git machete add {b}` or `git machete edit`")
+
+    def expect_at_least_one_managed_branch(self) -> None:
+        if not self.roots:
+            self.raise_no_branches_error()
+
+    def raise_no_branches_error(self) -> None:
```
PawelLipski, line 373: Consider making methods private where applicable 🔒

git_machete/cmd.py
```diff
@@ -339,409 +339,1171 @@ def popen_git(cli_ctxt: CommandLineContext, git_cmd: str, *args: str, **kwargs:
...
+            f.write("\n".join(self.render_tree()) + "\n")
+
+    # Allowed parameter values for show/go command
+    @staticmethod
+    def allowed_directions(allow_current: bool) -> str:
+        current = "c[urrent]|" if allow_current else ""
+        return current + "d[own]|f[irst]|l[ast]|n[ext]|p[rev]|r[oot]|u[p]"
+
+    # Parse and evaluate direction against current branch for show/go commands
+    def parse_direction(self, cli_ctxt: CommandLineContext, param: str, b: str, allow_current: bool, down_pick_mode: bool) -> str:
```
PawelLipski: Hmmm I'd say a method like this one (also `allowed_directions`) probably doesn't belong well to `MacheteContext`... this is rather part of the implementation of command-line parsing.

Imagine for instance that we had a GUI client for git-machete written in Python (like, as if https://github.com/VirtusLab/git-machete-intellij-plugin was written in Python instead of Java/Kotlin). `MacheteContext` could be perfectly shared b/w CLI and GUI, but a method like `parse_direction` would be probably be useless for GUI.

PawelLipski: Generally, assume that `MacheteContext` contains things that can be used **by any client**, whether it is CLI, GUI, or even (not very realistically, but still) a mobile/web one.

PawelLipski: There are now some methods in gray area... like for instance the current implementation of `traverse`... can we consider it CLI-specific, or rather more generic? things like that will require further deliberation 🤔

PawelLipski: So what I'd recommend is to move the methods that are **clearly CLI-specific**, like `parse_direction`, out of `MacheteContext`. In the following PRs, we'll think how to redesign this behemoth of a code that is cmd.py so that it's elegantly split into layers (classes/maybe packages)

git_machete/cmd.py
```diff
@@ -3627,7 +3603,7 @@ def check_required_param(in_args: List[str], allowed_values_string: str) -> str:
         else:
             return in_args[0]
 
-    global definition_file_path, up_branch
+    global definition_file_path
```
PawelLipski: Hmmm can we un-globalize this one?

git_machete/cmd.py
```diff
@@ -3482,6 +3457,7 @@ def main() -> None:
 def launch(orig_args: List[str]) -> None:
 
     cli_ctxt = CommandLineContext()
```
PawelLipski, line 3455: Hmmm `CommandLineContext` is created here... but isn't it also created inside of `MacheteContext` as a static field? 🤔 I don't think we should have 2 distinct instances of `CommandLineContext` in the application

git_machete/cmd.py
rifel123: Since we need `definition_file_path` at module level  and manipulate it from functions, the only way to get rid of global statement is to store `definition_file_path` in mutable variable.

git_machete/cmd.py
rifel123: I agree, at first step i wanted to stick to proposal in PR #124. In my opinion `CommandLineContext` object should be passed to `MacheteContext` constructor or created there if it's not passed.

git_machete/cmd.py
```diff
@@ -339,192 +339,1091 @@ def popen_git(cli_ctxt: CommandLineContext, git_cmd: str, *args: str, **kwargs:
...
+
+class MacheteContext:
 
+    cli_ctxt: CommandLineContext = CommandLineContext()
+    DISCOVER_DEFAULT_FRESH_BRANCH_COUNT = 10
 
-def expect_in_managed_branches(b: str) -> None:
-    if b not in managed_branches:
+    def __init__(self) -> None:
+        self.PICK_FIRST_ROOT: int = 0
```
PawelLipski: Nit: constants can be class fields rather than instance fields

git_machete/cmd.py
```diff
@@ -535,60 +1434,56 @@ def down(b: str, pick_mode: bool) -> str:
...
+    return machete_context.managed_branches[idx]
+
+
+def root_branch(machete_context: MacheteContext, b: str, if_unmanaged: int) -> str:
+    if b not in machete_context.managed_branches:
+        if machete_context.roots:
+            if if_unmanaged == machete_context.PICK_FIRST_ROOT:
+                warn(f"{b} is not a managed branch, assuming {machete_context.roots[0]} (the first root) instead as root")
+                return machete_context.roots[0]
+            else:  # if_unmanaged == self.PICK_LAST_ROOT
```
PawelLipski: Nit: `self` not applicable here

git_machete/cmd.py
```diff
@@ -339,359 +339,1027 @@ def popen_git(cli_ctxt: CommandLineContext, git_cmd: str, *args: str, **kwargs:
...
 # Manipulation on definition file/tree of branches
 
-managed_branches: List[str] = []
-down_branches: Dict[str, List[str]] = {}  # TODO (#110): default dict with []
-up_branch: Dict[str, str] = {}  # TODO (#110): default dict with None
-indent: Optional[str] = None
-roots: List[str] = []
-annotations: Dict[str, str] = {}
-
+branch_defs_by_sha_in_reflog: Optional[Dict[str, Optional[List[Tuple[str, str]]]]] = None
```
PawelLipski: Hmm can this one go into `MacheteClient` as well? 🤔

git_machete/cmd.py
```diff
@@ -339,359 +339,1027 @@ def popen_git(cli_ctxt: CommandLineContext, git_cmd: str, *args: str, **kwargs:
...
 
-def expect_in_managed_branches(b: str) -> None:
-    if b not in managed_branches:
-        raise MacheteException(
-            f"Branch `{b}` not found in the tree of branch dependencies.\nUse `git machete add {b}` or `git machete edit`")
+BRANCH_DEF = Tuple[str, str]
+empty_line_status: Optional[bool] = None
```
PawelLipski: Ditto

git_machete/cmd.py
```diff
@@ -339,359 +339,1027 @@ def popen_git(cli_ctxt: CommandLineContext, git_cmd: str, *args: str, **kwargs:
...
-
-        if pfx:
-            depth: int = len(pfx) // len(indent)
-            if pfx != indent * depth:
-                mapping: Dict[str, str] = {" ": "<SPACE>", "\t": "<TAB>"}
-                pfx_expanded: str = "".join(mapping[c] for c in pfx)
-                indent_expanded: str = "".join(mapping[c] for c in indent)
-                raise MacheteException(
-                    f"{definition_file_path}, line {idx + 1}: invalid indent `{pfx_expanded}`, expected a multiply of `{indent_expanded}`. {hint}")
+    def __init__(self, cli_ctxt: CommandLineContext = None) -> None:
```
PawelLipski: TBH I'd rather **not** leave the possibility to rely on the defaults... let the users of  `MacheteClient` pass the `CommandLineContext` explicitly

git_machete/cmd.py
```diff
@@ -339,359 +339,1027 @@ def popen_git(cli_ctxt: CommandLineContext, git_cmd: str, *args: str, **kwargs:
...
+BRANCH_DEF = Tuple[str, str]
+empty_line_status: Optional[bool] = None
 
 
-def expect_at_least_one_managed_branch() -> None:
-    if not roots:
-        raise_no_branches_error()
+class MacheteClient:
 
+    cli_ctxt: CommandLineContext = CommandLineContext()
```
PawelLipski: Hmmm this is treated as a **class attribute** here, but in the constructor it seems to be an **instance attribute** 🤔

git_machete/cmd.py
```diff
@@ -339,359 +339,1027 @@ def popen_git(cli_ctxt: CommandLineContext, git_cmd: str, *args: str, **kwargs:
...
+    DISCOVER_DEFAULT_FRESH_BRANCH_COUNT = 10
+    PICK_FIRST_ROOT: int = 0
+    PICK_LAST_ROOT: int = -1
+    definition_file_path: str = ""
```
PawelLipski: Same here, shouldn't it be an instance attribute rather than a class one?

git_machete/cmd.py
PawelLipski: I'm not even sure how Python treats an override like that... whether it's possible to have two separate class&instance attributes with the same name :/

PawelLipski: Anyway, this should probably be an instance attribute

git_machete/cmd.py
PawelLipski: Even if that's going to be just a freshly-created `CommandLineContext()`

git_machete/cmd.py
```diff
@@ -339,359 +339,1027 @@ def popen_git(cli_ctxt: CommandLineContext, git_cmd: str, *args: str, **kwargs:
...
+            for d in self.down_branches.get(b, []):
+                res += render_dfs(d, depth + 1)
+            return res
+
+        total: List[str] = []
+        for r in self.roots:
+            total += render_dfs(r, depth=0)
+        return total
+
+    @staticmethod
```
PawelLipski: If `definition_file_path` is converted to an instance attribute (which I think it should), then this should no longer be a static method

git_machete/cmd.py
```diff
@@ -339,359 +339,1027 @@ def popen_git(cli_ctxt: CommandLineContext, git_cmd: str, *args: str, **kwargs:
...
+                ans = ask_if(self.cli_ctxt, msg, opt_yes_msg)
+                if ans in ('y', 'yes'):
+                    run_git(self.cli_ctxt, "branch", "-D", b)
+                elif ans in ('q', 'quit'):
+                    return
+        else:
+            print("No branches to delete")
+
+    @staticmethod
+    def edit(cli_ctxt: CommandLineContext) -> int:
```
PawelLipski: Probably no method of `MacheteClient` whatsoever should have a `CommandLineContext` as a parameter, right? 🤔 all should rely on `self.cli_ctxt` instead

git_machete/cmd.py
```diff
@@ -339,359 +339,1027 @@ def popen_git(cli_ctxt: CommandLineContext, git_cmd: str, *args: str, **kwargs:
...
-    global annotations, managed_branches
+def slidable_after(machete_context: MacheteClient, b: str) -> List[str]:
+    if b in machete_context.up_branch:
+        dbs = machete_context.down_branches.get(b)
+        if dbs and len(dbs) == 1:
+            return dbs
+    return []
+
+
+def sync_annotations_to_github_prs(machete_context: MacheteClient, cli_ctxt: CommandLineContext) -> None:
```
PawelLipski: Pls use `machete_client` name for the params of `MacheteClient` type

PawelLipski: Also! passing a `CommandLineContext` is probably redundant here? coz this can/should be extracted from `machete_context` instead

git_machete/cmd.py
```diff
@@ -3476,12 +3467,10 @@ def main() -> None:
     launch(sys.argv[1:])
 
 
-definition_file_path: str = ""
-
-
 def launch(orig_args: List[str]) -> None:
 
     cli_ctxt = CommandLineContext()
+    machete_ctxt = MacheteClient(cli_ctxt)
```
```suggestion
    machete_client = MacheteClient(cli_ctxt)
```

git_machete/cmd.py
rifel123: cli_ctxt as class attribute is not used anywhere, hence it should be removed.

git_machete/cmd.py
rifel123: moved to instance method

git_machete/cmd.py
rifel123: I agree with empty_line_status but BRANCH_DEF and branch_defs_by_sha_in_reflog are not quite used in MacheteClient class. Hence it should be kept outside.

PawelLipski: Okay, let's think on next PR when some `utils` (?) package is extracted

git_machete/cmd.py
```diff
@@ -399,7 +395,7 @@ def read_definition_file(self, verify_branches: bool = True) -> None:
                 self.annotations[b] = b_a[1]
             if b in self.managed_branches:
                 raise MacheteException(
-                    f"{MacheteClient.definition_file_path}, line {idx + 1}: branch `{b}` re-appears in the tree definition. {hint}")
+                    f"{self.definition_file_path}, line {idx + 1}: branch `{b}` re-appears in the tree definition. {hint}")
```
PawelLipski: While you're here... pls expand `idx` into `index`, `pfx` into `prefix` wherever you find one

git_machete/cmd.py
```diff
@@ -1347,22 +1340,22 @@ def log(self, branch: str) -> None:
         run_git(self.cli_ctxt, "log", "^" + self.fork_point(branch, use_overrides=True), f"refs/heads/{branch}")
 
 
-def slidable(machete_context: MacheteClient) -> List[str]:
-    return [b for b in machete_context.managed_branches if b in machete_context.up_branch]
+def slidable(machete_client: MacheteClient) -> List[str]:
```
PawelLipski: Hmmm actually... now that `MacheteClient` has a wider responsibility (and no longer corresponds just to the CLI/GUI-independent parts of the logic)... I think methods like `slidable`, `down` and the similar ones can also go into `MacheteClient`.

PawelLipski: The things I would rather still NOT see in `MacheteClient` are:

1. Utilities that can be used in other pieces of software that know of `git` but don't know of `git-machete` (e.g. `get_git_version`) or the ones that don't even know of `git` (e.g. `directory_exists`) ➡️ they should go into separate Python subpackages
2. Any method that related to parsing command-line params/input &mdash; `MacheteClient` should be suitable for **programmatic access**, i.e. directly from Python code rather than from command-line

LMK if you've got any questions

git_machete/cmd.py
```diff
@@ -847,7 +845,7 @@ def print_new_line(new_status: bool) -> None:
         for b in itertools.dropwhile(lambda x: x != cb, self.managed_branches):
             u = self.up_branch.get(b)
 
-            needs_slide_out: bool = is_merged_to_upstream(self, self.cli_ctxt, b)
+            needs_slide_out: bool = is_merged_to_upstream(self, b)
```
PawelLipski: In the spirit of my previous longish comment, I think `is_merged_to_upstream` can also be a part of `MacheteContext`

PawelLipski: This is a very machete-specific thing, and can be used programmatically as-is (no parsing of command-line input involved)

git_machete/cmd.py
```diff
@@ -1396,16 +1394,16 @@ def sync_annotations_to_github_prs(machete_client: MacheteClient) -> None:
...
+def is_merged_to_upstream(machete_client: MacheteClient, b: str) -> bool:
     if b not in machete_client.up_branch:
         return False
-    return is_merged_to(cli_ctxt, b, machete_client.up_branch[b])
+    return is_merged_to(machete_client.cli_ctxt, b, machete_client.up_branch[b])
 
 
 # Parse and evaluate direction against current branch for show/go commands
-def parse_direction(machete_client: MacheteClient, cli_ctxt: CommandLineContext, param: str, b: str, allow_current: bool, down_pick_mode: bool) -> str:
+def parse_direction(machete_client: MacheteClient, param: str, b: str, allow_current: bool, down_pick_mode: bool) -> str:
```
PawelLipski: This one, in turn, probably should NOT be a part of `MacheteClient` &mdash; too much command-line input parsing involved, that won't be handy/useful for programmatic access from Python code

PawelLipski:
Btw pls give this PR a more adequate name :) best to use **imperative form**, see https://chris.beams.io/posts/git-commit/#imperative

git_machete/cmd.py
rifel123: What about methods, which are connected to fork point, egz. get_fork_point_override_data?
Also sync_annotations_to_github_prs looks like Machete specific.
@PawelLipski What do you think?

git_machete/cmd.py
```diff
@@ -1065,16 +1065,17 @@ def print_new_line(new_status: bool) -> None:
 
     def status(self, warn_on_yellow_edges: bool) -> None:
         dfs_res = []
+        prefix: str
 
-        def prefix_dfs(u_: str, prefix: List[Optional[str]]) -> None:
-            dfs_res.append((u_, prefix))
+        def prefix_dfs(u_: str, prefix_df: List[Optional[str]]) -> None:
```
PawelLipski: Maybe rename `prefix_df` to `accumulated_path`, it will be more telling (since this param basically corresponds to a path in a tree &mdash; from a tree root to the current vertex)

git_machete/cmd.py
PawelLipski: Hmmm TBH as I look on that now... this logic is also super-Machete-specific...

PawelLipski: I'd pull that (and the associated functions) into the Client as well :/

git_machete/cmd.py
PawelLipski: (as agreed, on a follow-up PR)

git_machete/cmd.py
```diff
@@ -339,362 +339,1207 @@ def popen_git(cli_ctxt: CommandLineContext, git_cmd: str, *args: str, **kwargs:
...
+                    elif ans in ('q', 'quit'):
+                        return
 
+                elif s == DIVERGED_FROM_AND_NEWER_THAN_REMOTE:
+                    print_new_line(False)
+                    rb = strict_counterpart_for_fetching_of_branch(self.cli_ctxt, b)
+                    ans = ask_if(
+                        self.cli_ctxt,
+                        f"Branch {bold(b)} diverged from (and has newer commits than) its remote counterpart {bold(rb)}.\n"
+                        f"Push {bold(b)} with force-with-lease to {bold(remote)}?" + pretty_choices('y', 'N', 'q',
```
PawelLipski: Nit: merge lines

git_machete/cmd.py
```diff
@@ -339,362 +339,1207 @@ def popen_git(cli_ctxt: CommandLineContext, git_cmd: str, *args: str, **kwargs:
...
+        for sha, short_sha, subject in commits:
+            print(f"\t{short_sha} {subject}")
+
+        latest_sha, latest_short_sha, latest_subject = commits[-1]
+        print()
+        print("To restore the original pre-squash commit, run:")
+        print()
+        print(fmt(f"\t`git reset {latest_sha}`"))
+
+    def filtered_reflog(self, b: str, prefix: str) -> List[str]:  # add
```
PawelLipski: `# add` to be removed probably

git_machete/cmd.py
```diff
@@ -339,362 +339,1207 @@ def popen_git(cli_ctxt: CommandLineContext, git_cmd: str, *args: str, **kwargs:
...
+              f"filtered_reflog({b}, {prefix})",
+              "computed filtered reflog (= reflog without branch creation "
+              "and branch reset events irrelevant for fork point/upstream inference): %s\n" % (
+                          ", ".join(result) or "<empty>"))
+        return result
 
 
-def sync_annotations_to_github_prs(cli_ctxt: CommandLineContext) -> None:
-    global annotations, managed_branches
+def sync_annotations_to_github_prs(machete_client: MacheteClient) -> None:  # ask if member of MacheteClient
```
PawelLipski: :yes:

PawelLipski: I changed my mind on this one, as the general idea of `MacheteContext` changed more into `MacheteClient` 😅

git_machete/cmd.py
```diff
@@ -766,14 +1632,14 @@ def find_executable(cli_ctxt: CommandLineContext, executable: str) -> Optional[s
     return None
 
 
-def get_default_editor(cli_ctxt: CommandLineContext) -> str:
+def get_default_editor(machete_client: MacheteClient) -> str:
```
PawelLipski: Hmmm I think passing just `cli_ctxt` is sufficient here? 🤔

PawelLipski: As in, this method doesn't seem to rely on anything specific to `MacheteClient`

git_machete/cmd.py
```diff
@@ -1661,12 +2410,7 @@ def get_latest_checkout_timestamps(cli_ctxt: CommandLineContext) -> Dict[str, in
     return result
 
 
-branch_defs_by_sha_in_reflog: Optional[Dict[str, Optional[List[Tuple[str, str]]]]] = None
-
-BRANCH_DEF = Tuple[str, str]
-
-
-def match_log_to_filtered_reflogs(cli_ctxt: CommandLineContext, b: str) -> Generator[Tuple[str, List[BRANCH_DEF]], None, None]:
+def match_log_to_filtered_reflogs(machete_context: MacheteClient,cli_ctxt: CommandLineContext, b: str) -> Generator[Tuple[str, List[BRANCH_DEF]], None, None]:
```
PawelLipski: As mentioned above, this is too Machete-specific, should go to `MacheteClient` (probably as a private method)  :sad-pepe:

git_machete/cmd.py
```diff
@@ -1749,14 +2493,8 @@ def is_merged_to(cli_ctxt: CommandLineContext, b: str, target: str) -> bool:
...
 
-def is_merged_to_upstream(cli_ctxt: CommandLineContext, b: str) -> bool:
-    if b not in up_branch:
-        return False
-    return is_merged_to(cli_ctxt, b, up_branch[b])
-
-
-def infer_upstream(cli_ctxt: CommandLineContext, b: str, condition: Callable[[str], bool] = lambda u: True, reject_reason_message: str = "") -> Optional[str]:
-    for sha, containing_branch_defs in match_log_to_filtered_reflogs(cli_ctxt, b):
+def infer_upstream(machete_context: MacheteClient, cli_ctxt: CommandLineContext, b: str, condition: Callable[[str], bool] = lambda u: True, reject_reason_message: str = "") -> Optional[str]:
```
PawelLipski: Also definitely a part of `MacheteClient`.

PawelLipski: Also! pls hunt down the remaining occurrences of `machete_context` (still 7 left)

PawelLipski (changes requested):
Probably the (pen)ultimate round of review :)

git_machete/cmd.py
```diff
@@ -532,7 +532,7 @@ def add(self, b: str) -> None:
             print(fmt(f"Added branch `{b}` as a new root"))
         else:
             if not onto:
-                u = infer_upstream(self.cli_ctxt, b, condition=lambda x: x in self.managed_branches, reject_reason_message="this candidate is not a managed branch")
+                u = infer_upstream(self, self.cli_ctxt, b, condition=lambda x: x in self.managed_branches, reject_reason_message="this candidate is not a managed branch")
```
PawelLipski: `infer_upstream` to be pulled into `MacheteContext` (probably as private method?)

git_machete/cmd.py
rifel123: MacheteClient object is used here to pass definition_file_path

git_machete/cmd.py
```diff
@@ -1111,14 +1111,14 @@ def fp_sha(b: str) -> Optional[str]:
...
-            for p in line_prefix[:-1]:
+            for p in pfx[:-1]:
                 if not p:
                     out.write("  ")
                 else:
                     out.write(colored(f"{vertical_bar()} ", edge_color[p]))
             out.write(colored(suffix, edge_color[b_]))
 
-        for b, line_prefix in dfs_res:
+        for b, pfx in dfs_res:
```
PawelLipski: `pfx` -> `accumulated_path` for consistency

git_machete/cmd.py
PawelLipski: Hmmm I'd refactor this piece so that `get_default_editor` doesn't need to refer to `definition_file_path`: make `get_default_editor` return `None` if nothing is found, and let the call site (`edit()` method) raise an error instead

PawelLipski: (since the only place where `definition_file_path` is needed is in the error message)

git_machete/cmd.py
```diff
@@ -1846,7 +1848,7 @@ def get_default_editor(machete_client: MacheteClient) -> str:
 
     # This case is extremely unlikely on a modern Unix-like system.
     raise MacheteException(
-        f"Cannot determine editor. Set `{git_machete_editor_var}` environment variable or edit {machete_client.definition_file_path} directly.")
+        f"Cannot determine editor. Set `{git_machete_editor_var}` environment variable.")
```
PawelLipski: Hmm I'd rather `return None` in this case (so the type of the method should change to `Optional[str]`)

PawelLipski: And raise an exception with the exact same message as before this PR, but from `edit` and not from `get_default_editor`

git_machete/cmd.py
```diff
@@ -1254,7 +1254,9 @@ def delete_unmanaged(self) -> None:
             print("No branches to delete")
 
     def edit(self) -> int:
-        return run_cmd(self.cli_ctxt, get_default_editor(self), self.definition_file_path)
+        if self.definition_file_path is None or self.definition_file_path == "":
```
PawelLipski: Hmm is there any valid execution path where `definition_file_path` ends up `None` or `""`, other than the invocation of `git machete help`? 🤔

PawelLipski: Actually! I think it makes sense to set `self.definition_file_path` in c'tor of this class rather than in `launch` function

git_machete/cmd.py
rifel123: an exception will be raised when get_default_editor returns None.

git_machete/cmd.py
```diff
@@ -339,410 +339,1452 @@ def popen_git(cli_ctxt: CommandLineContext, git_cmd: str, *args: str, **kwargs:
...
-                if ask_if(
-                        cli_ctxt,
-                        prompt_if_inferred_msg % (b, u),
-                        prompt_if_inferred_yes_opt_msg % (b, u)
-                ) in ('y', 'yes'):
-                    return u
+            print("No branches to delete")
+
+    def edit(self) -> int:
+        if get_default_editor(self.cli_ctxt) is None:
```
PawelLipski: Limit to just one call of `get_default_editor`! no point in calling it twice

git_machete/cmd.py
PawelLipski: (bump)

rifel123: added as class variable due to usage in flush_caches function

git_machete/cmd.py
```diff
@@ -805,12 +1847,7 @@ def get_default_editor(cli_ctxt: CommandLineContext) -> str:
                 return editor
 
     # This case is extremely unlikely on a modern Unix-like system.
-    raise MacheteException(
-        f"Cannot determine editor. Set `{git_machete_editor_var}` environment variable or edit {definition_file_path} directly.")
-
-
-def edit(cli_ctxt: CommandLineContext) -> int:
-    return run_cmd(cli_ctxt, get_default_editor(cli_ctxt), definition_file_path)
+    raise None
```
PawelLipski: Not `raise`, `return` 🙀

PawelLipski (approved):
Good job, approving 🎉

PR #167, opened 2021-08-06
Tests for go, show, update and advance 

### LINKED ISSUE #120: Provide missing functional tests for `advance`, `go`, `show`, `update`
Currently, only `discover`, `status`, `traverse` and `squash` are covered by function tests.

BartoszSiwinski:
Connected issue: #120 
Test coverage for:
- go up/down/first/last/next/prev/root
- show up/down/first/last/next/prev/root

### DIFF
--- a/git_machete/tests/functional/test_machete.py
+++ b/git_machete/tests/functional/test_machete.py
@@ -17,59 +17,64 @@ cli_opts: CommandLineOptions = CommandLineOptions()
 git: GitContext = GitContext(cli_opts)
 
 
-class SandboxSetup:
-    def __init__(self) -> None:
-        self.file_dir = os.path.dirname(os.path.abspath(__file__))
-        self.remote_path = os.popen("mktemp -d").read().strip()
-        self.sandbox_path = os.popen("mktemp -d").read().strip()
+def get_head_commit_hash() -> str:
+    """Returns hash of a commit of the current branch head."""
+    return os.popen("git rev-parse HEAD").read().strip()
 
-    def execute(self, command: str) -> "SandboxSetup":
+
+class GitRepositorySandbox:
+    def __init__(self) -> None:
+        self.remote_path = os.popen("mktemp -d").read().strip()
+        self.local_path = os.popen("mktemp -d").read().strip()
+
+    def execute(self, command: str) -> "GitRepositorySandbox":
         result = os.system(command)
         assert result == 0, f"{command} returned {result}"
         return self
 
-    def new_repo(self, *args: str) -> "SandboxSetup":
+    def new_repo(self, *args: str) -> "GitRepositorySandbox":
         os.chdir(args[0])
-        if len(args) > 1:
-            opt = args[1]
-            self.execute(f"git init {opt}")
-        else:
-            self.execute("git init")
+        opts = args[1:]
+        self.execute(f"git init -q {' '.join(opts)}")
         return self
 
-    def new_branch(self, branch_name: str) -> "SandboxSetup":
+    def new_branch(self, branch_name: str) -> "GitRepositorySandbox":
         self.execute(f"git checkout -q -b {branch_name}")
         return self
 
-    def check_out(self, branch: str) -> "SandboxSetup":
+    def new_root_branch(self, branch_name: str) -> "GitRepositorySandbox":
+        self.execute(f"git checkout --orphan {branch_name}")
+        return self
+
+    def check_out(self, branch: str) -> "GitRepositorySandbox":
         self.execute(f"git checkout -q {branch}")
         return self
 
-    def commit(self, message: str) -> "SandboxSetup":
+    def commit(self, message: str = "Some commit message.") -> "GitRepositorySandbox":
         f = "%s.txt" % "".join(random.choice(string.ascii_letters) for _ in range(20))
         self.execute(f"touch {f}")
         self.execute(f"git add {f}")
         self.execute(f'git commit -q -m "{message}"')
         return self
 
-    def commit_amend(self, message: str) -> "SandboxSetup":
+    def commit_amend(self, message: str) -> "GitRepositorySandbox":
         self.execute(f'git commit -q --amend -m "{message}"')
         return self
 
-    def push(self) -> "SandboxSetup":
+    def push(self) -> "GitRepositorySandbox":
         branch = os.popen("git symbolic-ref --short HEAD").read()
         self.execute(f"git push -q -u origin {branch}")
         return self
 
-    def sleep(self, seconds: int) -> "SandboxSetup":
+    def sleep(self, seconds: int) -> "GitRepositorySandbox":
         time.sleep(seconds)
         return self
 
-    def reset_to(self, revision: str) -> "SandboxSetup":
+    def reset_to(self, revision: str) -> "GitRepositorySandbox":
         self.execute(f'git reset --keep "{revision}"')
         return self
 
-    def delete_branch(self, branch: str) -> "SandboxSetup":
+    def delete_branch(self, branch: str) -> "GitRepositorySandbox":
         self.execute(f'git branch -d "{branch}"')
         return self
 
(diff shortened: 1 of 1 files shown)


git_machete/tests/functional/test_machete.py
```diff
@@ -37,39 +41,43 @@ def new_repo(self, *args: str) -> "SandboxSetup":
...
+    def new_root_branch(self, branch_name: str) -> "GitRepositorySandbox":
+        self.execute(f"git checkout --orphan {branch_name}")
+        return self
+
+    def check_out(self, branch: str) -> "GitRepositorySandbox":
         self.execute(f"git checkout -q {branch}")
         return self
 
-    def commit(self, message: str) -> "SandboxSetup":
+    def commit(self, message: str = "SomeCommitMsg") -> "GitRepositorySandbox":
```
PawelLipski: Nit: to make it look more like an actual commit message:
```suggestion
    def commit(self, message: str = "Some commit message") -> "GitRepositorySandbox":
```

git_machete/tests/functional/test_machete.py
```diff
@@ -87,29 +95,29 @@ def launch_command(*args: str) -> str:
...
+        self.repo_sandbox = GitRepositorySandbox()
 
         (
-            self.setup
+            self.repo_sandbox
             # Create the remote and sandbox repos, chdir into sandbox repo
-            .new_repo(self.setup.remote_path, "--bare")
-            .new_repo(self.setup.sandbox_path)
-            .execute(f"git remote add origin {self.setup.remote_path}")
+            .new_repo(self.repo_sandbox.remote_path, "--bare", "-q")
```
PawelLipski: Hmmm looking at the impl of `new_repo`... isn't `-q` ignored here? 🤔

PawelLipski: But generally good idea, I'd add some `--quiet` option wherever possible to the setup commands (as opposed to the actual tested `git machete` commands) wherever possible to cut on spam in output

git_machete/tests/functional/test_machete.py
```diff
@@ -672,3 +680,682 @@ def test_help(self) -> None:
                         self.fail(f'Unexpected exception raised: {e}')
                     else:
                         self.fail('SystemExit expected but not raised')
+
+    def test_go_up(self) -> None:
+        """Verify behaviour of a 'git machete go up' command.
+
+        Verify that 'git machete go up' performs 'git checkout' to the
+        parent/upstream branch of a current branch.
```
PawelLipski: Nit:
```suggestion
        parent/upstream branch of the current branch.
```
since there's only (at most) one, well-defined current branch

PawelLipski: Here and in the tests below

git_machete/tests/functional/test_machete.py
```diff
@@ -672,3 +680,682 @@ def test_help(self) -> None:
...
+            'level-1-branch',
+            self.launch_command("show", "current").strip(),
+            msg="Verify that 'git machete go down' performs 'git checkout' to "
+                "the child/downstream branch of a current branch."
+        )
+
+    def test_go_first_root_with_downstream(self) -> None:
+        """Verify behaviour of a 'git machete go first' command.
+
+        Verify that 'git machete go first' performs 'git checkout' to
```
PawelLipski, line 733: Hmmm since the message in assertion seems to be identical to the function doc for all tests... maybe putting the message in doc is redundant then?

git_machete/tests/functional/test_machete.py
```diff
@@ -672,3 +680,682 @@ def test_help(self) -> None:
...
+
+        self.assertEqual(
+            parents_new_commit_hash,
+            new_forkpoint_hash,
+            msg="Verify that 'git machete update --no-interactive rebase' perform"
+                "'git rebase' to the parent branch of a current branch."
+        )
+
+    def test_update_with_fork_point_specified(self) -> None:
+        """Verify behaviour of a 'git machete update --no-interactive rebase -f #commit_hash' cmd.
```
PawelLipski: Nit, the informal convention for placeholders in command line is:
```suggestion
        """Verify behaviour of a 'git machete update --no-interactive rebase -f <commit_hash>' cmd.
```

git_machete/tests/functional/test_machete.py
BartoszSiwinski: Accepted

git_machete/tests/functional/test_machete.py
BartoszSiwinski: Accepted

git_machete/tests/functional/test_machete.py
```diff
@@ -672,3 +680,682 @@ def test_help(self) -> None:
...
+            self.launch_command("status"),
+            msg="Verify that branch to which advance was performed is removed "
+                "from the git-machete tree and the structure of the git machete "
+                "tree is updated.")
+
+    def test_advance_with_few_possible_downstream_branches_and_yes_option(self) -> None:
+        """Verify behaviour of a 'git machete advance' command.
+
+        Verify that 'git machete advance -y' raises an error when current branch
+        has many synchronized downstream branches and option '-y' is passed.
```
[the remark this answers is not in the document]
BartoszSiwinski: Accepted

git_machete/tests/functional/test_machete.py
BartoszSiwinski: No entirely. Doc describes whole test method, whereas assertion describe purpose of a particular oracle.

See below case with multiple assertions/oracles.
image
image

If you prefer that I can remove it from doc, but I think it could help someone get a gist of what is tested by just reading docstring and not looking into the code unnecessarly.

PawelLipski: Okay! right, makes sense &mdash; let's leave as is

PawelLipski (approved):

git_machete/tests/functional/test_machete.py
BartoszSiwinski: You're right, it was ignored. When I swapped the order, so the -q comes before --bare it hides hints about "main/master" naming convention.

git_machete/tests/functional/test_machete.py
```diff
@@ -108,7 +108,7 @@ def setUp(self) -> None:
         (
             self.repo_sandbox
             # Create the remote and sandbox repos, chdir into sandbox repo
-            .new_repo(self.repo_sandbox.remote_path, "--bare", "-q")
+            .new_repo(self.repo_sandbox.remote_path, "-q", "--bare")
```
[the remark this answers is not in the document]
PawelLipski: Yeah, it definitely is... and the remote repo is created as non-bare repo, that's why the tests are failing. In general, the repos that are pushed to (i.e. serve as remote repos) should be created as bare (no working directory, only `.git/` folder).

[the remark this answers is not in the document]
PawelLipski: Okay, just `opts` instead of `opt`... also, isn't `-q` aka `--quiet` always passed anyway?

PawelLipski:
Also, as a general remark, pls push review fixes **on separate commits**... the Gerrit's amend-based flow doesn't work well with Github :/

PawelLipski:
And, on the other hand, Github provides good support for e.g. viewing just the changes introduced by a range of commits (rather than the entire PR)

BartoszSiwinski:
Ok, will do

git_machete/tests/functional/test_machete.py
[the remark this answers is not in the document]
BartoszSiwinski: Accepted

git_machete/tests/functional/test_machete.py
```diff
@@ -36,9 +36,9 @@ def new_repo(self, *args: str) -> "GitRepositorySandbox":
         os.chdir(args[0])
         if len(args) > 1:
             opts = args[1:]
-            self.execute(f"git init {' '.join(opts)}")
+            self.execute(f"git init -q {' '.join(opts)}")
```
PawelLipski: Nit: the params passed to `execute` are parsed by the shell (i.e. split on spaces into separate words etc., unlike with e.g. `subprocess.Popen` where you need to pass already parsed arguments as list of strings). Thus, you can probably remove the `if len(args) > 1` and just put `' '.join(args[1:])` unconditionally into the string. If `len(args) == 1`, then it'll simply expand to an empty string.

PawelLipski (approved):
Good job, minor comment left

PR #159, opened 2021-08-03
Split cmd.py into smaller modules

rifel123:
implements issue #34

PR #168, opened 2021-08-09
Add private and protected fields

rifel123:
Note: there is couple of open points done from previous PR #159

### DIFF
--- a/git_machete/cmd.py
+++ b/git_machete/cmd.py
@@ -30,33 +30,6 @@ T = TypeVar('T')
 
 BRANCH_DEF = Tuple[str, str]
 Hash_ShortHash_Message = Tuple[str, str, str]
-
-
-def ask_if(cli_opts: CommandLineOptions, msg: str, opt_yes_msg: Optional[str], override_answer: Optional[str] = None,
-           apply_fmt: bool = True) -> str:
-    if override_answer:
-        return override_answer
-    if cli_opts.opt_yes and opt_yes_msg:
-        print(fmt(opt_yes_msg) if apply_fmt else opt_yes_msg)
-        return 'y'
-    return input(fmt(msg) if apply_fmt else msg).lower()
-
-
-def pick(choices: List[str], name: str, apply_fmt: bool = True) -> str:
-    xs: str = "".join(f"[{index + 1}] {x}\n" for index, x in enumerate(choices))
-    msg: str = xs + f"Specify {name} or hit <return> to skip: "
-    try:
-        ans: str = input(fmt(msg) if apply_fmt else msg)
-        if not ans:
-            sys.exit(0)
-        index: int = int(ans) - 1
-    except ValueError:
-        sys.exit(1)
-    if index not in range(len(choices)):
-        raise MacheteException(f"Invalid index: {index + 1}")
-    return choices[index]
-
-
 initial_current_directory: Optional[str] = utils.current_directory_or_none() or os.getenv('PWD')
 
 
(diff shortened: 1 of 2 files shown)


BartoszSiwinski (commented):
Looks good.

git_machete/git_operations.py
```diff
@@ -373,41 +377,335 @@ def load_all_reflogs(self) -> None:
...
                 ]
-            return self.reflogs_cached[b]
+            return self.__reflogs_cached[b]
 
     def create_branch(self, b: str, out_of_revision: str) -> None:
         self.run_git("checkout", "-b", b, out_of_revision)
         self.flush_caches()  # the repository state has changed b/c of a successful branch creation, let's defensively flush all the caches
 
     def flush_caches(self) -> None:
         self.branch_defs_by_sha_in_reflog = None
```
PawelLipski: Huh this field is veeery Machete-specific... I suppose it's located in GitContext for the sake of easier cache flushing right?

PawelLipski: TBH I'd move this attr to `MacheteClient`, make private, and add a method `flush_caches` in `MacheteClient` that would clear `branch_defs_by_sha_in_reflog` and:
* either call `flush_caches()` on `GitContext` OR
* just discard the old `GitContext` and create a new one? that way `GitContext.flush_caches` method won't even be necessary

git_machete/cmd.py
```diff
@@ -718,15 +684,15 @@ def print_new_line(new_status: bool) -> None:
                     rmt: Optional[str] = self.git.inferred_remote_for_fetching_of_branch(b)
                     print_new_line(False)
                     if rmt:
-                        handle_untracked_branch(self.git, rmt, b)
+                        self.handle_untracked_branch(rmt, b)
```
PawelLipski, line 692: Shouldn't this method be private as well? 🤔

git_machete/cmd.py
```diff
@@ -391,11 +379,11 @@ def get_root_of(b: str) -> str:
         print(bold("Discovered tree of branch dependencies:\n"))
         self.status(warn_on_yellow_edges=False)
         print("")
-        do_backup = os.path.isfile(self.definition_file_path)
-        backup_msg = f"\nThe existing definition file will be backed up as {self.definition_file_path}~" if do_backup else ""
-        msg = f"Save the above tree to {self.definition_file_path}?{backup_msg}" + pretty_choices('y', 'e[dit]', 'N')
-        opt_yes_msg = f"Saving the above tree to {self.definition_file_path}... {backup_msg}"
-        ans = ask_if(self.cli_opts, msg, opt_yes_msg)
+        do_backup = os.path.isfile(self._definition_file_path)
```
PawelLipski, line 387: Hmm AFAICS `_definition_file_path` and `_managed_branches` are preceded with a single underscore... is it deliberate?

git_machete/cmd.py
```diff
@@ -67,30 +41,46 @@ class MacheteClient:
     def __init__(self, cli_opts: CommandLineOptions, git: GitContext) -> None:
         self.cli_opts: CommandLineOptions = cli_opts
         self.git: GitContext = git
-        self.definition_file_path: str = self.git.get_git_subpath("machete")
-        self.managed_branches: List[str] = []
+        self._definition_file_path: str = self.git.get_git_subpath("machete")
+        self._managed_branches: List[str] = []
         self.down_branches: Dict[str, List[str]] = {}  # TODO (#110): default dict with []
```
PawelLipski: Can the remaining 4 non-private attrs be made private as well?

git_machete/cmd.py
PawelLipski: Ok, `@property`

git_machete/cmd.py
```diff
@@ -67,30 +41,46 @@ class MacheteClient:
...
+        self.__indent: Optional[str] = None
+        self.__roots: List[str] = []
+        self.__annotations: Dict[str, str] = {}
+        self.__empty_line_status: Optional[bool] = None
+
+    @property
+    def definition_file_path(self) -> str:
+        return self._definition_file_path
+
+    @definition_file_path.setter
```
PawelLipski: Hmm is the setter needed here? 🤔

git_machete/cmd.py
```diff
@@ -67,30 +41,46 @@ class MacheteClient:
...
+    def definition_file_path(self, val: str) -> None:
+        self._definition_file_path = val
+
+    @property
+    def managed_branches(self) -> List[str]:
+        return self._managed_branches
+
+    @managed_branches.setter
+    def managed_branches(self, val: List[str]) -> None:
```
PawelLipski, line 63: Same here, do we need to expose the setter publicly?

git_machete/git_operations.py
```diff
@@ -19,19 +23,21 @@ def __init__(self, cli_opts: CommandLineOptions) -> None:
         self.git_version: Tuple[int, ...] = None
         self.root_dir: str = None
         self.git_dir: str = None
```
PawelLipski: Whoops the above 3 attrs are wrongly typed, they should be `Optional[<something>]` since you assign `None` to them

git_machete/cmd.py
rifel123: well, i found this method useful when it's used from API. Someone may want to sync certain branch only and not traverse whole repository or its parts.

git_machete/git_operations.py
rifel123: yes, it was to make flushing easier, in one place. Because `flush_caches` is used later in `GitContext`, i'd like to call `flush_caches` on `GitContext` from `MacheteClient`.

git_machete/cmd.py
rifel123: Since managed_branches fields is used also outside the class, i would stick to current solution to emulate protected field. Another option would be to make it private and make helper method to return it outside.

PawelLipski (approved):

PR #176, opened 2021-08-18
smooth out method naming

### LINKED ISSUE #160: Smooth out method naming
There's currently very little consistency in method naming: some use `get_` or `derive_` prefix while some don't, depending on nothing in particular. Let's tidy that up in either direction

garg-tanishka:
fixes #160

### DIFF
--- a/git_machete/cmd.py
+++ b/git_machete/cmd.py
@@ -50,7 +50,7 @@ def tupled(f: Callable[..., T]) -> Callable[[Any], T]:
     return lambda tple: f(*tple)
 
 
-def get_second(pair: Tuple[str, str]) -> str:
+def avoid_time_lag(pair: Tuple[str, str]) -> str:
     a, b = pair
     return b
 
@@ -626,7 +626,7 @@ def add(cli_ctxt: CommandLineContext, b: str) -> None:
         expect_in_managed_branches(onto)
 
     if b not in local_branches(cli_ctxt):
-        rb: Optional[str] = get_sole_remote_branch(cli_ctxt, b)
+        rb: Optional[str] = sole_remote_branch(cli_ctxt, b)
         if rb:
             common_line = f"A local branch `{b}` does not exist, but a remote branch `{rb}` exists.\n"
             msg = common_line + f"Check out `{b}` locally?" + pretty_choices('y', 'N')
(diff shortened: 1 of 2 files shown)


git_machete/cmd.py
```diff
@@ -50,7 +50,7 @@ def tupled(f: Callable[..., T]) -> Callable[[Any], T]:
     return lambda tple: f(*tple)
 
 
-def get_second(pair: Tuple[str, str]) -> str:
+def avoid_time_lag(pair: Tuple[str, str]) -> str:
```
PawelLipski, line 53: 😯 why specifically `avoid_time_lag`?

PawelLipski:
@garg-tanishka thanks for your contribution! pls check the CI above (`Log In with GitHub` to CircleCI), there are multiple errors from `mypy` type checker... you can run it locally by using `tox`, as outlined in CONTRIBUTING.md :)

git_machete/cmd.py
```diff
@@ -626,7 +626,7 @@ def add(cli_ctxt: CommandLineContext, b: str) -> None:
         expect_in_managed_branches(onto)
 
     if b not in local_branches(cli_ctxt):
-        rb: Optional[str] = get_sole_remote_branch(cli_ctxt, b)
+        rb: Optional[str] = sole_remote_branch(cli_ctxt, b)
```
BartoszSiwinski, line 629: sole_remote_branch name suggests that it is a variable containing branch, where in reality it is a method returning such. By removing 'get_' from the names of method it now looks like they these methods are variables. It's established programming convention to name methods as actions (like 'get_string_length' or 'notify_police') or questions (like 'is_even(number) or 'are_commits_in_the_same_format'). I don't recommend introducing this change and similar ones. Applies to all removal of 'get_' or 'is_' from methods and functions names.

BartoszSiwinski: Side Note, python itself is rather into the idea of using 'is' methods names. See https://docs.python.org/3/library/stdtypes.html#float.is_integer or https://docs.python.org/3/library/functions.html#isinstance

PawelLipski:
Also! please apply the changes onto `develop`, as `master` is used for keeping the released changes, and is pretty far behind `develop`

garg-tanishka:
So, should I close this PR?

PawelLipski:
Up to you! it might be more convenient for you to close this one, create a new branch out of `develop` and create a new PR indeed

garg-tanishka:
Yes. Thankyou!

Issue #171, opened 2021-08-11
Error message in case of `not a git repository` error is far too verbose


PawelLipski:
```
$ git machete 
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/site-packages/git_machete-3.3.1-py3.7.egg/git_machete/git_operations.py", line 116, in get_git_dir
  File "/usr/local/lib/python3.7/site-packages/git_machete-3.3.1-py3.7.egg/git_machete/git_operations.py", line 52, in popen_git
git_machete.exceptions.MacheteException: git rev-parse --git-dir returned 128

stderr:
fatal: not a git repository (or any parent up to mount point /)
Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).


During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/local/bin/git-machete", line 4, in <module>
    __import__('pkg_resources').run_script('git-machete==3.3.1', 'git-machete')
  File "/usr/lib/python3.7/site-packages/pkg_resources/__init__.py", line 666, in run_script
    self.require(requires)[0].run_script(script_name, ns)
  File "/usr/lib/python3.7/site-packages/pkg_resources/__init__.py", line 1469, in run_script
    exec(script_code, namespace, namespace)
  File "/usr/local/lib/python3.7/site-packages/git_machete-3.3.1-py3.7.egg/EGG-INFO/scripts/git-machete", line 4, in <module>
    __import__('pkg_resources').run_script('git-machete==3.3.1', 'git-machete')
  File "/usr/local/lib/python3.7/site-packages/git_machete-3.3.1-py3.7.egg/git_machete/cmd.py", line 1968, in main
  File "/usr/local/lib/python3.7/site-packages/git_machete-3.3.1-py3.7.egg/git_machete/cmd.py", line 1975, in launch
  File "/usr/local/lib/python3.7/site-packages/git_machete-3.3.1-py3.7.egg/git_machete/cmd.py", line 70, in __init__
    doc_header = "Documented commands (type help <topic>):"
  File "/usr/local/lib/python3.7/site-packages/git_machete-3.3.1-py3.7.egg/git_machete/git_operations.py", line 122, in get_git_subpath
  File "/usr/local/lib/python3.7/site-packages/git_machete-3.3.1-py3.7.egg/git_machete/git_operations.py", line 118, in get_git_dir
git_machete.exceptions.MacheteException: Not a git repository
```

[...]

PR #179, opened 2021-08-19
Fix wrong line number when branch reappears in definition file && limit stack trace in certain Exceptions

### LINKED ISSUE #153: Line numbers displayed in errors refer to non-empty lines, not all lines
.git/machete file:

```
master
  develop


  develop
```

The empty lines lead to an error in `git machete status`:

`.git/machete, line 3: branch develop re-appears in the tree definition. Edit the definition file manually with git machete edit`

while it should say:

`.git/machete, line 5: branch develop re-appears in the tree definition. Edit the definition file manually with git machete edit`

This is due to how `read_definition_file` is implemented, should be pretty easy to fix.

rifel123:
closes #153  and #171

### DIFF
--- a/git_machete/cmd.py
+++ b/git_machete/cmd.py
@@ -89,7 +89,7 @@ class MacheteClient:
 
     def read_definition_file(self, verify_branches: bool = True) -> None:
         with open(self._definition_file_path) as f:
-            lines: List[str] = [line.rstrip() for line in f.readlines() if not line.isspace()]
+            lines: List[str] = [line.rstrip() for line in f.readlines()]
 
         at_depth = {}
         last_depth = -1
@@ -98,6 +98,8 @@ class MacheteClient:
 
         invalid_branches: List[str] = []
         for index, line in enumerate(lines):
+            if line == "" or line.isspace():
+                continue
             prefix = "".join(itertools.takewhile(str.isspace, line))
             if prefix and not self.__indent:
                 self.__indent = prefix
--- a/git_machete/cmd.py
+++ b/git_machete/cmd.py
@@ -1821,7 +1821,6 @@ def launch(orig_args: List[str]) -> None:
 
     cli_opts = CommandLineOptions()
     git = GitContext(cli_opts)
-    #machete_client = MacheteClient(cli_opts, git)
 
     if sys.version_info.major == 2 or (sys.version_info.major == 3 and sys.version_info.minor < 6):
         version_str = f"{sys.version_info.major}.{sys.version_info.minor}.{sys.version_info.micro}"
--- a/git_machete/cmd.py
+++ b/git_machete/cmd.py
@@ -1821,7 +1821,7 @@ def launch(orig_args: List[str]) -> None:
 
     cli_opts = CommandLineOptions()
     git = GitContext(cli_opts)
-    machete_client = MacheteClient(cli_opts, git)
+    #machete_client = MacheteClient(cli_opts, git)
 
     if sys.version_info.major == 2 or (sys.version_info.major == 3 and sys.version_info.minor < 6):
         version_str = f"{sys.version_info.major}.{sys.version_info.minor}.{sys.version_info.micro}"
@@ -1969,6 +1969,7 @@ def launch(orig_args: List[str]) -> None:
             return in_args[0]
 
     try:
+        machete_client = MacheteClient(cli_opts, git)
         cmd = None
         # Let's first extract the common options like `--help` or `--verbose` that might appear BEFORE the command,
         # as in e.g. `git machete --verbose status`.
--- a/git_machete/exceptions.py
+++ b/git_machete/exceptions.py
@@ -1,9 +1,16 @@
+import sys
+
 from git_machete import utils
 
 
 class MacheteException(Exception):
-    def __init__(self, msg: str, apply_fmt: bool = True) -> None:
+    def __init__(self, msg: str, apply_fmt: bool = True, verbose: bool = True) -> None:
         self.parameter = utils.fmt(msg) if apply_fmt else msg
+        if not verbose:
+            sys.tracebacklimit = 0
+        else:
+            if hasattr(sys, 'tracebacklimit'):
+                del sys.tracebacklimit
 
     def __str__(self) -> str:
         return str(self.parameter)
--- a/git_machete/exceptions.py
+++ b/git_machete/exceptions.py
@@ -4,7 +4,7 @@ from git_machete import utils
 
 
 class MacheteException(Exception):
-    def __init__(self, msg: str, apply_fmt: bool = True, verbose: bool = True) -> None:
+    def __init__(self, msg: str, apply_fmt: bool = True, verbose: bool = False) -> None:
         self.parameter = utils.fmt(msg) if apply_fmt else msg
         if not verbose:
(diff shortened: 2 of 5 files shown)


git_machete/cmd.py
```diff
@@ -98,6 +98,8 @@ def read_definition_file(self, verify_branches: bool = True) -> None:
 
         invalid_branches: List[str] = []
         for index, line in enumerate(lines):
+            if line == "" or line.isspace():
```
PawelLipski, line 101: Whoops... so the previous impl did NOT sift out the lines that were **empty**?

PawelLipski: Generally! if you could add some regression test(s) that would fail on the previous impl but pass on the new one

git_machete/git_operations.py
```diff
@@ -111,15 +111,15 @@ def get_root_dir(self) -> str:
             try:
                 self.root_dir = self.popen_git("rev-parse", "--show-toplevel").strip()
             except MacheteException:
-                raise MacheteException("Not a git repository")
+                raise MacheteException("Not a git repository", verbose=False)
```
PawelLipski: Hmm okay... why specifically these 3 occurrences of `MacheteException` are created with `verbose=False`? Shouldn't `verbose=False` be the default actually?

rifel123: ok, that makes more sense :)

git_machete/cmd.py
rifel123: Previous impl did remove empty lines but before enumerate, so info about line number was lost in case there was empty line in .git-machete.

Ok, i'll add regression test for this issue.

git_machete/exceptions.py
```diff
@@ -4,7 +4,7 @@
 
 
 class MacheteException(Exception):
-    def __init__(self, msg: str, apply_fmt: bool = True, verbose: bool = True) -> None:
+    def __init__(self, msg: str, apply_fmt: bool = True, verbose: bool = False) -> None:
```
PawelLipski: Hmm can we go step further and just remove this param (and the associated `if` below)?

git_machete/exceptions.py
```diff
@@ -4,7 +4,7 @@
...
         self.parameter = utils.fmt(msg) if apply_fmt else msg
         if not verbose:
```
PawelLipski: Nit: if this `if` stays, pls flip the condition to remove the negation

git_machete/exceptions.py
```diff
@@ -1,9 +1,16 @@
+import sys
+
 from git_machete import utils
 
 
 class MacheteException(Exception):
```
PawelLipski: Also! back to the original issue:

image

I'd rather expect just `Not a git repository` error message... that's also how git-machete v3.3.0 behaves:

image

git_machete/github.py
```diff
@@ -163,7 +163,7 @@ def create_pull_request(org: str, repo: str, head: str, base: str, title: str, d
         pr = __fire_github_api_request('POST', f'/repos/{org}/{repo}/pulls', token, request_body)
         return __parse_pr_json(pr)
     else:
-        raise MacheteException(f'Pull request for branch {head} is already created under link {pr_found.html_url}!\nPR details: {pr_found}')
+        raise MacheteException(f'Pull request for branch {head} is already created under link {pr_found.html_url}!\nPR details: {pr_found}', verbose=False)
```
PawelLipski: `verbose=False` not needed?

git_machete/tests/functional/test_machete.py
```diff
@@ -175,6 +182,20 @@ def setup_discover_standard_tree(self) -> None:
...
         )
 
+    def test_branch_reappers_in_definition(self) -> None:
+        body: List[str] = ['master\n', '    develop\n', '    \n', '    \n', 'develop\n']
+        expected_error_msg: str = '.git/machete, line 5: branch `develop` re-appears in the tree definition. Edit the definition file manually with `git machete edit`'
+
+        self.repo_sandbox.new_branch("root")
+        self.edit_definition_file(body)
+
+        machete_client = cmd.MacheteClient(cli_opts, git)  # Only to workaround sys.exit while calling launch(['status])
```
PawelLipski:
```suggestion
        machete_client = cmd.MacheteClient(cli_opts, git)  # Only to workaround sys.exit while calling launch(['status'])
```

git_machete/tests/functional/test_machete.py
```diff
@@ -175,6 +182,20 @@ def setup_discover_standard_tree(self) -> None:
             """,
         )
 
+    def test_branch_reappers_in_definition(self) -> None:
+        body: List[str] = ['master\n', '    develop\n', '    \n', '    \n', 'develop\n']
```
PawelLipski: Pls include some empty lines as well! Also, this would read better as either list of strings split into multiple lines, or even better as a multiline string

PawelLipski: Also, are `\n`s at the end of each line intended? Aren't newline characters added automatically after each line by `writelines` method?

git_machete/github.py
rifel123: No, because stack trace is hidden for all cases now.

git_machete/exceptions.py
rifel123: outdated, if removed.

git_machete/exceptions.py
rifel123: whoops, solution was so unexpectedly simple, just move initialisation of MacheteClient couple line down to try clause.

git_machete/github.py
rifel123: this will also be handled by try in launch function

git_machete/tests/functional/test_machete.py
rifel123: I was also convinced that `\n` is added as separator but docs says something opposite. Nevertheless i changed it into multiline string.

git_machete/cmd.py
```diff
@@ -1821,7 +1821,7 @@ def launch(orig_args: List[str]) -> None:
 
     cli_opts = CommandLineOptions()
     git = GitContext(cli_opts)
-    machete_client = MacheteClient(cli_opts, git)
+    #machete_client = MacheteClient(cli_opts, git)
```
PawelLipski: Whoops okay, makes sense why the stack traces looked like that ;D

PawelLipski: Pls remove this line ofc

git_machete/exceptions.py
```diff
@@ -1,3 +1,5 @@
+import sys
```
PawelLipski: Unneeded?

git_machete/tests/functional/test_machete.py
```diff
@@ -175,6 +182,26 @@ def setup_discover_standard_tree(self) -> None:
             """,
         )
 
+    def test_branch_reappers_in_definition(self) -> None:
+        body: str = \
+            """master
+                develop
+                
+
```
PawelLipski: Hmm is there any _empty_ line here? AFAICS they have some whitespace in them

git_machete/github.py
PawelLipski: probably param to be removed now

PawelLipski (approved):

PR #180, opened 2021-08-19
Issue #142: Add test coverage reporting tool to CI of git-machete.

### LINKED ISSUE #142: Wire up test coverage reporting tool

BartoszSiwinski:
How test coverage reports look can be viewed in [here](https://app.circleci.com/pipelines/github/VirtusLab/git-machete/411/workflows/f67b45bf-61ab-47d2-846a-620f5b99f38f/jobs/416/artifacts).
I recommend starting with [index.html](https://416-122743101-gh.circle-artifacts.com/0/coverage/index.html) and than click from it to other.


.circleci/config.yml
```diff
@@ -3,7 +3,7 @@ version: 2.1
 jobs:
   build:
     machine:
-      image: ubuntu-2004:202101-01
+      image: ubuntu-2004:202107-02
```
PawelLipski, line 6: That was... weird 🤔 good that the fix was so easy

.circleci/config.yml
```diff
@@ -41,8 +41,11 @@ jobs:
       - run: PYTHON_VERSION=3.6 GIT_VERSION=1.8.0  bash ci/tox/ci-run.sh  # Earliest versions of python/git supported by git-machete
       - run: PYTHON_VERSION=3.7 GIT_VERSION=2.7.6  bash ci/tox/ci-run.sh
       - run: PYTHON_VERSION=3.8 GIT_VERSION=2.25.0 bash ci/tox/ci-run.sh
-      - run: PYTHON_VERSION=3.9 GIT_VERSION=2.30.1 bash ci/tox/ci-run.sh
+      - run: PYTHON_VERSION=3.9 GIT_VERSION=2.30.1 CHECK_COVERAGE="true" bash ci/tox/ci-run.sh
```
PawelLipski: Nit: quotes around `true` aren't necessary, the effect (i.e. the exact bytes passed in the env vars to the `bash` process) will be the same where you use `"`, `'` or no quotes.

This won't be so easy tho if the string wasn't `true` but contained spaces or other special characters which would be interpreted by bash, like `$` or `(`

PawelLipski: Rel: https://virtuslab.github.io/tips/#shell/input-output-redirections point 4

.circleci/config.yml
```diff
@@ -41,8 +41,11 @@ jobs:
...
 
+      - store_artifacts:
+          path: htmlcov
```
PawelLipski, line 55: CircleCI is RiGCz

.circleci/config.yml
```diff
@@ -41,8 +41,11 @@ jobs:
       - run: PYTHON_VERSION=3.6 GIT_VERSION=1.8.0  bash ci/tox/ci-run.sh  # Earliest versions of python/git supported by git-machete
       - run: PYTHON_VERSION=3.7 GIT_VERSION=2.7.6  bash ci/tox/ci-run.sh
```
PawelLipski: I'd add `name:` to specify a human-friendly name of the stage (like `Run tests against Python 3.x and git x.y.z (with coverage)`)

ci/tox/Dockerfile
```diff
@@ -24,6 +24,8 @@ RUN set -x \
 
 ARG python_version
 ENV PYTHON_VERSION=${python_version}
+ARG check_coverage
+ENV CHECK_COVERAGE=${check_coverage}
```
PawelLipski: Nit: move down since now `PYTHON_VERSION` and `PYTHON` are split

ci/tox/build-context/entrypoint.sh
```diff
@@ -8,7 +8,12 @@
 set -e -u -x
 
 $PYTHON -m pip install --user tox
-TOXENV="pep8,py${PYTHON_VERSION/./}" tox
+
+if [[ "${CHECK_COVERAGE}" = "true" ]]; then
```
PawelLipski: Nit:
```suggestion
if [[ $CHECK_COVERAGE = true ]]; then
```

ci/tox/docker-compose.yml
```diff
@@ -10,6 +10,7 @@ services:
         - group_id=${GROUP_ID:-0}
         - git_version=${GIT_VERSION:-0.0.0}
         - python_version=${PYTHON_VERSION:-0.0.0}
+        - check_coverage=${CHECK_COVERAGE:-"false"}
```
PawelLipski: Nit, this should be equivalent:
```suggestion
        - check_coverage=${CHECK_COVERAGE:-false}
```

tox.ini
```diff
@@ -30,5 +30,14 @@ commands = {posargs}
...
-exclude = ./.*,build,dist,*egg
+exclude = ./.*,build,dist,*egg,venv
 import-order-style = pep8
+
+[testenv:coverage]
+description = "Checking the test coverage of the code."
+deps = coverage
+commands =
+  coverage erase
+  coverage run --branch --include='git_machete/*' --omit='git_machete/tests/*' setup.py test
```
PawelLipski, line 41: Hmmm for clarity... against which Python version will this test be done? 🤔 Since this `coverage` env is distinct from `py36-py39`.

Actually, I now realize the same point applies to `pep8` tox env... if you could check 😅

ci/tox/Dockerfile
BartoszSiwinski: Ok, I will also add a comment explaining what it does since it does not have any reference in this file itself, but only in entrypoint.sh

tox.ini
BartoszSiwinski: I have checked `python --version` and `python3 --version`, since there I haven't found any explicit way to verify python version through flake8. I assume that default python is one that is being used since we don't specify anything else there. We could change flake8 to `python{PYTHON_VERSION} -m flake8` and `python{PYTHON_VERSION} -m coverage...`, so we explicitly run it using particular python.

py 3.6 - pep is run using python 3.6.14 as it is default python
py 3.7 - pep is run using 3.7.11 as it is default python
py 3.8 - pep is run using 3.8.11 as it is default python
py 3.9 - pep and coverage are run using 3.9.6 as it is default python

PawelLipski: Okay, sorry, I wasn't clear about what I mean here (or rather, I didn't think the question over enough):
* in CI setup, it's rather clear which Python version is used &mdash; the one that's present in the given step's Docker image
* but! in local setup, while for py36-py39 it's clear which Python version is used... which version of Python is used for `pep8` and `coverage`? Is it just whatever version is linked to `python` (`python3`?) executable in the system?

BartoszSiwinski: Both coverage and pep8 are run through tox, so I would expect that they're run using whatever is running tox. So if you just run `tox -e coverage` you will use python on which you installed tox. If there are more than one than you can check `tox --showconfig` and there are parameters in [tox] section - host_python and in [testenv:coverage] - basepython. You can also specify base_python for particular testenv.

PawelLipski: Ok! thanks, that explains it to me

ci/tox/build-context/entrypoint.sh
```diff
@@ -8,7 +8,12 @@
 set -e -u -x
 
 $PYTHON -m pip install --user tox
-TOXENV="pep8,py${PYTHON_VERSION/./}" tox
+
+if [[ $CHECK_COVERAGE = true ]]; then
+  TOXENV="pep8,py${PYTHON_VERSION/./},coverage" tox
```
PawelLipski: Also! since you've mentioned `-e` option to `tox`... would using this option instead of `TOXENV` improve readability here?

tox.ini
```diff
@@ -30,5 +30,14 @@ commands = {posargs}
 # W504 ignore wrap line after binary operator
 ignore = E501,W504
 show-source = True
-exclude = ./.*,build,dist,*egg
+exclude = ./.*,build,dist,*egg,venv
 import-order-style = pep8
+
+[testenv:coverage]
```
PawelLipski, line 36: After running `tox -e coverage`, I've got some untracked unignored files in the repo:

image

I think they should be added to .gitignore, and also, less obviously, as `tar-ignore` to `debian/source/options` (so that they're not included in Debian/Ubuntu packages)

ci/tox/build-context/entrypoint.sh
BartoszSiwinski: I think that it's a good idea.

PawelLipski (approved):

PR #182, opened 2021-08-23
Extract generic code across ci scripts

### LINKED ISSUE #138: Unify code for pulling/building/pushing Docker images for CI
There's currently a bunch of different `ci-run.sh` scripts in subfolders under `ci/`... most of them are doing the very same thing: pull the Docker image corresponding to the current state of the subfolder from DockerHub, or build&push the image if not found there. I think this can be unified into some utility script to be shared across these subfolders.

rifel123:
Implements #138


ci/ci-run.sh
```diff
@@ -4,21 +4,26 @@ set -e -o pipefail -u -x
...
-image_tag=$(docker-compose config | yq eval '.services.tox.image' -)
+image_tag=$(docker-compose config | yq eval ".services.$image_name.image" -)
 docker image inspect "$image_tag" &>/dev/null || {
-  docker-compose build --build-arg user_id="$(id -u)" --build-arg group_id="$(id -g)" tox
+  docker-compose build --build-arg user_id="$(id -u)" --build-arg group_id="$(id -g)" $image_name
   # In builds coming from forks, secret vars are unavailable for security reasons; hence, we have to skip pushing the newly built image.
   if [[ ${DOCKER_PASSWORD-} && ${DOCKER_USERNAME-} ]]; then
     echo "$DOCKER_PASSWORD" | docker login -u "$DOCKER_USERNAME" --password-stdin
-    docker-compose push tox
+    if [[ "$image_name" == "tox" ]]; then
```
PawelLipski: I'd rather remove the `if` here, let's just leave the version with `|| true` + retain the original comment from `ci/rpm/ci-run.sh` on why this is needed

.circleci/config.yml
```diff
@@ -98,7 +99,7 @@ jobs:
```
PawelLipski: Would it be possible to unify the other Dockerized deployment steps as well? I see `ci/deb-ppa-upload/ci-deploy.sh` as a potential candidate

PawelLipski: Although there's a slight difference there (at least wrt. one vs two `docker-compose run` calls), so probably some extra refactors will be needed, leaving that up to you

.circleci/config.yml
```diff
@@ -38,18 +38,19 @@ jobs:
             ensure-version-bumped.sh
             prohibit-tab-character.sh
             prohibit-trailing-whitespace.sh
+
       - run:
           name: Run tests against Python 3.6 and git 1.8.0 (without checking coverage)
-          command: PYTHON_VERSION=3.6 GIT_VERSION=1.8.0  bash ci/tox/ci-run.sh  # Earliest versions of python/git supported by git-machete
+          command: PYTHON_VERSION=3.6 GIT_VERSION=1.8.0  bash ci/ci-run.sh tox # Earliest versions of python/git supported by git-machete
```
PawelLipski: I'd consider renaming `ci-run.sh` to sth more specific now that its location changed... maybe `run-dockerized.sh`?

.circleci/config.yml
```diff
@@ -98,7 +99,7 @@ jobs:
 
             - deploy:
                 name: 'Build rpm package'
-                command: bash ci/rpm/ci-run.sh
+                command: bash ci/ci-run.sh rpm
```
PawelLipski: I'd also consider unifying the `local-run.sh` scripts if feasible

ci/ci-run.sh
```diff
@@ -4,21 +4,26 @@ set -e -o pipefail -u -x
...
 DIRECTORY_HASH=$(git rev-parse HEAD:ci/tox)
 export DIRECTORY_HASH
-cd ci/tox/
+image_name=$1
+cd ci/$image_name/
 
 # If there is no cached image for the expected Git&Python versions and the current state of ci/tox,
 # build the image and push it to the Docker Hub.
-docker-compose --no-ansi pull tox
+docker-compose --no-ansi pull $image_name
```
PawelLipski: While you're here, pls fix the following warn from `docker-compose` logs (see in the CI): `--no-ansi option is deprecated and will be removed in future versions. Use '--ansi never' instead.`

PawelLipski: Plus also pass this flag to other invocations of `docker-compose` in **CI scripts** (local runs are fine), to avoid weird formatting in the output like:

image

.circleci/config.yml
```diff
@@ -38,22 +38,33 @@ jobs:
...
       - run:
           name: Run tests against Python 3.6 and git 1.8.0 (without checking coverage)
-          command: PYTHON_VERSION=3.6 GIT_VERSION=1.8.0  bash ci/tox/ci-run.sh  # Earliest versions of python/git supported by git-machete
+          command: PYTHON_VERSION=3.6 GIT_VERSION=1.8.0  bash ci/run-dockerized.sh tox # Earliest versions of python/git supported by git-machete
       - run:
           name: Run tests against Python 3.7 and git 2.7.6 (without checking coverage)
-          command: PYTHON_VERSION=3.7 GIT_VERSION=2.7.6  bash ci/tox/ci-run.sh
+          command: PYTHON_VERSION=3.7 GIT_VERSION=2.7.6  bash ci/run-dockerized.sh tox
       - run:
           name: Run tests against Python 3.8 and git 2.25.0 (without checking coverage)
```
PawelLipski: Nit: while you're here, remove `(without checking coverage)`, feels too verbose (leave just `(with checking coverage)` below)

ci/local-run.sh
```diff
@@ -0,0 +1,46 @@
...
+image_name=$1
+
+if [[ $OSTYPE == 'darwin'* ]]; then  # group id 20 is usually taken in MacOS
+  group_ID="500"
+else
+  group_ID="$(id -g)"
+fi
+
+cd "$(git rev-parse --show-toplevel)"/ci/$image_name/
+if [[ "$image_name" == "tox" ]]; then
```
PawelLipski: I'd rather avoid `if`s like that, that generally leads to spaghetti code pretty quickly 🍝
Instead, I'd extract some generic scripts/functions that can be reused across the `local-run.sh` scripts (I think at least deriving `DIRECTORY_HASH` and running `docker-compose build` are good candidates)

ci/local-run.sh
```diff
@@ -0,0 +1,46 @@
...
+  cp .env-sample .env  # and optionally edit if needed to change git/python version
```
PawelLipski: Hmmm should this `cp` be done by a script? This way whatever the user kept in their .env file will be overwritten with the sample :/

ci/README.md
```diff
@@ -0,0 +1,21 @@
+## Building the RPM locally
```
PawelLipski: For consistency, the one README.md still left (`ci/deb-ppa-test-install/README.md`) should be moved here as well

ci/README.md
```diff
@@ -0,0 +1,21 @@
...
+./local-run.sh deb-ppa-upload
+```
+
+## Dockerized test against given git & Python version locally
+
+```shell script
+cp .env-sample .env  # and optionally edit if needed to change git/python version
+./local-run.sh tox
+```
+Note: optionally edit `ci/tox/.env-sample` file if needed to change git/python version.
```
PawelLipski: Hmmm `.env.sample` is **not** supposed to be edited 🤔 it's `.env` that's picked up by `docker-compose build`!

PawelLipski: Also, pls note that for that exact reason, `.env.sample` is tracked under git, while `.env` is git-ignored

ci/local-run.sh
```diff
@@ -0,0 +1,46 @@
...
+
+function check_var() {
+  [[ ${!1-} ]] && return 0 || true
+  [[ -f .env ]] || { echo "Var $1 missing from environment and no .env file found; see .env.sample file"; return 1; }
+  grep -q "^$1=" .env || { echo "Var $1 missing both from environment and .env file"; return 1; }
+}
+
+image_name=$1
+
+if [[ $OSTYPE == 'darwin'* ]]; then  # group id 20 is usually taken in MacOS
```
PawelLipski: Hmmm what's the specific problem that happened with group 20? It shouldn't be a problem if the group is _taken_ on the host system... in fact we want to specify `group_id` build ARG that's the indeed the same as the group id in the host system, in the spirit of https://medium.com/virtuslab/nifty-docker-tricks-for-your-ci-vol-2-c5191a67f1a4 -> Ctrl+F `if only`

ci/local-run.sh
PawelLipski: (I'd opt for shell functions here, these things seem too trivial to merit separate scripts TBH)

ci/run-dockerized.sh
```diff
@@ -0,0 +1,38 @@
+#!/usr/bin/env bash
+
+set -e -o pipefail -u -x
+
+image_name=$1
+
+if [[ ${2-} == "--dry-run" || ${CIRCLE_BRANCH-} != "master" ]]; then
+  do_dput=false
```
PawelLipski: In the same spirit as in https://github.com/VirtusLab/git-machete/pull/182/files#r694837997 ... I'd avoid making "switches" against the individual known cases... I'd rather provide a set of reusable scripts/functions.

This can in particular mean that the specific `ci-{run,deploy}.sh` scripts in `ci/{tox,rpm,...}` folders might need to be restored... which is okay IMO.

ci/run-dockerized.sh
```diff
@@ -0,0 +1,38 @@
...
+cd ci/$image_name/
+
+# If there is no cached image for the expected Git&Python versions and the current state of ci/tox,
+# build the image and push it to the Docker Hub.
+docker-compose --ansi never pull $image_name
+# A very unpleasant workaround for https://github.com/docker/compose/issues/7258
+# (since v1.25.1, `docker-compose pull` is NOT failing when it can't fetch the image).
+image_tag=$(docker-compose config | yq eval ".services.$image_name.image" -)
+docker image inspect "$image_tag" &>/dev/null || {
+  docker-compose build --build-arg user_id="$(id -u)" --build-arg group_id="$(id -g)" $image_name
```
PawelLipski: Nit: `--ansi never` for each `docker-compose` run (unless there're no weird artifacts in the CI output from this invocation, but IIRC there are)

PawelLipski: Btw FYI, `ansi` here is a mental shortcut for https://en.wikipedia.org/wiki/ANSI_escape_code ... there are zillions of them, but the most important ones are for colors (we use a few of them in `constants.py`)

ci/tox/local-run.sh
```diff
@@ -8,18 +8,20 @@ function check_var() {
   grep -q "^$1=" .env || { echo "Var $1 missing both from environment and .env file"; return 1; }
 }
```
PawelLipski, line 6: Hmm generally `ci/tox/local-run.sh` can probably be retained indeed... just with the remark about the reusable functions

git_machete/cmd.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
```
PawelLipski: Hmmm that's a potentially risky move, there are two alternatives:
* running `python` and then raising an error if it happens to be Python 2, see what we do in cmd.py at the beginning of `def launch`
* running `python3` and then ending up with `command not found` if the system doesn't happen to have `python3` executable, only `python` (which might correspond to Python 3 btw :trollface:)... I've seen that on a Windows recently ;p

PawelLipski: Was there any specific reason for this change? 🤔

git_machete/github.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
```
PawelLipski: Hmmm actually, do we even need a shebang in any Python file, other than `git_machete/cmd.py` and the `git-machete` launcher script and in top-level folder?

ci/local-run.sh
rifel123: Screenshot 2021-08-25 at 07 37 54
AFAICS this is quite common on mac..

git_machete/cmd.py
rifel123: hmm, that breaks rpm package build:
Screenshot 2021-08-25 at 09 13 54

rifel123: according to : https://fedoraproject.org/wiki/Changes/Make_ambiguous_python_shebangs_error
this was previously a warning only

ci/local-run.sh
PawelLipski: Hmmm okay but this is failing within the Docker image, so the error is caused by group id 20 being taken in the **base Docker image** and not in Mac OS ;)

PawelLipski: I think the easiest fix here would be to first check in Dockerfile whether the group to be created exists, and skip the creation if it already does :)

git_machete/cmd.py
PawelLipski: Okay! didn't know about that 😯

ci/local-run.sh
rifel123: I resigned from implementing this.  Check if group exist can be done easily but later on, while calling `adduser` i need to pass exact name of group. This is because `adduser` in alpine comes from busybox and `--gid` option is not recognized. On the other hand i cannot also retrieve exact name of group with `grep -w $group_id /etc/group | cut -d: -f4`, because if group was created in earlier step, this info is not yet available.

rifel123: This is the solution i could find and work in every case but it's far from being pretty:
Screenshot 2021-08-25 at 15 11 05

PawelLipski (commented):
Very close to ready, minor comments left :)

PawelLipski (approved):

PR #187, opened 2021-08-25
Issue #146 Optimise tests in CI to shorten the execution.

### LINKED ISSUE #146: Optimize tests in CI so that `tox` isn't installed separately for each Python version
An obvious solution to speed up installation would be to include `tox` within Docker image used for running the tests... but this in turn will increase the time needed for downloading the images themselves.

Let's check if this brings any speedup and think about the alternatives if it doesn't.

BartoszSiwinski:


.circleci/config.yml
```diff
@@ -38,80 +42,128 @@ jobs:
...
+      # 1) every time a commit is pushed to `master`, a Snap build for `edge` channel
+      #    is triggered automatically via GitHub webhook on https://build.snapcraft.io/user/VirtusLab/git-machete,
+      # 2) every time a https://github.com/r-ryantm bot runs on NixOS/nixpkgs repository,
+      #    Nix git-machete package is updated with the fresh version/package hash according to the data from PyPI.
+
+workflows:
+  version: 2
+  test-deploy:
+    jobs:
+      - initial_checks
```
PawelLipski, line 145: Just a loose suggestion, might make sense or not... maybe create a no-op/empty job with a super-lightweight executor that would depend on all of `initial_checks...tests_py39`? Then it can be used as both the required check for Github and as `requires` for `deploy_from_develop`/`deploy_from_master`

.circleci/config.yml
```diff
@@ -38,80 +42,128 @@ jobs:
...
+      - tests_py36
+      - tests_py37
+      - tests_py38
+      - tests_py39
+      - deploy_from_develop:
```
PawelLipski: No deployment actually happens... maybe `dry_run_deploy_from_develop` for clarity?

PawelLipski (approved):
Approving, subject to sorting out the required checks on Github... probably some re-run should be enough now that CircleCI Checks has been installed in our organization.

.circleci/config.yml
```diff
@@ -38,80 +42,128 @@ jobs:
...
+
+      # Note: there are also releases happening outside of our CI:
+      # 1) every time a commit is pushed to `master`, a Snap build for `edge` channel
+      #    is triggered automatically via GitHub webhook on https://build.snapcraft.io/user/VirtusLab/git-machete,
+      # 2) every time a https://github.com/r-ryantm bot runs on NixOS/nixpkgs repository,
+      #    Nix git-machete package is updated with the fresh version/package hash according to the data from PyPI.
+
+workflows:
+  version: 2
+  test-deploy:
```
PawelLipski: I'd call it just `build`, `test-deploy` suggests that _deployment is being tested_ (?), which is only the case for `develop` builds

.circleci/config.yml
BartoszSiwinski: I agree, that's better.

.circleci/config.yml
BartoszSiwinski: It will look better in the code, but it will skew the circle ci diagram (in reality deployment runs are waiting for tests, not for empty step) and it will drop the performance a bit, because CI still have to assign a VM to it, which sometimes takes a moment.

PR #181, opened 2021-08-23
Improve/method names

garg-tanishka:
#160

Issue #178, opened 2021-08-19
Add test for `github` command


rifel123:
Since `github` command has been added in version 3.4.0, there is need to  add tests for following commands:
- `github create-pr`
- `github anno-prs`
- `github retarget-pr`

Can be done together or separately with issue #170 

EDIT: plus also `github checkout-pr` once it's implemented within #175 (preferably tests added alongside the new subcommand)

PawelLipski:
This would ideally include opening a local server that would mimic the parts of github API that we're interested in... although this might be hard without a full Github on-premises instance 😅 maybe sth like https://github.com/altany/docker-github-api ? Or just a mock http server that would replay pre-recorded responses?

PawelLipski:
If not, then opening an test repo on github (within @git-machete-bot user) is also an option, albeit potentially a more unstable/less reproducible one

rifel123:
Closed with PR #214

PR #201, opened 2021-08-31
Fix recent regression in github.py (token taken only from GITHUB_TOKEN and not from `gh` or `hub`)

PawelLipski:


git_machete/github.py
```diff
@@ -148,7 +151,7 @@ def __check_pr_already_created(pull: GitHubPullRequest, pull_requests: List[GitH
 
 
 def create_pull_request(org: str, repo: str, head: str, base: str, title: str, description: str, draft: bool) -> GitHubPullRequest:
-    token: Optional[str] = _get_token_from_env()
```
PawelLipski, line 151: WHOOPS

PawelLipski: A regression that happened on https://github.com/VirtusLab/git-machete/pull/181/files#diff-ede0d58ccd3510984ee2caca86265a8d8df90c0c108e1a8208611a524cc1b66a ... I didn't notice that

PawelLipski: Which is a lesson that `github` command should be tested - also against failure scenarios - before it's released... I've moved https://github.com/VirtusLab/git-machete/issues/178 to the current milestone... @rifel123 for you to research the options on how this can be tested

git_machete/github.py
```diff
@@ -125,14 +126,16 @@ def __fire_github_api_request(method: str, url: str, token: Optional[str], reque
     except HTTPError as err:
         if err.code == http.HTTPStatus.UNPROCESSABLE_ENTITY:
             raise UnprocessableEntityHTTPError(str(err.reason))
-        elif err.code in (http.HTTPStatus.UNAUTHORIZED, http.HTTPStatus.FORBIDDEN):
-            first_line = fmt(f'GitHub API returned {err.code} HTTP status with error message: `{err.reason}`\n')
+        elif err.code in (http.HTTPStatus.UNAUTHORIZED, http.HTTPStatus.FORBIDDEN, http.HTTPStatus.NOT_FOUND):
+            first_line = f'GitHub API returned {err.code} HTTP status with error message: `{err.reason}`\n'
             if token:
-                raise MacheteException(first_line + fmt(f'Make sure that the token provided in `gh auth status` or `~/.config/hub` or <b>{GITHUB_TOKEN_ENV_VAR}</b> is valid and allows for access to `{method.upper()}` https://{host}{url}`.'))
```
PawelLipski, line 131: `fmt` not needed here, see `MacheteException.__init__`

git_machete/github.py
```diff
@@ -125,14 +126,16 @@ def __fire_github_api_request(method: str, url: str, token: Optional[str], reque
...
+                raise MacheteException(first_line + f'Make sure that the token provided in `gh auth status` or `~/.config/hub`'
+                                                    f' or <b>{GITHUB_TOKEN_ENV_VAR}</b> is valid and allows for access to `{method.upper()}` https://{host}{path}`.')
             else:
                 raise MacheteException(
-                    first_line + fmt(f'This repository might be private. Provide a GitHub API token with `repo` access via `gh` or `hub` or <b>{GITHUB_TOKEN_ENV_VAR}</b> env var.\n'
-                                     'Visit `https://github.com/settings/tokens` to generate a new one.'))
+                    first_line + f'You might not have the required permissions for this repository. '
```
PawelLipski, line 136: This might as well happen for a public repository... when you try pushing without a valid token

git_machete/github.py
```diff
@@ -116,7 +113,11 @@ def __fire_github_api_request(method: str, url: str, token: Optional[str], reque
         headers['Authorization'] = 'Bearer ' + token
 
     host = 'https://api.' + GITHUB_DOMAIN
-    http_request = Request(host + url, headers=headers, data=json.dumps(request_body).encode(), method=method.upper())
+    url = host + path
+    json_body: Optional[str] = json.dumps(request_body) if request_body else None
+    http_request = Request(url, headers=headers, data=json_body.encode() if json_body else None, method=method.upper())
```
PawelLipski, line 118: A little fix: previously, if `None` was passed as `request_body`, `data` would be the string `"null"` (as this is a JSON value corresponding to Python `None`) rather than `None`

git_machete/github.py
```diff
@@ -44,69 +44,66 @@ def __parse_pr_json(pr_json: Any) -> GitHubPullRequest:
...
+    def get_token_from_env() -> Optional[str]:
+        return os.environ.get(GITHUB_TOKEN_ENV_VAR)
 
-def _get_github_token() -> Optional[str]:
-    return _get_token_from_env() or __get_token_from_gh() or _get_token_from_hub()
+    return get_token_from_env() or get_token_from_gh() or get_token_from_hub()
 
 
-def __fire_github_api_request(method: str, url: str, token: Optional[str], request_body: Optional[Dict[str, Any]] = None) -> Any:
+def __fire_github_api_request(method: str, path: str, token: Optional[str], request_body: Optional[Dict[str, Any]] = None) -> Any:
```
PawelLipski, line 106: A little clarification, since it wasn't a full URL that was needed here, just the path

git_machete/github.py
```diff
@@ -44,69 +44,66 @@ def __parse_pr_json(pr_json: Any) -> GitHubPullRequest:
     return GitHubPullRequest(int(pr_json['number']), pr_json['user']['login'], pr_json['base']['ref'], pr_json['head']['ref'], pr_json['html_url'])
 
 
-def __get_token_from_gh() -> Optional[str]:
+def __get_github_token() -> Optional[str]:
+    def get_token_from_gh() -> Optional[str]:
```
PawelLipski, line 48: Made them into inner methods, this way nobody will make this mistake again 😅

rifel123 (approved):

PR #204, opened 2021-09-01
Reduce the amount of spam displayed

garg-tanishka:
#183 Before the number of lines was - 930, Now they are 746

Issue #161, opened 2021-08-03
Decouple `MacheteClient` and `GitContext` API from command-line options


PawelLipski:
Imagine someone wants to call a method on `MacheteClient` programatically... currently to create an instance on `MacheteClient`, they need to pass the entire `CommandLineContext` (`CommandLineOptions`?) object, even though their use case has nothing to do with CLI.

Let's consider decoupling `CommandLineContext` from what's passed to API methods. `MacheteClient` should know nothing about `CommandLineContext`; probably each method should only receive the options that are relevant to its execution, and the naming of this options doesn't need to exactly follow the naming of command-line options.

BartoszSiwinski:
How about instead of decoupling it, we could make it in such way that `MacheteClient` itself is creating instance of `GitContext` and `CommandLineOptions`?

I would propose also to reconsider what is called in the `launch` method in `cli.py`. We could `get a parser` there, `parse the arguments` using it and than `pass the parsed arguments` as only input parameter to `MacheteClient.__init__()`. After that everything else would happen inside `MacheteClient` including recognizing what cmd was called.

PawelLipski:
Hmmm IIUC that problem then would be that `MacheteClient` would be again strongly coupled to CLI, right? 🤔 for the sake of making the graph of dependencies less dense/easier to grasp, my idea was to make `MacheteClient` as oblivious to the existence of command line as possible

PawelLipski:
See e.g. places where `cli_options` are set before calling method `add` just so that the behavior of `add` is changed (`git_machete/client.py:1708`)... this should instead be done by passing an argument to `add`

PR #205, opened 2021-09-02
Introduce github checkout-pr command

### LINKED ISSUE #175: Add `git machete github checkout-pr` subcommand
This will be significantly complex since it ideally should be able to traverse the PR chain upwards, adding parents to the branch layout one by one

rifel123:
Known bug until now:
when `checkout-pr` detects opened PR on parent branches, it will try to add the chain of branches into definition file and also checkout them locally. then user is asked if he wants to add those branches one by one, and if he answers no in the middle the adding next branch will cause error because `onto` branch does not exists for git-machete

### DIFF
--- a/README.md
+++ b/README.md
@@ -156,6 +156,11 @@ Annotate the branches with GitHub PR numbers: <br/>
 git machete github anno-prs
 ```
 
+Checks out the PR into local branch, if pull request is found on any parent branch, it will add every parent branch up to root branch as seen by git-machete: <br/>
+```shell script
+git machete github checkout-pr [PR_number]
+```
+
 Create the PR, using the upstream (parent) branch from `.git/machete` as the base: <br/>
 ```shell script
 git machete github create-pr [--draft]
--- a/README.md
+++ b/README.md
@@ -158,7 +158,7 @@ git machete github anno-prs
 
 Check out the PR into local branch, also traverse chain of pull requests upwards, adding branches one by one to git-machete and check them out locally: <br/>
 ```shell script
-git machete github checkout-pr <PR_number>
+git machete github checkout-prs <PR_number>
 ```
 
 Create the PR, using the upstream (parent) branch from `.git/machete` as the base: <br/>
--- a/README.md
+++ b/README.md
@@ -156,7 +156,7 @@ Annotate the branches with GitHub PR numbers: <br/>
 git machete github anno-prs
 ```
 
-Checks out the PR into local branch, if pull request is found on any parent branch, it will add every parent branch up to root branch as seen by git-machete: <br/>
+Check out the PR into local branch, also traverse chain of pull requests upwards, adding branches one by one to git-machete and check them out locally: <br/>
 ```shell script
 git machete github checkout-pr [PR_number]
 ```
--- a/README.md
+++ b/README.md
@@ -158,7 +158,7 @@ git machete github anno-prs
 
 Check out the PR into local branch, also traverse chain of pull requests upwards, adding branches one by one to git-machete and check them out locally: <br/>
 ```shell script
-git machete github checkout-pr [PR_number]
+git machete github checkout-pr <PR_number>
 ```
 
 Create the PR, using the upstream (parent) branch from `.git/machete` as the base: <br/>
--- a/completion/git-machete.completion.bash
+++ b/completion/git-machete.completion.bash
@@ -21,6 +21,7 @@ _git_machete() {
   local discover_opts="-C --checked-out-since= -l --list-commits -r --roots= -y --yes"
   local fork_point_opts="--inferred --override-to= --override-to-inferred --override-to-parent --unset-override"
   local github_create_pr_opts="--draft"
+  local github_checkout_pr_opts="PR_number"
   local reapply_opts="-f --fork-point="
   local slide_out_opts="-d --down-fork-point= -M --merge -n --no-edit-merge --no-interactive-rebase"
   local squash_opts="-f --fork-point="
@@ -48,6 +49,8 @@ _git_machete() {
         github)
           if [[ ${COMP_WORDS[3]} == "create-pr" ]]; then
             __gitcomp "$common_opts $github_create_pr_opts"
+          elif [[ ${COMP_WORDS[3]} == "checkout-pr" ]]; then
+            __gitcomp "$common_opts $github_checkout_pr_opts"
           else
             __gitcomp "$common_opts"
           fi ;;
@@ -94,6 +97,8 @@ _git_machete() {
                   __gitcomp "$github_subcommands"
                 elif [[ ${COMP_WORDS[3]} == "create-pr" ]]; then
                   __gitcomp "$common_opts $github_create_pr_opts"
+                elif [[ ${COMP_WORDS[3]} == "checkout-pr" ]]; then
+                  __gitcomp "$common_opts $github_checkout_pr_opts"
                 else
                   COMPREPLY=('')
                 fi ;;
--- a/completion/git-machete.completion.bash
+++ b/completion/git-machete.completion.bash
@@ -6,7 +6,7 @@ _git_machete() {
 
   local categories="addable managed slidable slidable-after unmanaged with-overridden-fork-point"
   local directions="down first last next prev root up"
-  local github_subcommands="anno-prs checkout-pr create-pr retarget-pr"
+  local github_subcommands="anno-prs checkout-prs create-pr retarget-pr"
   local locations="current $directions"
   local opt_color_args="always auto never"
   local opt_return_to_args="here nearest-remaining stay"
@@ -94,8 +94,6 @@ _git_machete() {
                   __gitcomp "$github_subcommands"
                 elif [[ ${COMP_WORDS[3]} == "create-pr" ]]; then
                   __gitcomp "$common_opts $github_create_pr_opts"
-                elif [[ ${COMP_WORDS[3]} == "checkout-pr" ]]; then
-                  __gitcomp "$common_opts $github_checkout_pr_opts"
                 else
                   COMPREPLY=('')
                 fi ;;
--- a/completion/git-machete.completion.bash
+++ b/completion/git-machete.completion.bash
@@ -6,7 +6,7 @@ _git_machete() {
 
   local categories="addable managed slidable slidable-after unmanaged with-overridden-fork-point"
   local directions="down first last next prev root up"
-  local github_subcommands="anno-prs create-pr retarget-pr"
+  local github_subcommands="anno-prs checkout-pr create-pr retarget-pr"
   local locations="current $directions"
   local opt_color_args="always auto never"
   local opt_return_to_args="here nearest-remaining stay"
@@ -21,7 +21,6 @@ _git_machete() {
   local discover_opts="-C --checked-out-since= -l --list-commits -r --roots= -y --yes"
   local fork_point_opts="--inferred --override-to= --override-to-inferred --override-to-parent --unset-override"
   local github_create_pr_opts="--draft"
-  local github_checkout_pr_opts="PR_number"
   local reapply_opts="-f --fork-point="
   local slide_out_opts="-d --down-fork-point= -M --merge -n --no-edit-merge --no-interactive-rebase"
   local squash_opts="-f --fork-point="
@@ -49,8 +48,6 @@ _git_machete() {
         github)
           if [[ ${COMP_WORDS[3]} == "create-pr" ]]; then
             __gitcomp "$common_opts $github_create_pr_opts"
-          elif [[ ${COMP_WORDS[3]} == "checkout-pr" ]]; then
-            __gitcomp "$common_opts $github_checkout_pr_opts"
           else
             __gitcomp "$common_opts"
           fi ;;
--- a/completion/git-machete.completion.zsh
+++ b/completion/git-machete.completion.zsh
@@ -217,7 +217,7 @@ __git_machete_github_subcommands() {
   local github_subcommands
   github_subcommands=(
     'anno-prs:annotate the branches based on their corresponding GitHub PR numbers and authors'
-    'checkout-pr:check out the given pull request locally'
+    'checkout-prs:check out the given pull request locally'
     'create-pr:create a PR for the current branch, using the upstream (parent) branch as the PR base'
     'retarget-pr:set the base of the current branch PR to upstream (parent) branch'
   )
--- a/completion/git-machete.completion.zsh
+++ b/completion/git-machete.completion.zsh
@@ -64,11 +64,11 @@ _git-machete() {
         (g|go)
           _arguments '1:: :__git_machete_directions_go' && ret=0
           ;;
-        (github)
-          _arguments '1:: :__git_machete_github_subcommands' && ret=0
-          ;;
         (help)
           _arguments '1:: :__git_machete_help_topics' && ret=0
+          ;;
+        (github)
+          _arguments '1:: :__git_machete_github_subcommands' && ret=0
           ;;
         (is-managed|l|log)
           _arguments '1:: :__git_branch_names' && ret=0
@@ -217,6 +217,7 @@ __git_machete_github_subcommands() {
   local github_subcommands
   github_subcommands=(
     'anno-prs:annotate the branches based on their corresponding GitHub PR numbers and authors'
+    'checkout-pr:checks out locally given pull request'
     'create-pr:create a PR for the current branch, using the upstream (parent) branch as the PR base'
     'retarget-pr:set the base of the current branch PR to upstream (parent) branch'
   )
--- a/completion/git-machete.completion.zsh
+++ b/completion/git-machete.completion.zsh
@@ -64,11 +64,11 @@ _git-machete() {
         (g|go)
           _arguments '1:: :__git_machete_directions_go' && ret=0
           ;;
-        (help)
-          _arguments '1:: :__git_machete_help_topics' && ret=0
-          ;;
         (github)
           _arguments '1:: :__git_machete_github_subcommands' && ret=0
+          ;;
+        (help)
+          _arguments '1:: :__git_machete_help_topics' && ret=0
           ;;
         (is-managed|l|log)
           _arguments '1:: :__git_branch_names' && ret=0
@@ -217,7 +217,7 @@ __git_machete_github_subcommands() {
   local github_subcommands
   github_subcommands=(
     'anno-prs:annotate the branches based on their corresponding GitHub PR numbers and authors'
-    'checkout-pr:checks out locally given pull request'
+    'checkout-pr:check out the given pull request locally'
     'create-pr:create a PR for the current branch, using the upstream (parent) branch as the PR base'
     'retarget-pr:set the base of the current branch PR to upstream (parent) branch'
   )
--- a/docs/source/cli_help/github.rst
+++ b/docs/source/cli_help/github.rst
@@ -8,7 +8,7 @@ github
 
     git machete github <subcommand>
 
-where ``<subcommand>`` is one of: ``anno-prs``, ``create-pr``, ``retarget-pr``.
+where ``<subcommand>`` is one of: ``anno-prs``, ``checkout-pr``, ``create-pr``, ``retarget-pr``.
 
 Creates, checks out and manages GitHub PRs while keeping them reflected in branch definition file.
 
@@ -24,6 +24,16 @@ a GitHub API token with ``repo`` scope is required, see https://github.com/setti
   Annotates the branches based on their corresponding GitHub PR numbers and authors.
   Any existing annotations are overwritten for the branches that have an opened PR; annotations for the other branches remain untouched.
   Equivalent to ``git machete anno --sync-github-prs``.
+
+``checkout-pr``:
+
+  Check out given pull request (by number) locally under head branch name of pull request,
+  also traverse chain of pull requests upwards, adding branches one by one to git-machete and check them out locally.
+  Once pull request is checked-out locally, modify definition file according to structure found and synchronize pull requests annotations.
+
+  **Parameters:**
+
+    <PR-number>    Pull request number to checkout.
 
 ``create-pr [--draft]``:
 
--- a/docs/source/cli_help/github.rst
+++ b/docs/source/cli_help/github.rst
@@ -27,9 +27,9 @@ a GitHub API token with ``repo`` scope is required, see https://github.com/setti
 
 ``checkout-pr``:
 
-  Check out given pull request (by number) locally under head branch name of pull request,
+  Check out the head branch of the given pull request (specified by number),
   also traverse chain of pull requests upwards, adding branches one by one to git-machete and check them out locally.
-  Once pull request is checked-out locally, modify definition file according to structure found and synchronize pull requests annotations.
+  Once pull request is checked out locally, add new branches to the definition file according to the PR structure and annotate them with pull request numbers.
 
   **Parameters:**
 
--- a/docs/source/cli_help/github.rst
+++ b/docs/source/cli_help/github.rst
@@ -8,7 +8,7 @@ github
 
     git machete github <subcommand>
 
-where ``<subcommand>`` is one of: ``anno-prs``, ``checkout-pr``, ``create-pr``, ``retarget-pr``.
+where ``<subcommand>`` is one of: ``anno-prs``, ``checkout-prs``, ``create-pr``, ``retarget-pr``.
 
 Creates, checks out and manages GitHub PRs while keeping them reflected in branch definition file.
 
@@ -25,11 +25,11 @@ a GitHub API token with ``repo`` scope is required, see https://github.com/setti
   Any existing annotations are overwritten for the branches that have an opened PR; annotations for the other branches remain untouched.
   Equivalent to ``git machete anno --sync-github-prs``.
 
-``checkout-pr``:
+``checkout-prs``:
 
   Check out the head branch of the given pull request (specified by number),
   also traverse chain of pull requests upwards, adding branches one by one to git-machete and check them out locally.
-  Once pull request is checked out locally, add new branches to the definition file according to the PR structure and annotate them with pull request numbers.
+  Once pull request is checked out locally, annotate local branches with corresponding pull request numbers.
 
   **Parameters:**
 
--- a/git_machete/cli.py
+++ b/git_machete/cli.py
@@ -423,8 +423,8 @@ def launch(orig_args: List[str]) -> None:
             elif len(list_args) > 2:
                 raise MacheteException(f"Too many arguments to `git machete list {list_args[0]}` ")
             elif (list_args[0] in (
-                 "addable", "managed", "slidable", "unmanaged", "with-overridden-fork-point") and\
-                 len(list_args) > 1):
+                    "addable", "managed", "slidable", "unmanaged", "with-overridden-fork-point") and
+                  len(list_args) > 1):
                 raise MacheteException(
                     f"`git machete list {list_args[0]}` does not expect extra arguments")
             elif list_args[0] == "slidable-after" and len(list_args) != 2:
--- a/git_machete/cli.py
+++ b/git_machete/cli.py
@@ -365,7 +365,7 @@ def launch(orig_args: List[str]) -> None:
             if dest != current_branch:
                 git.checkout(dest)
         elif cmd == "github":
-            github_allowed_subcommands = "anno-prs|create-pr|retarget-pr|checkout-pr"
+            github_allowed_subcommands = "anno-prs|checkout-pr|create-pr|retarget-pr"
             list_args = parse_options(args) if args[0] == 'checkout-pr' \
                 else check_required_param(parse_options(args, "", ["draft"]), github_allowed_subcommands)
             if not list_args:
@@ -375,14 +375,6 @@ def launch(orig_args: List[str]) -> None:
             machete_client.read_definition_file()
             if param == "anno-prs":
                 machete_client.sync_annotations_to_github_prs()
-            elif param == "create-pr":
-                check_required_param(parse_options(args, "", ["draft"]), github_allowed_subcommands)
-                current_branch = git.get_current_branch()
-                machete_client.create_github_pr(current_branch, draft=cli_opts.opt_draft)
-            elif param == "retarget-pr":
-                current_branch = git.get_current_branch()
-                machete_client.expect_in_managed_branches(current_branch)
-                machete_client.retarget_github_pr(current_branch)
             elif param == "checkout-pr":
                 if len(list_args) == 1:
                     raise MacheteException(
@@ -392,6 +384,14 @@ def launch(orig_args: List[str]) -> None:
                 except ValueError:
                     raise MacheteException("PR number is not integer value!")
                 machete_client.checkout_github_pr(pr_no)
+            elif param == "create-pr":
+                check_required_param(parse_options(args, "", ["draft"]), github_allowed_subcommands)
+                current_branch = git.get_current_branch()
+                machete_client.create_github_pr(current_branch, draft=cli_opts.opt_draft)
+            elif param == "retarget-pr":
+                current_branch = git.get_current_branch()
+                machete_client.expect_in_managed_branches(current_branch)
+                machete_client.retarget_github_pr(current_branch)
             else:
                 raise MacheteException(
                     f"`github` requires a subcommand: one of `{github_allowed_subcommands}`")
--- a/git_machete/cli.py
+++ b/git_machete/cli.py
@@ -365,19 +365,33 @@ def launch(orig_args: List[str]) -> None:
             if dest != current_branch:
                 git.checkout(dest)
         elif cmd == "github":
-            github_allowed_subcommands = "anno-prs|create-pr|retarget-pr"
-            param = check_required_param(parse_options(args, "", ["draft"]),
-                                         github_allowed_subcommands)
+            github_allowed_subcommands = "anno-prs|create-pr|retarget-pr|checkout-pr"
+            list_args = parse_options(args) if args[0] == 'checkout-pr' \
+                else check_required_param(parse_options(args, "", ["draft"]), github_allowed_subcommands)
+            if not list_args:
+                raise MacheteException(f"`git machete github` expects argument(s): {github_allowed_subcommands}")
+
+            param = args[0]
             machete_client.read_definition_file()
             if param == "anno-prs":
                 machete_client.sync_annotations_to_github_prs()
             elif param == "create-pr":
+                check_required_param(parse_options(args, "", ["draft"]), github_allowed_subcommands)
                 current_branch = git.get_current_branch()
                 machete_client.create_github_pr(current_branch, draft=cli_opts.opt_draft)
             elif param == "retarget-pr":
                 current_branch = git.get_current_branch()
                 machete_client.expect_in_managed_branches(current_branch)
                 machete_client.retarget_github_pr(current_branch)
+            elif param == "checkout-pr":
+                if len(list_args) == 1:
+                    raise MacheteException(
+                        "Argument to `git machete github checkout-pr` cannot be empty; expected PR number.")
+                try:
+                    pr_no: int = int(list_args[1])
+                except ValueError:
+                    raise MacheteException("PR number is not integer value!")
+                machete_client.checkout_github_pr(pr_no)
             else:
                 raise MacheteException(
                     f"`github` requires a subcommand: one of `{github_allowed_subcommands}`")
@@ -409,8 +423,8 @@ def launch(orig_args: List[str]) -> None:
             elif len(list_args) > 2:
                 raise MacheteException(f"Too many arguments to `git machete list {list_args[0]}` ")
             elif (list_args[0] in (
-                  "addable", "managed", "slidable", "unmanaged", "with-overridden-fork-point") and
-                  len(list_args) > 1):
+                 "addable", "managed", "slidable", "unmanaged", "with-overridden-fork-point") and\
+                 len(list_args) > 1):
                 raise MacheteException(
                     f"`git machete list {list_args[0]}` does not expect extra arguments")
             elif list_args[0] == "slidable-after" and len(list_args) != 2:
@@ -495,8 +509,8 @@ def launch(orig_args: List[str]) -> None:
             machete_client.status(warn_on_yellow_edges=True)
         elif cmd == "traverse":
             traverse_long_opts = ["fetch", "list-commits", "merge",
-                                  "no-detect-squash-merges", "no-edit-merge",
-                                  "no-interactive-rebase",
+                                  "no-detect-squash-merges",
+                                  "no-edit-merge", "no-interactive-rebase",
                                   "no-push", "no-push-untracked", "push", "push-untracked",
                                   "return-to=", "start-from=", "whole", "yes"]
             expect_no_param(parse_options(args, "FlMnWwy", traverse_long_opts))
--- a/git_machete/cli.py
+++ b/git_machete/cli.py
@@ -365,8 +365,8 @@ def launch(orig_args: List[str]) -> None:
             if dest != current_branch:
                 git.checkout(dest)
         elif cmd == "github":
-            github_allowed_subcommands = "anno-prs|checkout-pr|create-pr|retarget-pr"
-            list_args = parse_options(args) if args[0] == 'checkout-pr' \
+            github_allowed_subcommands = "anno-prs|checkout-prs|create-pr|retarget-pr"
+            list_args = parse_options(args) if args[0] == 'checkout-prs' \
                 else check_required_param(parse_options(args, "", ["draft"]), github_allowed_subcommands)
             if not list_args:
                 raise MacheteException(f"`git machete github` expects argument(s): {github_allowed_subcommands}")
@@ -375,15 +375,15 @@ def launch(orig_args: List[str]) -> None:
             machete_client.read_definition_file()
             if param == "anno-prs":
                 machete_client.sync_annotations_to_github_prs()
-            elif param == "checkout-pr":
+            elif param == "checkout-prs":
                 if len(list_args) == 1:
                     raise MacheteException(
-                        "Argument to `git machete github checkout-pr` cannot be empty; expected PR number.")
+                        "Argument to `git machete github checkout-prs` cannot be empty; expected PR number.")
                 try:
                     pr_no: int = int(list_args[1])
                 except ValueError:
                     raise MacheteException("PR number is not integer value!")
-                machete_client.checkout_github_pr(pr_no)
+                machete_client.checkout_github_prs(pr_no)
             elif param == "create-pr":
                 check_required_param(parse_options(args, "", ["draft"]), github_allowed_subcommands)
                 current_branch = git.get_current_branch()
--- a/git_machete/client.py
+++ b/git_machete/client.py
@@ -15,9 +15,9 @@ from git_machete.constants import (
 from git_machete.exceptions import MacheteException, StopTraversal
 from git_machete.git_operations import GitContext
 from git_machete.github import (
-    GitHubPullRequest, derive_pull_request_by_head, set_base_of_pull_request,
-    get_parsed_github_remote_url, is_github_remote_url, create_pull_request,
-    derive_current_user_login, set_milestone_of_pull_request,
+    GitHubPullRequest, check_pr_already_created_by_number, derive_pull_requests,
+    derive_pull_request_by_head, set_base_of_pull_request, get_parsed_github_remote_url, is_github_remote_url,
+    create_pull_request, derive_current_user_login, set_milestone_of_pull_request,
     add_assignees_to_pull_request, add_reviewers_to_pull_request)
 from git_machete.utils import (
     get_pretty_choices, flat_map, excluding, fmt, tupled, warn, debug, bold,
@@ -1655,6 +1655,56 @@ class MacheteClient:
     def flush_caches(self) -> None:
         self.__branch_defs_by_sha_in_reflog = None
         self.__git.flush_caches()
+
+    def checkout_github_pr(self, pr_no: int) -> None:
+        org: str
+        repo: str
+        remote: str
+        remote, (org, repo) = self.__derive_remote_and_github_org_and_repo()
+        debug('checkout_github_pr', f'organization is {org}, repository is {repo}')
+
+        all_prs: List[GitHubPullRequest] = derive_pull_requests(org, repo)
+        pr: GitHubPullRequest = check_pr_already_created_by_number(pr_no, all_prs)
+
+        if not pr:
+            raise MacheteException(f"PR number {pr_no} is not found in repository {repo}")
+        debug('checkout_github_pr', f'found {pr}')
+
+        self.__git.fetch_remote(remote)
+
+        path: List[str] = self.get_path(pr.head)
+        reversed_path: List[str] = path[::-1]  # need to add from root do down
+        for index, branch in enumerate(reversed_path):
+            if branch not in self.managed_branches:
+                if index == 0:
+                    self.__cli_opts.opt_as_root = True
+                    self.add(branch)
+                    self.__cli_opts.opt_as_root = False
+                    continue
+                self.__cli_opts.opt_onto = reversed_path[index - 1]
+                self.add(branch)
+
+        self.sync_annotations_to_github_prs()
+
+        self.__git.checkout(pr.head)
+        print(fmt(f"Pull request #{pr.number} checked out at local branch {pr.head}"))
+
+    def get_path(self, last_branch: str) -> List[str]:
+        def get_parent_branch_from_pr(branch: str) -> Optional[str]:
+            for pr in all_prs:
+                if pr.head == branch:
+                    return pr.base
+            return None
+
+        path: List[str] = []
+        current_branch: str = last_branch
+        remote, (org, repo) = self.__derive_remote_and_github_org_and_repo()
+        all_prs: List[GitHubPullRequest] = derive_pull_requests(org, repo)
+        while current_branch:
+            path.append(current_branch)
+            current_branch = get_parent_branch_from_pr(current_branch)
+
+        return path
 
     def retarget_github_pr(self, head: str) -> None:
         org: str
--- a/git_machete/client.py
+++ b/git_machete/client.py
@@ -1291,8 +1291,8 @@ class MacheteClient:
         all_prs: List[GitHubPullRequest] = derive_pull_requests(org, repo)
         self.__sync_annotations_to_definition_file(all_prs, current_user)
 
-    def __sync_annotations_to_definition_file(self, pr_base: List[GitHubPullRequest], current_user: Optional[str] = None) -> None:
-        for pr in pr_base:
+    def __sync_annotations_to_definition_file(self, prs: List[GitHubPullRequest], current_user: Optional[str] = None) -> None:
+        for pr in prs:
             if pr.head in self.managed_branches:
                 debug('sync_annotations_to_definition_file()',
                       f'{pr} corresponds to a managed branch')
@@ -1659,7 +1659,7 @@ class MacheteClient:
         self.__branch_defs_by_sha_in_reflog = None
         self.__git.flush_caches()
 
-    def checkout_github_pr(self, pr_no: int) -> None:
+    def checkout_github_prs(self, pr_no: int) -> None:
         org: str
         repo: str
         remote: str
@@ -1670,7 +1670,7 @@ class MacheteClient:
         pr: GitHubPullRequest = find_pr_by_number(pr_no, all_prs)
 
         if not pr:
-            raise MacheteException(f"PR number {pr_no} is not found in repository {repo}")
+            raise MacheteException(f"PR #{pr_no} is not found in repository {org}/{repo}")
         debug('checkout_github_pr()', f'found {pr}')
 
         print(f"Fetching {remote}...")
@@ -1679,7 +1679,7 @@ class MacheteClient:
             self.flush_caches()
 
         self.__cli_opts.opt_yes = True  # TODO (#161): pass only needed options to methods
-        path: List[str] = self.__get_path_from_pr_chain(pr.head, all_prs)
+        path: List[str] = self.__get_path_from_pr_chain(pr, all_prs)
         reversed_path: List[str] = path[::-1]  # need to add from root downwards
         for index, branch in enumerate(reversed_path):
             if branch not in self.managed_branches:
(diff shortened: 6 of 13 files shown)


git_machete/cmd.py
```diff
@@ -1515,6 +1516,84 @@ def flush_caches(self) -> None:
...
+        all_prs: List[GitHubPullRequest] = derive_pull_requests(org, repo)
+        pr: GitHubPullRequest = check_pr_already_created_by_number(pr_no, all_prs)
+
+        if not pr:
+            raise MacheteException(f"PR number {pr_no} is not found in repository {repo}")
+        debug('checkout_github_pr', f'found {pr}')
+
+        current_branch: Optional[str] = self.__git.get_currently_checked_out_branch_or_none()
+        if current_branch == 'PR_' + str(pr.number):
+            self.__git.checkout(pr.head)
```
rifel123: Situation when user already checked out given PR, but in the mean time someone added commit into this pull request. This way committed changes are also reflected locally.

git_machete/cmd.py
```diff
@@ -1515,6 +1516,84 @@ def flush_caches(self) -> None:
...
+                    break
+
+        if upstream_prs:
+            reversed_path = path[::-1]  # need to add from root do down
+            for index, branch in enumerate(reversed_path):
+                if branch not in self.managed_branches:
+                    self.__cli_opts.opt_onto = reversed_path[index - 1]
+                    self.add(branch)
+
+            self.sync_annotations_to_github_prs()
```
rifel123: custom annotations will be lost here ;/

git_machete/cmd.py
```diff
@@ -1515,6 +1516,84 @@ def flush_caches(self) -> None:
...
+                checked[commit_timestamp] = b
+
+            parent: str = checked[max(checked.keys())]
+            branch_first_commit_time: int = int(self.__git.get_first_reflog_entry_timestamp(remote, branch))
+
+            while int(self.__git.get_first_reflog_entry_timestamp(remote, parent)) > branch_first_commit_time:
+                del checked[max(checked.keys())]
+                parent = checked[max(checked.keys())]
+            return parent  # the youngest common commit found is considered to be from a parent branch
+            # but there is case unhadled when 2 different branches has the youngest common commit (which branch to take?)
```
rifel123: When parent branch has another child branch and that child branch doesn't have any commits

git_machete/cmd.py
```diff
@@ -1515,6 +1516,84 @@ def flush_caches(self) -> None:
...
+
+            for b in base:
+                common_commit_hash: str = self.__git.popen_git("merge-base", '/'.join([remote, b]), '/'.join([remote, branch])).rstrip()
+                commit_timestamp: int = int(self.__git.get_commit_timestamp(common_commit_hash))
+                checked[commit_timestamp] = b
+
+            parent: str = checked[max(checked.keys())]
+            branch_first_commit_time: int = int(self.__git.get_first_reflog_entry_timestamp(remote, branch))
+
+            while int(self.__git.get_first_reflog_entry_timestamp(remote, parent)) > branch_first_commit_time:
```
rifel123: assumption: parent branch has to be older than child.

git_machete/cmd.py
```diff
@@ -1515,6 +1516,84 @@ def flush_caches(self) -> None:
...
+                parent = checked[max(checked.keys())]
+            return parent  # the youngest common commit found is considered to be from a parent branch
+            # but there is case unhadled when 2 different branches has the youngest common commit (which branch to take?)
+
+        path: List[str] = []
+        current_branch: str = last_branch
+
+        while current_branch not in self.__roots:
+            path.append(current_branch)
+            current_branch = get_parent_remote_branch(current_branch)
```
rifel123: hmm, is it possible that we never hit root branch here?

git_machete/cmd.py
```diff
@@ -1515,6 +1516,84 @@ def flush_caches(self) -> None:
...
+
+        if not pr:
+            raise MacheteException(f"PR number {pr_no} is not found in repository {repo}")
+        debug('checkout_github_pr', f'found {pr}')
+
+        current_branch: Optional[str] = self.__git.get_currently_checked_out_branch_or_none()
+        if current_branch == 'PR_' + str(pr.number):
+            self.__git.checkout(pr.head)
+
+        self.__git.fetch_pull_request(remote, pr.number)
```
rifel123: maybe it would be better to checkout just head branch of given PR instead of checking out PR which is read only.

git_machete/cmd.py
rifel123: whole this function would not be necessary if `checkout-pr` checks only if parent brach has opened PR and so on.

completion/git-machete.completion.zsh
```diff
@@ -64,12 +64,12 @@ _git-machete() {
...
           ;;
-        (github)
-          _arguments '1:: :__git_machete_github_subcommands' && ret=0
-          ;;
         (help)
           _arguments '1:: :__git_machete_help_topics' && ret=0
           ;;
+        (github)
+          _arguments '1:: :__git_machete_github_subcommands' && ret=0
+          ;;
```
BartoszSiwinski: Why  did you move it? It looks like the intent was to keep it in alphabetic order (it's a bit debatable whether it was or not, because 'github' is after 'g' but before 'go').

completion/git-machete.completion.bash
```diff
@@ -21,6 +21,7 @@ _git_machete() {
   local discover_opts="-C --checked-out-since= -l --list-commits -r --roots= -y --yes"
   local fork_point_opts="--inferred --override-to= --override-to-inferred --override-to-parent --unset-override"
   local github_create_pr_opts="--draft"
+  local github_checkout_pr_opts="PR_number"
```
PawelLipski: Pls update `github_subcommands` as well! coz as for not `git machete github ch<tab>` doesn't autocomplete as expected

PawelLipski: Also, wrt. `PR_number`... what's the purpose of this?... AFAICS it'll complete `git machete github checkout-pr P<tab>`... but this doesn't seem very useful (nor an established convention in CLIs) ;)

PawelLipski: TBH I'd leave an empty `COMPREPLY=('')` after `checkout-pr`... we could fetch the open PR numbers from github but that's way too much engineering FWIW

completion/git-machete.completion.zsh
PawelLipski: Whoops but now it's after `help` 😅

completion/git-machete.completion.zsh
```diff
@@ -217,6 +217,7 @@ __git_machete_github_subcommands() {
   local github_subcommands
   github_subcommands=(
     'anno-prs:annotate the branches based on their corresponding GitHub PR numbers and authors'
+    'checkout-pr:checks out locally given pull request'
```
PawelLipski: Dunno why, but adverb at the end feels more natural:
```suggestion
    'checkout-pr:check out the given pull request locally'
```

PawelLipski: Pls use imperative rather than 3rd person consistently

README.md
```diff
@@ -156,6 +156,11 @@ Annotate the branches with GitHub PR numbers: <br/>
 git machete github anno-prs
 ```
 
+Checks out the PR into local branch, if pull request is found on any parent branch, it will add every parent branch up to root branch as seen by git-machete: <br/>
```
PawelLipski: Hmmm `any parent branch` suggests there might be a few different parent branches for the given PR 🤔

PawelLipski: Pls clarify the sentence

PawelLipski:
Whoops I've just tested locally on a sample PR chain that used to look like that:

image

I removed the 3 `chore`/`refactor` branches from .git/machete and executed `git machete github checkout-pr 1217` to restore them... unfortunately:
1. branch `PR_1217` has been checked out, I'd rather expect the corresponding branch `chore/GMV3-1494-jitpack-to-sonatype` to be checked out, this is more practical
2. no branch has been added to .git/machete :/// and I'd rather expect the 3-PR chain to be re-added
3. nit: pls rebase onto `develop` so that the fix from #201 is included

Pls take a look!

git_machete/cmd.py
```diff
@@ -1515,6 +1516,84 @@ def flush_caches(self) -> None:
...
+                if branch not in self.managed_branches:
+                    self.__cli_opts.opt_onto = reversed_path[index - 1]
+                    self.add(branch)
+
+            self.sync_annotations_to_github_prs()
+
+        self.__git.checkout(f'PR_{pr.number}')
+        print(fmt(f"Pull request #{pr.number} checked out at local branch PR_{pr.number}"))
+
+    def get_path(self, remote: str, last_branch: str) -> List[str]:
```
PawelLipski: Hmmm I didn't exactly investigate the logic here... but as discussed on a call yesterday, I wouldn't expect checking commit timestamps/reflogs etc. in finding out how the path up to an already machete-tracked branch should look like 🤔

PawelLipski: I'd rather completely rely on Github PR base branches for that purpose! See my PR comment about restoring the 3 branches based on their Github PRs

git_machete/docs.py
```diff
@@ -329,6 +329,12 @@
           Any existing annotations are overwritten for the branches that have an opened PR; annotations for the other branches remain untouched.
           Equivalent to `git machete anno --sync-github-prs`.
 
+        <b>`checkout-pr` [PR_number]`:</b>
+
+          Checks out given pull request (by number) locally under <PR_number> branch. Note this branch is read only, so there is no possibility to push changes there.
```
PawelLipski: Yeah, as commented on PR, that's the problem with `refs/pull/<PR>/head` branches... TBH they're of very limited usefulness :/ I'd rather check them out as regular local branches instead

git_machete/cmd.py
PawelLipski: Also, pls note that there is already `git machete discover` that tries to infer the parent for each branch based on things like reflog/latest checkout times etc.... although after dunno-how-many rounds of improvements it still ~sucks~ is far from perfection.

Actually, I've conceived `git machete github checkout-pr` as a more reliable `discover` alternative of sorts (at least for the case when there're already PRs in Github for the branches in question)

git_machete/cmd.py
```diff
@@ -2069,18 +2148,32 @@ def check_required_param(in_args: List[str], allowed_values_string: str) -> str:
...
                 machete_client.sync_annotations_to_github_prs()
             elif param == "create-pr":
+                check_required_param(parse_options(args, "", ["draft"]), github_allowed_subcommands)
                 cb = git.get_current_branch()
                 machete_client.create_github_pr(cb, draft=cli_opts.opt_draft)
             elif param == "retarget-pr":
                 cb = git.get_current_branch()
                 machete_client.expect_in_managed_branches(cb)
                 machete_client.retarget_github_pr(cb)
+            elif param == "checkout-pr":
```
PawelLipski: Pls keep cases in abc order

git_machete/cmd.py
```diff
@@ -2069,18 +2148,32 @@ def check_required_param(in_args: List[str], allowed_values_string: str) -> str:
             if dest != cb:
                 git.checkout(dest)
         elif cmd == "github":
-            github_allowed_subcommands = "anno-prs|create-pr|retarget-pr"
-            param = check_required_param(parse_options(args, "", ["draft"]), github_allowed_subcommands)
+            github_allowed_subcommands = "anno-prs|create-pr|retarget-pr|checkout-pr"
```
PawelLipski: Abc order

git_machete/cmd.py
```diff
@@ -2069,18 +2148,32 @@ def check_required_param(in_args: List[str], allowed_values_string: str) -> str:
...
+            list_args = parse_options(args) if args[0] == 'checkout-pr' else check_required_param(parse_options(args, "", ["draft"]), github_allowed_subcommands)
```
PawelLipski: Whoops the logic isn't right here...

```
$ git machete github create-pr 

github requires a subcommand: one of anno-prs|create-pr|retarget-pr|checkout-pr
```

I'd strongly suggest providing tests for the existing logic first (before this PR is further developed) to avoid painful regressions :/

git_machete/cmd.py
```diff
@@ -266,6 +266,7 @@ def add(self, b: str) -> None:
             else:
                 self.__down_branches[onto] = [b]
             print(fmt(f"Added branch `{b}` onto `{onto}`"))
+            self.managed_branches.append(b)
```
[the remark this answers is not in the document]
rifel123: hmm, when I added re-reading of definition_file inside `save_definition_file()`, then all branches in definition_file was dubled. Some further investigation is needed. I'll raise a ticket for that. As for now, let's leave adding branches into `self.managed_branches`.

git_machete/cmd.py
rifel123: it works nice, until reflogs are available :/ which is rather rare. Right now search is done by chain if pull requests upwards, so the code is not that complicated anymore.

rifel123:
About known bug described at the very top: i see three different scenarios here:
1) do not ask user if he wants to add branches or not, just add them (kind of brutal)
2) if answer to add is `n`, try to add child branches with asking (but then first branch after `n` will be a root branch)
3) if answer to add is `n`, add only head pr branch as root, if it's not managed already

I am looking for your votes/suggestions :)

PawelLipski:
I took the liberty to change the PR base to PR #214 (`git machete github retarget-pr` FTW ;) )

README.md
```diff
@@ -156,6 +156,11 @@ Annotate the branches with GitHub PR numbers: <br/>
 git machete github anno-prs
 ```
 
+Check out the PR into local branch, also traverse chain of pull requests upwards, adding branches one by one to git-machete and check them out locally: <br/>
```
PawelLipski, line 159: Pls update RELEASE_NOTES.md as well

README.md
```diff
@@ -156,6 +156,11 @@ Annotate the branches with GitHub PR numbers: <br/>
...
+```shell script
+git machete github checkout-pr [PR_number]
```
PawelLipski: `[...]` typically denote **optional** params in command-line docs, see `man grep` for a sample

PawelLipski: So:
```suggestion
git machete github checkout-pr <PR-number>
```

git_machete/client.py
```diff
@@ -1656,6 +1657,56 @@ def flush_caches(self) -> None:
...
+                    self.add(branch)
+                    self.__cli_opts.opt_as_root = False
+                    continue
+                self.__cli_opts.opt_onto = reversed_path[index - 1]
+                self.add(branch)
+
+        self.sync_annotations_to_github_prs()
+
+        self.__git.checkout(pr.head)
+        print(fmt(f"Pull request #{pr.number} checked out at local branch {pr.head}"))
```
PawelLipski: Nit, for consistency with how most messages are formatted:
```suggestion
        print(fmt(f"Pull request `#{pr.number}` checked out at local branch `{pr.head}`"))
```

git_machete/client.py
```diff
@@ -1656,6 +1657,56 @@ def flush_caches(self) -> None:
...
+        def get_parent_branch_from_pr(branch: str) -> Optional[str]:
+            for pr in all_prs:
+                if pr.head == branch:
+                    return pr.base
+            return None
+
+        path: List[str] = []
+        current_branch: str = last_branch
+        remote, (org, repo) = self.__derive_remote_and_github_org_and_repo()
+        all_prs: List[GitHubPullRequest] = derive_pull_requests(org, repo)
```
PawelLipski: Hmmm where's `all_prs` used?

git_machete/docs.py
```diff
@@ -314,6 +314,12 @@
           Any existing annotations are overwritten for the branches that have an opened PR; annotations for the other branches remain untouched.
           Equivalent to `git machete anno --sync-github-prs`.
 
+        <b>`checkout-pr` [PR_number]`:</b>
```
```suggestion
        <b>`checkout-pr <PR-number>`:</b>
```

PawelLipski: Pls update the .rst docs under `docs/` as well (we'll need to keep them in sync until #211 is done :/)

git_machete/git_operations.py
```diff
@@ -705,3 +705,18 @@ def get_latest_checkout_timestamps(self) -> Dict[str, int]:  # TODO (#110): defa
...
+        branches = self.get_remote_branches()
+        return [strip_remote(b, remote) for b in branches if 'HEAD' not in b]
+
+    def get_commit_timestamp(self, commit_hash: str) -> str:
+        return self.popen_git('show', '-s', '--format=%ct', commit_hash).rstrip()
+
+    def get_first_reflog_entry_timestamp(self, remote: str, branch: str) -> str:
+        output: List[str] = self.popen_git('reflog', '--format=%ct', "/".join([remote, branch])).split('\n')
+        output.remove('')
+        return output[-1].strip()  # take only last value
```
PawelLipski: These methods are no longer used anywhere

git_machete/cli.py
```diff
@@ -365,13 +365,27 @@ def check_required_param(in_args: List[str], allowed_values_string: str) -> str:
...
+                raise MacheteException(f"`git machete github` expects argument(s): {github_allowed_subcommands}")
+
+            param = args[0]
             machete_client.read_definition_file()
             if param == "anno-prs":
                 machete_client.sync_annotations_to_github_prs()
+            elif param == "checkout-pr":
+                if len(list_args) == 1:
+                    raise MacheteException(
+                        "Argument to `git machete github checkout-pr` cannot be empty; expected PR number.")
```
```
$ git machete github checkout-pr 204

PR number 204 is not found in repository git-machete
```

PR #204 exists but is already closed. Pls open an issue to make this distinction in the error message (not to be done on this PR)

PawelLipski:
> when checkout-pr detects opened PR on parent branches, it will try to add the chain of branches into definition file and also checkout them locally. then user is asked if he wants to add those branches one by one, and if he answers no in the middle the

Hmmm wait:

```
$ git machete github checkout-pr 205
Added branch testing/github_methods onto develop
Added branch enchance/checkout-pr onto testing/github_methods
Annotating testing/github_methods as PR #214 (rifel123)
Annotating enchance/checkout-pr as PR #205 (rifel123)
Pull request #205 checked out at local branch enchance/checkout-pr
```

`checkout-pr` didn't ask me about anything here 🤔

git_machete/client.py
rifel123: in nested function `get_parent_branch_from_pr` at line 1695

rifel123:
@PawelLipski `checkout-pr` will not ask if branch is already checked out locally, in this case only definition file will be updated. But if you delete those local branches and then say NO when asked, it will break flow on next branch ;/

PawelLipski:
> if you delete those local branches and then say NO when asked, it will break flow on next branch ;/

OK, if the local branches corresponding to the PRs are missing, then let's just force-checkout them locally, even without asking the user &mdash; that's what I'd expect `checkout-pr` to do

docs/source/cli_help/github.rst
```diff
@@ -25,6 +25,16 @@ a GitHub API token with ``repo`` scope is required, see https://github.com/setti
   Any existing annotations are overwritten for the branches that have an opened PR; annotations for the other branches remain untouched.
   Equivalent to ``git machete anno --sync-github-prs``.
 
+``checkout-pr``:
+
+  Check out given pull request (by number) locally under head branch name of pull request,
```
PawelLipski:
```suggestion
  Check out the head branch of the given pull request (specified by number),
```

docs/source/cli_help/github.rst
```diff
@@ -25,6 +25,16 @@ a GitHub API token with ``repo`` scope is required, see https://github.com/setti
...
+  also traverse chain of pull requests upwards, adding branches one by one to git-machete and check them out locally.
+  Once pull request is checked-out locally, modify definition file according to structure found and synchronize pull requests annotations.
```
PawelLipski:
```suggestion
  Once pull request is checked out locally, add new branches to the definition file according to the PR structure and annotate them with pull request numbers.
```

git_machete/client.py
```diff
@@ -1656,6 +1657,57 @@ def flush_caches(self) -> None:
...
+
+        self.__git.fetch_remote(remote)
+
+        self.__cli_opts.opt_yes = True
+        path: List[str] = self.get_path(pr.head)
+        reversed_path: List[str] = path[::-1]  # need to add from root do down
+        for index, branch in enumerate(reversed_path):
+            if branch not in self.managed_branches:
+                if index == 0:
+                    self.__cli_opts.opt_as_root = True
```
PawelLipski, line 1687: Ouch... pls add a TODO for issue #161 here so that we don't forget (no other action required on this PR)

git_machete/client.py
```diff
@@ -1656,6 +1657,57 @@ def flush_caches(self) -> None:
...
+                    continue
+                self.__cli_opts.opt_onto = reversed_path[index - 1]
+                self.add(branch)
+        self.__cli_opts.opt_yes = False
+        self.sync_annotations_to_github_prs()
+
+        self.__git.checkout(pr.head)
+        print(fmt(f"Pull request `#{pr.number}` checked out at local branch `{pr.head}`"))
+
+    def get_path(self, last_branch: str) -> List[str]:
```
PawelLipski: Whoops... pls make the method name more specific... also, shouldn't it be private?

git_machete/client.py
PawelLipski: Whoops okay, that's pretty brittle/surprising (that the only reference to a local var is **before** its declaration)... I think the `get_parent_branch_from_pr` method can be replaced with a one-liner using `next(filter(pred, iterable), None)`

git_machete/docs.py
```diff
@@ -314,6 +314,12 @@
           Any existing annotations are overwritten for the branches that have an opened PR; annotations for the other branches remain untouched.
           Equivalent to `git machete anno --sync-github-prs`.
 
+        <b>`checkout-pr <PR-number>`:</b>
+
+          Check out given pull request (by number) locally under head branch name of pull request,
```
PawelLipski: Update to keep in sync with .rst

git_machete/github.py
```diff
@@ -161,7 +168,7 @@ def create_pull_request(org: str, repo: str, head: str, base: str, title: str, d
     }
     prs: List[GitHubPullRequest] = derive_pull_requests(org, repo)
     to_load: GitHubPullRequest = GitHubPullRequest(1, 'user', base, head, '')
```
PawelLipski: Wow I've only seen that now &mdash; that's pretty unclean, to create a fake `GitHubPullRequest` object just so that it matches the interface of `check_pr_already_created`

git_machete/github.py
```diff
@@ -161,7 +168,7 @@ def create_pull_request(org: str, repo: str, head: str, base: str, title: str, d
...
-    pr_found: Optional[GitHubPullRequest] = __check_pr_already_created(to_load, prs)
+    pr_found: Optional[GitHubPullRequest] = check_pr_already_created(to_load, prs)
```
PawelLipski: I'd rather change this method to sth like `check_pr_already_created_by_head_and_base`

PawelLipski: Also, both `check_pr_already_created_...` methods could in fact be `find_pr_...`

git_machete/tests/functional/test_machete.py
```diff
@@ -1645,6 +1644,7 @@ def test_anno_prs(self) -> None:
     git_api_state_for_test_create_pr = MockGithubAPIState([{'head': {'ref': 'ignore-trailing'}, 'user': {'login': 'github_user'}, 'base': {'ref': 'hotfix/add-trigger'}, 'number': '3', 'html_url': 'www.github.com'}])
 
     @mock.patch('git_machete.options.CommandLineOptions', FakeCommandLineOptions)
+    @mock.patch('git_machete.github.GITHUB_REMOTE_PATTERNS', FAKE_GITHUB_REMOTE_PATTERNS)
```
PawelLipski: Hmmm why is mocking `GITHUB_REMOTE_PATTERNS` necessary? 🤔

git_machete/client.py
```diff
@@ -1656,6 +1657,57 @@ def flush_caches(self) -> None:
...
+
+        if not pr:
+            raise MacheteException(f"PR number {pr_no} is not found in repository {repo}")
+        debug('checkout_github_pr', f'found {pr}')
+
+        self.__git.fetch_remote(remote)
+
+        self.__cli_opts.opt_yes = True
+        path: List[str] = self.get_path(pr.head)
+        reversed_path: List[str] = path[::-1]  # need to add from root do down
```
```suggestion
        reversed_path: List[str] = path[::-1]  # need to add from root downwards
```

git_machete/client.py
```diff
@@ -1656,6 +1657,57 @@ def flush_caches(self) -> None:
...
+            if branch not in self.managed_branches:
+                if index == 0:
+                    self.__cli_opts.opt_as_root = True
+                    self.add(branch)
+                    self.__cli_opts.opt_as_root = False
+                    continue
+                self.__cli_opts.opt_onto = reversed_path[index - 1]
+                self.add(branch)
+        self.__cli_opts.opt_yes = False
+        self.sync_annotations_to_github_prs()
```
PawelLipski: Actually.. this requires another series of Github API calls... while in fact you could probably annotate the branches while you traverse `reversed_path`

git_machete/client.py
```diff
@@ -1656,6 +1657,57 @@ def flush_caches(self) -> None:
...
+        debug('checkout_github_pr', f'organization is {org}, repository is {repo}')
+
+        all_prs: List[GitHubPullRequest] = derive_pull_requests(org, repo)
+        pr: GitHubPullRequest = check_pr_already_created_by_number(pr_no, all_prs)
+
+        if not pr:
+            raise MacheteException(f"PR number {pr_no} is not found in repository {repo}")
+        debug('checkout_github_pr', f'found {pr}')
+
+        self.__git.fetch_remote(remote)
```
PawelLipski, line 1677: I'd print some `Fetching <remote-name>...` to stdout to reduce user confusion (_while nothing shows up right after I've run `checkout-pr`?)

git_machete/client.py
```diff
@@ -1656,6 +1657,57 @@ def flush_caches(self) -> None:
...
+        self.__cli_opts.opt_yes = True
+        path: List[str] = self.get_path(pr.head)
+        reversed_path: List[str] = path[::-1]  # need to add from root do down
+        for index, branch in enumerate(reversed_path):
+            if branch not in self.managed_branches:
+                if index == 0:
+                    self.__cli_opts.opt_as_root = True
+                    self.add(branch)
+                    self.__cli_opts.opt_as_root = False
+                    continue
```
PawelLipski, line 1685: `if/else` pls instead of `if/continue`

git_machete/client.py
```diff
@@ -1656,6 +1657,57 @@ def flush_caches(self) -> None:
...
         self.__git.flush_caches()
 
+    def checkout_github_pr(self, pr_no: int) -> None:
+        org: str
+        repo: str
+        remote: str
+        remote, (org, repo) = self.__derive_remote_and_github_org_and_repo()
+        debug('checkout_github_pr', f'organization is {org}, repository is {repo}')
+
+        all_prs: List[GitHubPullRequest] = derive_pull_requests(org, repo)
```
PawelLipski, line 1669: `GET /repos/.../.../pulls` is invoked 3 times in total, not good ☹️ this causes the entire operation to take like 5-10 seconds :/

That's what I've got from calling `git machete github checkout-pr 229 --debug`:

```
checkout_github_pr: organization is VirtusLab, repository is git-machete
__fire_github_api_request(GET, /repos/VirtusLab/git-machete/pulls, ..., ...): firing a GET request to https://api.github.com/repos/VirtusLab/git-machete/pulls with a bearer token and request body <none>
checkout_github_pr: found PR #229 by BartoszSiwinski: improve/parser -> develop
>>> git fetch origin
>>> git config --get remote.asford.url
<stdout>:
git@github.com:asford/git-machete.git

>>> git config --get remote.origin.url
<stdout>:
git@github.com:VirtusLab/git-machete.git

__fire_github_api_request(GET, /repos/VirtusLab/git-machete/pulls, ..., ...): firing a GET request to https://api.github.com/repos/VirtusLab/git-machete/pulls with a bearer token and request body <none>
>>> git config --get remote.asford.url
<stdout>:
git@github.com:asford/git-machete.git

>>> git config --get remote.origin.url
<stdout>:
git@github.com:VirtusLab/git-machete.git

__fire_github_api_request(GET, /user, ..., ...): firing a GET request to https://api.github.com/user with a bearer token and request body <none>
sync_annotations_to_github_prs(): Current GitHub user is PawelLipski
__fire_github_api_request(GET, /repos/VirtusLab/git-machete/pulls, ..., ...): firing a GET request to https://api.github.com/repos/VirtusLab/git-machete/pulls with a bearer token and request body <none>
sync_annotations_to_github_prs(): PR #230 by rifel123: bugfix/create-pr -> enchance/checkout-pr does NOT correspond to a managed branch
sync_annotations_to_github_prs(): PR #229 by BartoszSiwinski: improve/parser -> develop corresponds to a managed branch
sync_annotations_to_github_prs(): PR #214 by rifel123: testing/github_methods -> develop corresponds to a managed branch
sync_annotations_to_github_prs(): PR #205 by rifel123: enchance/checkout-pr -> testing/github_methods corresponds to a managed branch
>>> git checkout --quiet improve/parser --
Pull request #229 checked out at local branch improve/parser
```

git_machete/client.py
```diff
@@ -1656,6 +1657,57 @@ def flush_caches(self) -> None:
...
+        path: List[str] = self.get_path(pr.head)
+        reversed_path: List[str] = path[::-1]  # need to add from root do down
+        for index, branch in enumerate(reversed_path):
+            if branch not in self.managed_branches:
+                if index == 0:
+                    self.__cli_opts.opt_as_root = True
+                    self.add(branch)
+                    self.__cli_opts.opt_as_root = False
+                    continue
+                self.__cli_opts.opt_onto = reversed_path[index - 1]
```
PawelLipski: Pls replace `if/continue` with `if/else`, that's way easier to reason about

git_machete/cli.py
```diff
@@ -365,13 +365,27 @@ def check_required_param(in_args: List[str], allowed_values_string: str) -> str:
...
+            list_args = parse_options(args) if args[0] == 'checkout-pr' \
+                else check_required_param(parse_options(args, "", ["draft"]), github_allowed_subcommands)
+            if not list_args:
+                raise MacheteException(f"`git machete github` expects argument(s): {github_allowed_subcommands}")
+
+            param = args[0]
             machete_client.read_definition_file()
             if param == "anno-prs":
                 machete_client.sync_annotations_to_github_prs()
+            elif param == "checkout-pr":
```
PawelLipski: Shower thought: https://github.com/VirtusLab/git-machete/issues/231

git_machete/tests/functional/test_machete.py
rifel123: This is to bypass error: `Remotes are defined for this repository, but none of them corresponds to GitHub`. I could simply add another remote which corresponds to Github but then in `test_checkout_pr` i would never reach correct repository name. To sum up, this is only to show up correct repository name in error message while testing wrong pull request number.

git_machete/client.py
rifel123: Screenshot 2021-09-14 at 15 33 16
API calls were limited to one per endpoint, but still it took 2,84 seconds on my machine while calling from command line. Mainly due to `git fetch remote`, so I am afraid we can't speed up `checkout-pr` much.

PawelLipski: Ok! ~3 seconds is a good time for such an operation.

docs/source/cli_help/github.rst
```diff
@@ -8,7 +8,7 @@ github
 
     git machete github <subcommand>
 
-where ``<subcommand>`` is one of: ``anno-prs``, ``create-pr``, ``retarget-pr``.
+where ``<subcommand>`` is one of: ``anno-prs``, ``checkout-pr``, ``create-pr``, ``retarget-pr``.
```
PawelLipski: Yeah... so... last minute change to be considered... as per #231, it might make more sense to use `checkout-prs` (plural) name for this subcommand 🤔

docs/source/cli_help/github.rst
```diff
@@ -25,6 +25,16 @@ a GitHub API token with ``repo`` scope is required, see https://github.com/setti
   Any existing annotations are overwritten for the branches that have an opened PR; annotations for the other branches remain untouched.
   Equivalent to ``git machete anno --sync-github-prs``.
 
+``checkout-pr``:
+
+  Check out the head branch of the given pull request (specified by number),
+  also traverse chain of pull requests upwards, adding branches one by one to git-machete and check them out locally.
+  Once pull request is checked out locally, add new branches to the definition file according to the PR structure and annotate them with pull request numbers.
```
PawelLipski: The same thing mentioned twice: _adding branches one by one to git-machete_ vs  _adding branches one by one to git-machete and check them out locally_, pls condense

git_machete/client.py
```diff
@@ -1289,9 +1288,13 @@ def sync_annotations_to_github_prs(self) -> None:
         debug('sync_annotations_to_github_prs()',
               'Current GitHub user is ' + (current_user or '<none>'))
         pr: GitHubPullRequest
-        for pr in derive_pull_requests(org, repo):
+        all_prs: List[GitHubPullRequest] = derive_pull_requests(org, repo)
+        self.__sync_annotations_to_definition_file(all_prs, current_user)
+
+    def __sync_annotations_to_definition_file(self, pr_base: List[GitHubPullRequest], current_user: Optional[str] = None) -> None:
```
PawelLipski: `pr_base` or just `prs`?

git_machete/docs.py
```diff
@@ -314,6 +314,15 @@
           Any existing annotations are overwritten for the branches that have an opened PR; annotations for the other branches remain untouched.
           Equivalent to `git machete anno --sync-github-prs`.
 
+        <b>`checkout-pr <PR-number>`:</b>
+
+          Check out the head branch of the given pull request (specified by number),
+          also traverse chain of pull requests upwards, adding branches one by one to git-machete and check them out locally.
```
PawelLipski, line 320: Reflect the change from .rst

git_machete/github.py
```diff
@@ -143,9 +143,16 @@ def __fire_github_api_request(method: str, path: str, token: Optional[str], requ
...
-def __check_pr_already_created(pull: GitHubPullRequest, pull_requests: List[GitHubPullRequest]) -> Optional[GitHubPullRequest]:
+def find_pr_by_base_and_head(base: str, head: str, pull_requests: List[GitHubPullRequest]) -> Optional[GitHubPullRequest]:
     for pr in pull_requests:
-        if pull.base == pr.base and pull.head == pr.head:
+        if base == pr.base and head == pr.head:
+            return pr
+    return None
+
+
+def find_pr_by_number(no: int, pull_requests: List[GitHubPullRequest]) -> Optional[GitHubPullRequest]:
```
PawelLipski: Nit, for consistency:
```suggestion
def find_pr_by_number(number: int, pull_requests: List[GitHubPullRequest]) -> Optional[GitHubPullRequest]:
```

git_machete/tests/functional/test_machete.py
```diff
@@ -1783,3 +1782,177 @@ def test_github_create_pr(self) -> None:
         if e:
             self.assertEqual(e.exception.parameter, expected_error_message,
                              'Verify that expected error message has appeared when given pull request to create is already created.')
+
+    git_api_state_for_test_checkout_pr = MockGithubAPIState([
+        {'head': {'ref': 'chore/redundant_checks'}, 'user': {'login': 'github_user'}, 'base': {'ref': 'restrict_access'}, 'number': '18', 'html_url': 'www.github.com'},
+        {'head': {'ref': 'restrict_access'}, 'user': {'login': 'github_user'}, 'base': {'ref': 'allow-ownership-link'}, 'number': '17', 'html_url': 'www.github.com'},
+        {'head': {'ref': 'allow-ownership-link'}, 'user': {'login': 'github_user'}, 'base': {'ref': 'bugfix/feature'}, 'number': '12', 'html_url': 'www.github.com'},
+        {'head': {'ref': 'bugfix/feature'}, 'user': {'login': 'github_user'}, 'base': {'ref': 'enchance/feature'}, 'number': '6', 'html_url': 'www.github.com'},
+        {'head': {'ref': 'enchance/add_user'}, 'user': {'login': 'github_user'}, 'base': {'ref': 'develop'}, 'number': '19', 'html_url': 'www.github.com'},
```
PawelLipski: Nit: `enhance` ;)

git_machete/tests/functional/test_machete.py
PawelLipski: Hmmm okay two things I can't see here:
1. why specifically in `test_checkout_pr` you'd never reach a correct repository name? would you reach that correct repository name in other test methods?
2. `FAKE_GITHUB_REMOTE_PATTERNS` is now basically supposed to match the URL of `origin`, which corresponds to a local directory, right? how do these URLs look like, just `/tmp/<folder>`?

PawelLipski: (I'd intuitively still favor adding an actual github remote to reduce the amount of mocking... but maybe that's indeed gonna end up more complicated)

git_machete/client.py
```diff
@@ -1656,6 +1659,60 @@ def flush_caches(self) -> None:
...
+        repo: str
+        remote: str
+        remote, (org, repo) = self.__derive_remote_and_github_org_and_repo()
+        debug('checkout_github_pr()', f'organization is {org}, repository is {repo}')
+
+        all_prs: List[GitHubPullRequest] = derive_pull_requests(org, repo)
+        pr: GitHubPullRequest = find_pr_by_number(pr_no, all_prs)
+
+        if not pr:
+            raise MacheteException(f"PR number {pr_no} is not found in repository {repo}")
```
PawelLipski: Nit:
```suggestion
            raise MacheteException(f"PR #{pr_no} is not found in repository `{org}/{repo}`")
```

git_machete/client.py
```diff
@@ -1656,6 +1659,60 @@ def flush_caches(self) -> None:
...
+        reversed_path: List[str] = path[::-1]  # need to add from root downwards
+        for index, branch in enumerate(reversed_path):
+            if branch not in self.managed_branches:
+                if index == 0:
+                    self.__cli_opts.opt_as_root = True
+                    self.add(branch)
+                    self.__cli_opts.opt_as_root = False
+                else:
+                    self.__cli_opts.opt_onto = reversed_path[index - 1]
+                    self.add(branch)
```
PawelLipski, line 1692: Whooops I've seen a bug here: clone repo https://github.com/VirtusLab/akka-serialization-helper/ , add just `main` to .git/machete and then run:

```
git branch -r -D origin/feachure/state-dump
git machete github checkout-pr 95
```

You'll probably see:

```
$ git machete github checkout-pr 95
Fetching origin...
From github.com:VirtusLab/akka-serialization-helper
 * [new branch]      feachure/state-dump -> origin/feachure/state-dump
A local branch feachure/state-dump does not exist. Creating out of main
Switched to a new branch 'feachure/state-dump'
Added branch feachure/state-dump onto main
Annotating feachure/state-dump as PR #95 (MarconZet)
Pull request #95 checked out at local branch feachure/state-dump
```

In fact it's not specific to that particular repo, you can quite easily reproduce it on git-machete or the repos used in our functional tests.

PawelLipski: `feachure/state-dump` will be checked out of **`main`** and NOT of `origin/feachure/state-dump`... that's probably because `origin/feachure/state-dump` only appears once `fetch_remote` is executed.

PawelLipski: `__remote_branches_cached` should be updated after a fetch happens... see how it's handled e.g. in `def traverse`.

PawelLipski: Pls also first a regression test and let it fail, in the spirit of [TDD](https://pl.wikipedia.org/wiki/Test-driven_development) ‼️

git_machete/tests/functional/test_machete.py
```diff
@@ -187,10 +187,10 @@ def make_response_object(status_code: int, response_data: Union[List[Dict[str, A
         return MockGithubAPIResponse(status_code, response_data)
 
     @staticmethod
-    def find_number(url: str) -> Optional[str]:
-        m = re.search(r'\d+', url)
+    def find_number(url: str, entity: str) -> Optional[str]:
+        m = re.search(f'{entity}\/(\d+)', url)
```
PawelLipski: Nit: slash doesn't need to be escaped in regexes

git_machete/tests/functional/test_machete.py
rifel123: yes, this will work on other github methods tests but not in checkout-pr test. Even if I grab exact repository name of `tmp.XXXXXX`, and add new remote with url `https://github.com/user/repo_name.git` it will fail on `fetch` because given url does not exists. `FAKE_GITHUB_REMOTE_PATTERNS` allows to select local (non hosted) repository for fetching.

PawelLipski: Awwww okay I see, it's the **fetch** that's the problem specific to `checkout-pr(s)`... okay!

git_machete/client.py
```diff
@@ -1670,14 +1670,16 @@ def checkout_github_pr(self, pr_no: int) -> None:
         pr: GitHubPullRequest = find_pr_by_number(pr_no, all_prs)
 
         if not pr:
-            raise MacheteException(f"PR number {pr_no} is not found in repository {repo}")
+            raise MacheteException(f"PR #{pr_no} is not found in repository {org}/{repo}")
```
PawelLipski: Nit:
```suggestion
            raise MacheteException(f"PR #{pr_no} is not found in repository `{org}/{repo}`")
```

git_machete/client.py
```diff
@@ -1698,19 +1700,11 @@ def checkout_github_pr(self, pr_no: int) -> None:
...
-        current_branch: str = last_branch
-        while current_branch:
-            path.append(current_branch)
-            current_branch = get_parent_branch_from_pr(current_branch)
-
+    def __get_path_from_pr_chain(self, current_pr: GitHubPullRequest, all_prs: List[GitHubPullRequest]) -> List[str]:
+        path: List[str] = [current_pr.head]
+        while current_pr:
+            path.append(current_pr.base)
+            current_pr = next(filter(lambda x: x.head == current_pr.base, all_prs), None)
```
PawelLipski: I think this can be extracted to `utils` next to the things like `flat_map`, under the name `find_or_none`

git_machete/github.py
```diff
@@ -150,9 +150,9 @@ def find_pr_by_base_and_head(base: str, head: str, pull_requests: List[GitHubPul
     return None
 
 
-def find_pr_by_number(no: int, pull_requests: List[GitHubPullRequest]) -> Optional[GitHubPullRequest]:
+def find_pr_by_number(number: int, pull_requests: List[GitHubPullRequest]) -> Optional[GitHubPullRequest]:
```
PawelLipski: Probably can be replaced with a call to the aforementioned `find_or_none`

git_machete/tests/functional/test_machete.py
```diff
@@ -1811,25 +1815,25 @@ def test_github_create_pr(self) -> None:
...
-        {'head': {'ref': 'enchance/add_user'}, 'user': {'login': 'github_user'}, 'base': {'ref': 'develop'}, 'number': '19', 'html_url': 'www.github.com'},
+        {'head': {'ref': 'bugfix/feature'}, 'user': {'login': 'github_user'}, 'base': {'ref': 'enhance/feature'}, 'number': '6', 'html_url': 'www.github.com'},
+        {'head': {'ref': 'enhance/add_user'}, 'user': {'login': 'github_user'}, 'base': {'ref': 'develop'}, 'number': '19', 'html_url': 'www.github.com'},
         {'head': {'ref': 'testing/add_user'}, 'user': {'login': 'github_user'}, 'base': {'ref': 'bugfix/add_user'}, 'number': '22', 'html_url': 'www.github.com'},
         {'head': {'ref': 'chore/comments'}, 'user': {'login': 'github_user'}, 'base': {'ref': 'testing/add_user'}, 'number': '24', 'html_url': 'www.github.com'},
         {'head': {'ref': 'ignore-trailing'}, 'user': {'login': 'github_user'}, 'base': {'ref': 'hotfix/add-trigger'}, 'number': '3', 'html_url': 'www.github.com'}
     ])
 
-    @mock.patch('git_machete.options.CommandLineOptions', FakeCommandLineOptions)
     @mock.patch('git_machete.github.GITHUB_REMOTE_PATTERNS', FAKE_GITHUB_REMOTE_PATTERNS)
```
PawelLipski, line 1826: I'd still add a comment here that explains
```
We need to mock GITHUB_REMOTE_PATTERNS in the tests for `test_checkout_prs` due to `git fetch` executed by `checkout-prs` subcommand.
```

git_machete/client.py
```diff
@@ -1670,14 +1670,16 @@ def checkout_github_pr(self, pr_no: int) -> None:
...
 
         if not pr:
-            raise MacheteException(f"PR number {pr_no} is not found in repository {repo}")
+            raise MacheteException(f"PR #{pr_no} is not found in repository {org}/{repo}")
         debug('checkout_github_pr()', f'found {pr}')
 
         print(f"Fetching {remote}...")
         self.__git.fetch_remote(remote)
+        if self.__git.remotes():
+            self.flush_caches()
```
PawelLipski, line 1679: Let's leave as is for now, but now I see this requires a more systematic approach: pls add a TODO here for https://github.com/VirtusLab/git-machete/issues/237

git_machete/github.py
rifel123: Not without changes, `find_or_none` (if extracted) would return PR object which has head equal to base of PR passed to this function. In this case, i would expect to return current PR, not the next one in pr chain.

git_machete/client.py
rifel123: Yes, but then i would need to annotate this function arguments with `GithubPullRequests` and this will cause circular import error.

PawelLipski: Nope! pls make this function generic, see the mentioned `flat_map`

git_machete/github.py
PawelLipski: Hmmm? how about sth like `find_or_none(lambda pr: number == pr.number, pull_requests)`? can't see how that's limited to _returning PR object which has head equal to base of PR passed to this function_

git_machete/tests/functional/test_machete.py
```diff
@@ -1670,10 +1673,10 @@ def test_anno_prs(self) -> None:
     git_api_state_for_test_create_pr = MockGithubAPIState([{'head': {'ref': 'ignore-trailing'}, 'user': {'login': 'github_user'}, 'base': {'ref': 'hotfix/add-trigger'}, 'number': '3', 'html_url': 'www.github.com'}])
 
     @mock.patch('git_machete.options.CommandLineOptions', FakeCommandLineOptions)
+    @mock.patch('git_machete.github.GITHUB_REMOTE_PATTERNS', FAKE_GITHUB_REMOTE_PATTERNS)
```
rifel123: Here, `FAKE_GITHUB_REMOTE_PATTERNS` was added because somehow, `Remotes are defined for this repository, but none of them corresponds to GitHub` error is still raised. Same logic is used in `test_retarget_pr` and `test_anno_prs`, where solution works as expected.

PawelLipski: Whoops okay this error might be specific not just to the tests, but also happen in the actual (production) use. Let's not ignore it... pls:
1. remove this `@mock.patch`
2. add some extra information to the `Remotes are ...` error message, like e.g. an output of `git remotes -v` for instance; it's even okay if we release v3.4.0 with that extended error message

git_machete/tests/functional/test_machete.py
```diff
@@ -1822,6 +1823,7 @@ def test_github_create_pr(self) -> None:
         {'head': {'ref': 'ignore-trailing'}, 'user': {'login': 'github_user'}, 'base': {'ref': 'hotfix/add-trigger'}, 'number': '3', 'html_url': 'www.github.com'}
     ])
 
+    # We need to mock GITHUB_REMOTE_PATTERNS in the tests for `test_checkout_prs` due to `git fetch` executed by `checkout-prs` subcommand.
```
PawelLipski, line 1825: This one is okay coz you've explained the underlying mechanism to me, it doesn't seem that any bug is lurking here, unlike with `create-pr` above

git_machete/tests/functional/test_machete.py
PawelLipski: Although... wait, isn't the problem related to the `git push` that happens before a PR is opened? pls investigate 🕵🏻

PawelLipski: If so, then mocking `GITHUB_REMOTE_PATTERNS` is probably a reasonable idea, same as in the test for `checkout-prs`

git_machete/tests/functional/test_machete.py
```diff
@@ -1808,3 +1811,256 @@ def test_github_create_pr(self) -> None:
         if e:
             self.assertEqual(e.exception.parameter, expected_error_message,
                              'Verify that expected error message has appeared when given pull request to create is already created.')
+
+    git_api_state_for_test_checkout_pr = MockGithubAPIState([
```
PawelLipski: Nit: `..._prs`

git_machete/tests/functional/test_machete.py
```diff
@@ -1808,3 +1811,256 @@ def test_github_create_pr(self) -> None:
...
+        (org, repo) = get_parsed_github_remote_url(self.repo_sandbox.remote_path)
+        expected_error_message = f"PR #100 is not found in repository `{org}/{repo}`"
+        machete_client.read_definition_file()
+        with self.assertRaises(MacheteException) as e:
+            machete_client.checkout_github_prs(100)
+        if e:
+            self.assertEqual(e.exception.parameter, expected_error_message,
+                             'Verify that expected error message has appeared when given pull request to checkout does not exists.')
+
+    git_api_state_for_test_checkout_pr_fresh_repo = MockGithubAPIState([
```
PawelLipski: Ditto, `prs`

git_machete/tests/functional/test_machete.py
```diff
@@ -1808,3 +1811,256 @@ def test_github_create_pr(self) -> None:
...
+        {'head': {'ref': 'comments/add_docstrings'}, 'user': {'login': 'github_user'}, 'base': {'ref': 'improve/refactor'}, 'number': '2', 'html_url': 'www.github.com'},
+        {'head': {'ref': 'restrict_access'}, 'user': {'login': 'github_user'}, 'base': {'ref': 'allow-ownership-link'}, 'number': '17', 'html_url': 'www.github.com'},
+        {'head': {'ref': 'improve/refactor'}, 'user': {'login': 'github_user'}, 'base': {'ref': 'chore/sync_to_docs'}, 'number': '1', 'html_url': 'www.github.com'},
+    ])
+
+    # We need to mock GITHUB_REMOTE_PATTERNS in the tests for `test_checkout_prs` due to `git fetch` executed by `checkout-prs` subcommand.
```
PawelLipski: Nit:
```suggestion
    # We need to mock GITHUB_REMOTE_PATTERNS in the tests for `test_checkout_prs_freshly_cloned` due to `git fetch` executed by `checkout-prs` subcommand.
```

PawelLipski (commented):
Almost there... minor comments left

git_machete/client.py
```diff
@@ -1746,7 +1746,9 @@ def __derive_remote_and_github_org_and_repo(self) -> Tuple[str, Tuple[str, str]]
         if not org_and_repo_for_github_remote:
             raise MacheteException(
                 fmt('Remotes are defined for this repository, but none of them '
-                    'corresponds to GitHub (see `git remote -v` for details)'))
+                    'corresponds to GitHub (see `git remote -v` for details)\n'
```
PawelLipski: Any success in replicating the error? is it due to `git push` indeed?

git_machete/tests/functional/test_machete.py
rifel123: Problem is very tricky. Unfortunately it's not connected to `push` operation because error is raised before, while program determines what is correct remote for github, at `__derive_remote_and_github_org_and_repo`. Since error occurs always at same line which points to line 1815 in test, where `create_github_pr` is called directly from `MacheteClient` instance, not from `launch` command. During failures `__remotes_cached` holds only remote but `git remote -v` outputs two as it's supposed to be. 

I tried to catch it on my machine , with same  test repository setup and with direct calls to `create_github_pr`, but  on 1000 tries every single one was correct. Same story with running tox locally. I also tried to `flush_caches` right before `__derive_remote_and_github_org_and_repo` to enforce reading remotes again, but it doesn't solve the problem. Adding delays does not do the job too.

rifel123: finally clearing out `__remotes_cached` with `flush_caches` fixes the problem. 

But it's still unknown, why it happens so randomly, and only in CI environment

git_machete/client.py
rifel123: I think so 👯 , see comments above.

git_machete/client.py
```diff
@@ -1765,7 +1767,7 @@ def create_github_pr(self, head: str, draft: bool) -> None:
         except MacheteException as e:
             print(e)
             return
-
+        self.flush_caches()
```
PawelLipski, line 1768: Looks like another piece to be ultimately sorted out by @BartoszSiwinski 's task #237

git_machete/git_operations.py
```diff
@@ -413,6 +413,7 @@ def flush_caches(self) -> None:
         self.__reflogs_cached = None
         self.__remaining_log_shas_cached = {}
         self.__remote_branches_cached = None
+        self.__remotes_cached = None
```
PawelLipski, line 416: Commented about that on #237

git_machete/tests/functional/test_machete.py
```diff
@@ -1672,8 +1672,8 @@ def test_anno_prs(self) -> None:
 
     git_api_state_for_test_create_pr = MockGithubAPIState([{'head': {'ref': 'ignore-trailing'}, 'user': {'login': 'github_user'}, 'base': {'ref': 'hotfix/add-trigger'}, 'number': '3', 'html_url': 'www.github.com'}])
 
+    # We need to mock GITHUB_REMOTE_PATTERNS in the tests for `test_create_pr` to avoid situation where there is no remotes pointing to Github.
```
PawelLipski: No longer relevant, remove

git_machete/client.py
```diff
@@ -1746,7 +1746,9 @@ def __derive_remote_and_github_org_and_repo(self) -> Tuple[str, Tuple[str, str]]
         if not org_and_repo_for_github_remote:
             raise MacheteException(
                 fmt('Remotes are defined for this repository, but none of them '
-                    'corresponds to GitHub (see `git remote -v` for details)'))
+                    'corresponds to GitHub (see `git remote -v` for details)\n'
+                    f'Actual remotes from git:\n {self.__git.popen_git("remote", "-v")}\n'
+                    f'Actual remotes from git-machete: {self.__git.remotes()}'))
```
PawelLipski: This line is maybe too low-level to display 😅 also, I think with the addition of `self.__remotes_cached = None` in `flush_caches` the error is no longer likely to happen right?

git_machete/client.py
PawelLipski: Remove `(see `git remote -v` for details)` now that you display the result of this command 😅

git_machete/client.py
```suggestion
                    'corresponds to GitHub (see `git remote -v` for details):\n\n'
                    f'{self.__git.popen_git("remote", "-v")}'))
```

PawelLipski (commented):
Even smaller comments...

git_machete/client.py
rifel123: right, i'll just remove those extended messages.

PawelLipski (approved):
🥳

PR #214, opened 2021-09-07
Add regression tests for github methods

rifel123:


git_machete/git_operations.py
```diff
@@ -161,7 +161,7 @@ def remotes(self) -> List[str]:
         return self.__remotes_cached
 
     def get_url_of_remote(self, remote: str) -> str:
-        return self.popen_git("remote", "get-url", "--", remote).strip()
+        return self.popen_git("config", "--get", f"remote.{remote}.url").strip()
```
rifel123: Because `get-url` is not supported in git 1.8

git_machete/tests/functional/test_machete.py
```diff
@@ -25,6 +28,170 @@ def get_head_commit_hash() -> str:
...
+        self.opt_override_to_parent: bool = False
+        self.opt_push_tracked: Optional[bool] = True
+        self.opt_push_untracked: Optional[bool] = True
+        self.opt_return_to: str = "stay"
+        self.opt_roots: List[str] = list()
+        self.opt_start_from: str = "here"
+        self.opt_stat: bool = False
+        self.opt_sync_github_prs: bool = False
+        self.opt_unset_override: bool = False
+        self.opt_yes: bool = True
```
rifel123, line 37: This is needed to set `opt_yes` and `opt_no_interactive_rebase` to true, in order to perform `__sync_before_creating_pr`. The other way to acomplish such a settitng in tests, is to call `create_github_pr` directly from MacheteClient class, like it is in lines from 1780 till bottom. Which way would be better?

git_machete/git_operations.py
PawelLipski: Whoops if you could extract this change to a separate PR

git_machete/tests/functional/test_machete.py
```diff
@@ -25,6 +28,170 @@ def get_head_commit_hash() -> str:
...
+
+
+class GitAPIState:
+    pulls: List[Dict[str, Any]] = [{'head': {'ref': 'bugfix/remove'}, 'user': {'login': 'github_user'}, 'base': {'ref': 'develop'}, 'number': '15', 'html_url': 'www.github.com'}]
+    user: Dict[str, str] = {'login': 'github_user', 'type': 'User', 'company': 'VirtusLab'}
+    issues: List[Dict[str, Any]] = []
+    remote_branches: List[str] = []
+
+    def __init__(self, url: str, headers: Dict[str, str] = None, data: Union[str, bytes, None] = None, method: str = ''):
```
PawelLipski: Looks nice <3

git_machete/git_operations.py
PawelLipski: I'm just looking into git sources... seems that it's only been added in 2015 lol

PawelLipski: git v2.5.1 or so

git_machete/tests/functional/test_machete.py
```diff
@@ -25,6 +28,170 @@ def get_head_commit_hash() -> str:
...
+
+    @staticmethod
+    def get_new_number(entities: List[Dict[str, Any]]) -> str:
+        numbers = []
+        for item in entities:
+            numbers.append(int(item['number']))
+        return str(max(numbers) + 1)
+
+
+class Opener:
```
PawelLipski: This should probably be called `...ContextManager`: https://www.python.org/dev/peps/pep-0343/#id33

git_machete/tests/functional/test_machete.py
```diff
@@ -1381,3 +1552,235 @@ def test_update_with_fork_point_specified(self) -> None:
...
+            """,
+        )
+        # check against attempt to create already existing pull request
+        machete_client = cmd.MacheteClient(cli_opts, git)
+        expected_error_message = "Pull request for branch hotfix/add-trigger is already created under link www.github.com!\nPR details: PR #6 by githubuser: hotfix/add-trigger -> master"
+        try:
+            machete_client.create_github_pr('hotfix/add-trigger', draft=False)
+        except MacheteException as e:
+            if e.parameter != expected_error_message:
+                self.fail(f'Actual Exception message: {e} \nis not equal to expected message: {expected_error_message}')
```
BartoszSiwinski: Your version won't fail if no error was raised.

Proposal below checks if `error` was `raised`, so if it didn't it will fail, which is desired.
If error was raised, so `e` is not `None`, we check that `msg` is as expected. If no error was raised this will be `None` so we can skip it (and we should, because `None` does not have argument `parameter`)
```suggestion
with self.assertRaises(MacheteException) as e:  
    machete_client.create_github_pr(....)
 if e:
     self.assertEqual(e.parameter, expected_error_message, 'Verify that expected error message has appeared when <insert_valid_conditions_of_the_tests>'). 

```

This follows also a bit convention presented in example to assertRaises in untitest lib:
```
with self.assertRaises(SomeException) as cm:
   do_something()
the_exception = cm.exception
self.assertEqual(the_exception.error_code, 3)
```

git_machete/git_operations.py
rifel123: moved to separate PR #219

git_machete/tests/functional/test_machete.py
rifel123: yes, you are right, good spot :)

git_machete/tests/functional/test_machete.py
BartoszSiwinski: Other option would be to inherit from `CommandLineOptions` and in `init` first call `init from super` and then adjusting only these variables that should be adjusted (preferable with some comment justifying these adjustments). This solution would be fairly small, ale resilient to some of the changes that can come (i.e. addition of a new instance variable that is required in create_github_pr).

This worked for me when I run it locally.
```
class FakeCommandLineOptions(CommandLineOptions):

    def __init__(self) -> None:
        super().__init__()
        self.opt_no_interactive_rebase: bool = True
        self.opt_yes: bool = True
```

git_machete/tests/functional/test_machete.py
```diff
@@ -1381,3 +1552,235 @@ def test_update_with_fork_point_specified(self) -> None:
...
+                .push()
+                .reset_to("ignore-trailing@{1}")
+                .delete_branch("root")
+                .new_branch('chore/fields')
+                .commit("remove outdated fields")
+                .add_remote('new_origin', 'https://github.com/user/repo.git')
+                .check_out("call-ws")
+        )
+        GitAPIState.pulls = [{'head': {'ref': 'ignore-trailing'}, 'user': {'login': 'github_user'}, 'base': {'ref': 'hotfix/add-trigger'}, 'number': '3', 'html_url': 'www.github.com'}]
+        self.launch_command("discover", "-y")
```
BartoszSiwinski: Is this `-y` needed here if `FakeCommandLineOptions` already sets `self.opt_yes`?

git_machete/tests/functional/test_machete.py
```diff
@@ -25,6 +28,170 @@ def get_head_commit_hash() -> str:
...
+        self.status_code: Optional[int] = None
+        self.json_data: Union[str, bytes] = data
+        self.return_data: Optional[Union[List[Dict[str, Any]], Dict[str, Any]]] = None
+        self.headers: Dict[str, str] = headers
+        if method == "GET":
+            self.handle_get()
+        elif method == "PATCH":
+            self.handle_patch()
+        elif method == "POST":
+            self.handle_post()
```
BartoszSiwinski: Proposal: could be separate method `handle_method_argument()` that does that `if`.

git_machete/tests/functional/test_machete.py
```diff
@@ -25,6 +28,170 @@ def get_head_commit_hash() -> str:
...
+                self.update_issue(new=True)
+        elif 'pulls' in self.url:
+            if self.url.endswith('pulls'):
+                self.update_pull(new=True)
+
+    def update_pull(self, new: bool = False) -> None:
+        pull: Dict[str, Any]
+        if new:
+            pull = {'number': self.get_new_number(GitAPIState.pulls),
+                    'user': {'login': 'githubuser'},
```
BartoszSiwinski: Minor: I would rename it slightly, because some people when they read code they're paranoid (see @me) and they're read such places in case someone decided to put a crucial information like `login:not_really_login_actually_a_password`, because of some dirty hack somewhere else. With _ You can read it faster and at a first glance understand what it says. As a plus it will stop looking like `git abuser` 🗡️ 

```suggestion
                    'user': {'login': 'github_user'},
```

git_machete/tests/functional/test_machete.py
```diff
@@ -25,6 +28,170 @@ def get_head_commit_hash() -> str:
...
+    @staticmethod
+    def find_number(url: str) -> str:
+        m = re.search(r'\d+', url)
+        return m.group()
+
+    @staticmethod
+    def get_new_number(entities: List[Dict[str, Any]]) -> str:
+        numbers = []
+        for item in entities:
+            numbers.append(int(item['number']))
```
BartoszSiwinski:
```suggestion
        numbers = [int(item['number']) for item in entities]
```

git_machete/tests/functional/test_machete.py
```diff
@@ -25,6 +28,170 @@ def get_head_commit_hash() -> str:
...
+                return GitAPIState.pulls.pop(index)
+        return None
+
+    @staticmethod
+    def find_number(url: str) -> str:
+        m = re.search(r'\d+', url)
+        return m.group()
+
+    @staticmethod
+    def get_new_number(entities: List[Dict[str, Any]]) -> str:
```
BartoszSiwinski: It's just a proposal. It read maybe a bit worse, but it stays what it exactly does. Your choice
```suggestion
    def get_next_free_number(entities: List[Dict[str, Any]]) -> str:
```

git_machete/tests/functional/test_machete.py
PawelLipski: Sounds good, this way we can clearly see which options are actually overridden

git_machete/tests/functional/test_machete.py
```diff
@@ -1381,3 +1552,236 @@ def test_update_with_fork_point_specified(self) -> None:
...
+              |
+              x-drop-constraint (untracked)
+            """,
+        )
+        # check against attempt to create already existing pull request
+        machete_client = cmd.MacheteClient(cli_opts, git)
+        expected_error_message = "Pull request for branch hotfix/add-trigger is already created under link www.github.com!\nPR details: PR #6 by githubuser: hotfix/add-trigger -> master"
+        with self.assertRaises(MacheteException) as e:
+            machete_client.read_definition_file()
+            machete_client.create_github_pr('hotfix/add-trigger', draft=False)
```
BartoszSiwinski: Which line should raise exception? I would expect that only one line will be in the context and other, that we don't expect to raise exception will not be in the context. Otherwise we can get all sort of wrong results from the assertRaises.

git_machete/tests/functional/test_machete.py
BartoszSiwinski: If you have these variables as class variables aren't you modifying them in each tests and these modification persist from test to test?

This is rather risky, because now environment in which you're running your tests is dependent on what other tests have been run before.

PawelLipski: uuu

PawelLipski:
Whoops must be rebased onto recent develop... also https://github.com/VirtusLab/git-machete/issues/220

git_machete/tests/functional/test_machete.py
rifel123: That was intended. New instance of `GithubAPIState` is created on each call to github API and each test makes several calls. It's not desired to lose this data. 
When needed, initial values of class variables are set on the beginning of test.

git_machete/tests/functional/test_machete.py
rifel123: assertEqual will catch desired exception, but ok.

git_machete/tests/functional/test_machete.py
BartoszSiwinski: I stand by my opinion that it's risky. I had in the past several tests in other project that were failing due to the sequence of the tests (they were not cleaning properly after themselves) and debugging/fixing such issues is a lengthy nightmare. If certain environment is expected before test or tests require certain setup you can put make a method out of it and start a test with it. If it's not required for test than we can clear it in teardown via `self.issues=[]`

Also if you require one test to be run after another one, it may break eventually while running with `stestr` or when one of the tests will be deleted/modified. It's asking for trouble in maintenance.

PawelLipski: I don't have time today to analyse it more deeply... but if that's the case that passing or failure of any test **depends on the order in which the tests are executed**, then I'm also strongly against such approach ☹️

git_machete/tests/functional/test_machete.py
```diff
@@ -26,6 +29,148 @@ def get_head_commit_hash() -> str:
...
+    def set_initial_values() -> None:
+        GitAPIState.pulls = [
+            {'head': {'ref': 'bugfix/remove'}, 'user': {'login': 'github_user'}, 'base': {'ref': 'develop'},
+             'number': '15', 'html_url': 'www.github.com'}]
+        GitAPIState.user = {'login': 'github_user', 'type': 'User', 'company': 'VirtusLab'}
+        GitAPIState.issues = []
+        GitAPIState.remote_branches = []
+
+
+class ContextManager:
```
BartoszSiwinski: It would be good to specify what kind of context it manages. Given its implementation and purpose I would suggest one of the followings:
- `MockContextManager` <- my preference
- `EmptyContextManager`
- `FakeContextManager`

PawelLipski: Right, that's what I suggested originally (albeit implicitly) by `...ContextManager` ;)

git_machete/client.py
```diff
@@ -18,7 +19,6 @@
     get_parsed_github_remote_url, is_github_remote_url, create_pull_request,
     derive_current_user_login, set_milestone_of_pull_request,
     add_assignees_to_pull_request, add_reviewers_to_pull_request)
-from git_machete.options import CommandLineOptions
```
PawelLipski, line 21: Any reason for that change? 🤔

git_machete/tests/functional/test_machete.py
```diff
@@ -26,6 +29,148 @@ def get_head_commit_hash() -> str:
     return os.popen("git rev-parse HEAD").read().strip()
 
 
+class FakeCommandLineOptions(CommandLineOptions):
+    def __init__(self) -> None:
+        super().__init__()
```
PawelLipski, line 35: Whoops so without explicit call to superclass c'tor... it just won't be called by default? or an error will be thrown?

git_machete/tests/functional/test_machete.py
```diff
@@ -1382,3 +1531,239 @@ def test_update_with_fork_point_specified(self) -> None:
                 "<commit_hash>' drops the commits until (included) fork point "
                 "specified by the option '-f' from the current branch."
         )
+
+    @mock.patch('urllib.request.urlopen', ContextManager)
+    @mock.patch('urllib.request.Request', GitAPIState)
```
PawelLipski: Hmmm since `GitAPIState` is a mock for **Request**... should it be called `GitAPIState`? Shouldn't it rather be called `MockGithubAPIRequest` instead?

git_machete/tests/functional/test_machete.py
```diff
@@ -1382,3 +1531,239 @@ def test_update_with_fork_point_specified(self) -> None:
...
+                .push()
+                .new_branch('feature')
+                .commit('introduce feature')
+                .push()
+                .check_out('feature')
+                .add_remote('new_origin', 'https://github.com/user/repo.git')
+        )
+        GitAPIState.pulls = [{'head': {'ref': 'feature'}, 'user': {'login': 'github_user'}, 'base': {'ref': 'root'}, 'number': '15', 'html_url': 'www.github.com'}]
+        self.launch_command("discover", "-y")
+        self.assert_command(['github', 'retarget-pr'], 'The base branch of PR #15 has been switched to `branch-1`\n', adapt=False)
```
PawelLipski: I'd change the `adapt` param name to sth more adequate 🤔

PawelLipski: Since it's hard to figure out this param's purpose from its name now

git_machete/tests/functional/test_machete.py
```diff
@@ -1382,3 +1531,239 @@ def test_update_with_fork_point_specified(self) -> None:
...
+        )
+
+        (
+            self.repo_sandbox.check_out('hotfix/add-trigger')
+            .commit('trigger released')
+            .commit('minor changes applied')
+        )
+
+        # diverged from and newer than origin
+        self.launch_command("github", "create-pr")
```
PawelLipski, line 1779: I'd add `--draft` option in one of the cases

git_machete/tests/functional/test_machete.py
```diff
@@ -1382,3 +1531,239 @@ def test_update_with_fork_point_specified(self) -> None:
...
+              |
+              x-drop-constraint (untracked)
+            """,
+        )
+
+    @mock.patch('git_machete.options.CommandLineOptions', FakeCommandLineOptions)
+    @mock.patch('urllib.request.urlopen', ContextManager)
+    @mock.patch('urllib.request.Request', GitAPIState)
+    def test_github_create_pr(self) -> None:
+        GitAPIState.set_initial_values()
```
PawelLipski: `set_initial_values()` probably belongs to the unittest's standard `setUp()` method

git_machete/tests/functional/test_machete.py
```diff
@@ -26,6 +29,148 @@ def get_head_commit_hash() -> str:
...
+            self.handle_patch()
+        elif method == "POST":
+            self.handle_post()
+
+    def read(self) -> Union[str, bytes]:
+        return json.dumps(self.return_data).encode()
+
+    def handle_get(self) -> None:
+        if 'pulls' in self.url:
+            if '?head' in self.url:
```
PawelLipski: Instead of this brittle low-level parsing... I'd rather use some library to parse the URL (esp. extract query params from after `?`)

git_machete/tests/functional/test_machete.py
```diff
@@ -26,6 +29,148 @@ def get_head_commit_hash() -> str:
...
+            issue_no = self.find_number(self.url)
+            issue = self.get_issue(issue_no)
+        for key in json.loads(self.json_data).keys():
+            issue[key] = json.loads(self.json_data)[key]
+        GitAPIState.issues.append(issue)
+        self.status_code = 201
+        self.return_data = issue
+        return
+
+    @staticmethod
```
PawelLipski, line 174: Hmmm there's quite a lot a static methods (and a lot of class attributes) in this class... this suggests that probably this class should be split into two classes, one with what's now static attributes but converted to instance attributes, and another with what's currently instance attributes. I'd imagine something like:

```python
class MockGitAPIState:
  # pulls, user, issues etc. are instance attributes
  def get_issue(...): ...
  # etc.
  def new_request() -> "MockGitAPIRequest":
   ...


class MockGitAPIRequest:
  def get_my_api_state() -> MockGitAPIState: return self.api_state

  def handle_get(...): ...
  # etc.
```

The question remains how to change `@mock.patch` annotations then... probably sth like:

```python
git_api_state_for_test_retarget_pr = MockGitAPIState({'head': {'ref': 'feature'}, 'user': {'login': 'github_user'}, 'base': {'ref': 'root'}, 'number': '15', 'html_url': 'www.github.com'})

@mock.patch('urllib.request.Request', git_api_state_for_test_retarget_pr.new_request)
....
```

**General idea**: the less global (incl. static) state, the better** (the less risk of interference b/w individual test cases)

git_machete/tests/functional/test_machete.py
rifel123: Without `super()` it won't be called. It has to be told explicitly to python to call it.

git_machete/tests/functional/test_machete.py
rifel123: A little bit outdated: since `MockGitAPIState` was extracted from `MockGitAPIRequest` i don't see a posiibility to reach it out from `setUp()` method

PawelLipski: Whoops right, this is no longer relevant, resolving

git_machete/client.py
rifel123: this is connected to how mocking mechanism works. In fact it mocks a namespace. on the other hand `from import` creates a new namespace within current file scope which makes from imported stuff unmockable.

git_machete/tests/functional/test_machete.py
```diff
@@ -36,18 +36,48 @@ def __init__(self) -> None:
...
+
+
+class MockGitAPIRequest:
+
+    git_api_state: MockGitAPIState = MockGitAPIState()
 
     def __init__(self, url: str, headers: Dict[str, str] = None, data: Union[str, bytes, None] = None, method: str = ''):
-        self.url: str = url
+        self.parsed_url: ParseResult = urlparse(url, allow_fragments=True)
+        self.parsed_query = parse_qs(self.parsed_url.query)
```
PawelLipski: Nit: pls add explicit type here if possible for clarity

git_machete/tests/functional/test_machete.py
```diff
@@ -72,74 +104,60 @@ def handle_get(self) -> None:
...
             self.update_issue()
-        elif 'pulls' in self.url:
+        elif 'pulls' in self.parsed_url.path:
             self.update_pull()
 
     def handle_post(self) -> None:
-        if 'issues' in self.url:
-            if self.url.endswith('issues'):
+        if self.parsed_url.path.endswith('issues'):
+            if not self.parsed_query:
```
PawelLipski: In fact since you're NOT expecting any query params to be present (I guess that's the intention here), you can add an assertion

git_machete/tests/functional/test_machete.py
```diff
@@ -1532,10 +1541,10 @@ def test_update_with_fork_point_specified(self) -> None:
                 "specified by the option '-f' from the current branch."
         )
 
-    @mock.patch('urllib.request.urlopen', ContextManager)
-    @mock.patch('urllib.request.Request', GitAPIState)
+    @mock.patch('urllib.request.urlopen', MockContextManager)
+    @mock.patch('urllib.request.Request', MockGitAPIRequest)
     def test_retarget_pr(self) -> None:
-        GitAPIState.set_initial_values()
+        MockGitAPIRequest.git_api_state.set_initial_values()
```
PawelLipski: It's definitely cleaner now after the split of MockRequest/MockAPIState... but still, there's a **global mutable state** here, **shared b/w all test cases** :/

PawelLipski: Has there been any obstacle in following the path suggested in https://github.com/VirtusLab/git-machete/pull/214/files/674de5bc22d1ab98ef84227514ea9e7edb17438f..de3711b32d3d14f93485aa6bfdaa4c11c70b7b89#r705527546 ? maybe `new_callable` (instead of `new`) param instead of `mock.patch` should be used? 🤔 so that you can construct a `MockGitAPIRequest` object in `MockGitAPIState.new_request`, instead of relying on mock framework to construct `MockGitAPIRequest`s :/

PawelLipski: (I've introduced some minor fixes to the snippet with the new suggested layout)

PawelLipski: To recap: we want **each test case to have its own `MockGitAPIState`** that is NOT shared with other test cases.

git_machete/tests/functional/test_machete.py
```diff
@@ -36,18 +36,48 @@ def __init__(self) -> None:
         self.opt_yes: bool = True
 
 
-class GitAPIState:
-    pulls: List[Dict[str, Any]] = [{'head': {'ref': 'bugfix/remove'}, 'user': {'login': 'github_user'}, 'base': {'ref': 'develop'}, 'number': '15', 'html_url': 'www.github.com'}]
-    user: Dict[str, str] = {'login': 'github_user', 'type': 'User', 'company': 'VirtusLab'}
-    issues: List[Dict[str, Any]] = []
-    remote_branches: List[str] = []
+class MockGitAPIState:
```
BartoszSiwinski: Is this mock common for all git based services like gitlab and github or is it only for github? If it's only for github it could be a good idea to call it `MockGithubAPIState`.

git_machete/tests/functional/test_machete.py
```diff
@@ -72,74 +104,60 @@ def handle_get(self) -> None:
                 return
             else:
                 self.status_code = 200
```
BartoszSiwinski:
```suggestion
from http import HTTPStatus
<....skip....>
                self.status_code = HTTPStatus.OK
```

git_machete/tests/functional/test_machete.py
```diff
@@ -72,74 +104,60 @@ def handle_get(self) -> None:
...
         else:
-            issue_no = self.find_number(self.url)
-            issue = self.get_issue(issue_no)
+            issue_no = self.find_number(self.parsed_url.path)
+            issue = MockGitAPIRequest.git_api_state.get_issue(issue_no)
         for key in json.loads(self.json_data).keys():
             issue[key] = json.loads(self.json_data)[key]
-        GitAPIState.issues.append(issue)
+        MockGitAPIRequest.git_api_state.issues.append(issue)
         self.status_code = 201
```
BartoszSiwinski:
```suggestion
        self.status_code = HTTPStatus.CREATED
```

git_machete/tests/functional/test_machete.py
PawelLipski: Yeah, things like issue and pull request are unknown to git as such, these are Github-specific terms... so it should definitely  be `MockGithubAPIState/Request`

git_machete/tests/functional/test_machete.py
```diff
@@ -36,110 +37,128 @@ def __init__(self) -> None:
...
+            if pull['number'] == pull_no:
+                return self.pulls.pop(index)
+        return None
+
+
+class ResponseWrapper:
+    def __init__(self, data: Union[List[Dict[str, Any]], Dict[str, Any]]) -> None:
+        self.data: Union[List[Dict[str, Any]], Dict[str, Any]] = data
+
+    def read(self) -> Union[str, bytes]:
```
PawelLipski: Hmm I'd expect it to be just `bytes`... why union with `str` as well?

git_machete/tests/functional/test_machete.py
```diff
@@ -36,110 +37,128 @@ def __init__(self) -> None:
...
+        self.user: Dict[str, str] = {'login': 'github_user', 'type': 'User', 'company': 'VirtusLab'}
+        self.issues: List[Dict[str, Any]] = []
+
+    def new_request(self) -> "MockGithubAPIRequest":
+        return MockGithubAPIRequest(self)
+
+    def get_issue(self, issue_no: str) -> Optional[Dict[str, Any]]:
+        for index, issue in enumerate(self.issues):
+            if issue['number'] == issue_no:
+                return self.issues.pop(index)
```
PawelLipski: Hmmm it **removes** the issue from the collection? that's an unexpected side effect for a method called `get_...`

git_machete/tests/functional/test_machete.py
```diff
@@ -36,110 +37,128 @@ def __init__(self) -> None:
...
+        return None
+
+    def get_pull(self, pull_no: str) -> Optional[Dict[str, Any]]:
+        for index, pull in enumerate(self.pulls):
+            if pull['number'] == pull_no:
+                return self.pulls.pop(index)
```
PawelLipski: Same here, why remove the pull request?

git_machete/tests/functional/test_machete.py
```diff
@@ -36,110 +37,128 @@ def __init__(self) -> None:
...
+    def read(self) -> Union[str, bytes]:
+        return json.dumps(self.data).encode()
+
+
+class MockGithubAPIRequest:
 
-    def __init__(self, url: str, headers: Dict[str, str] = None, data: Union[str, bytes, None] = None, method: str = ''):
-        self.url: str = url
+    def __init__(self, git_api_state: MockGithubAPIState) -> None:
+        self.git_api_state: MockGithubAPIState = git_api_state
```
PawelLipski: Rename the param&attribute accordingly: `github_api_state`

git_machete/tests/functional/test_machete.py
```diff
@@ -36,110 +37,128 @@ def __init__(self) -> None:
...
+
+    def __call__(self, url: str, headers: Dict[str, str] = None, data: Union[str, bytes, None] = None, method: str = '') -> "ResponseWrapper":
+        self.parsed_url: ParseResult = urlparse(url, allow_fragments=True)
+        self.parsed_query: Dict[str, List[str]] = parse_qs(self.parsed_url.query)
         self.status_code: Optional[int] = None
         self.json_data: Union[str, bytes] = data
         self.return_data: Optional[Union[List[Dict[str, Any]], Dict[str, Any]]] = None
         self.headers: Dict[str, str] = headers
+        self.handle_method(method)
+        return ResponseWrapper(self.return_data)
```
PawelLipski: I'd simplify the logic here: `handle_method` can return the response instead of assigning it to an attribute

git_machete/tests/functional/test_machete.py
```diff
@@ -36,110 +37,128 @@ def __init__(self) -> None:
...
 
-    def __init__(self, url: str, headers: Dict[str, str] = None, data: Union[str, bytes, None] = None, method: str = ''):
-        self.url: str = url
+    def __init__(self, git_api_state: MockGithubAPIState) -> None:
+        self.git_api_state: MockGithubAPIState = git_api_state
+
+    def __call__(self, url: str, headers: Dict[str, str] = None, data: Union[str, bytes, None] = None, method: str = '') -> "ResponseWrapper":
+        self.parsed_url: ParseResult = urlparse(url, allow_fragments=True)
+        self.parsed_query: Dict[str, List[str]] = parse_qs(self.parsed_url.query)
         self.status_code: Optional[int] = None
```
PawelLipski: Hmm I think these attributes should probably be set on the `ResponseWrapper` rather than on `MockGithubAPIRequest`

PawelLipski: I don't see the use for `status_code` now 🤔

git_machete/tests/functional/test_machete.py
```diff
@@ -36,110 +37,128 @@ def __init__(self) -> None:
...
-        if 'pulls' in self.url:
-            if '?head' in self.url:
-                head: str = self.url[self.url.rfind(':') + 1:]
-                for pr in GitAPIState.pulls:
-                    if pr['head']['ref'] == head:
+        if self.parsed_url.path.endswith('pulls'):
+            full_head_name: Optional[List[str]] = self.parsed_query.get('head')
+            if full_head_name:
+                head: str = full_head_name[0]
+                head_name: str = head[head.rfind(':') + 1:]
```
PawelLipski: Or maybe just sth like `head.split(':')[0]` ?

git_machete/tests/functional/test_machete.py
```diff
@@ -36,110 +37,128 @@ def __init__(self) -> None:
...
-        elif 'pulls' in self.url:
-            if self.url.endswith('pulls'):
-                self.update_pull(new=True)
+        assert not self.parsed_query
+        if self.parsed_url.path.endswith('issues'):
+            self.update_issue(new=True)
+        elif self.parsed_url.path.endswith('pulls'):
+            self.update_pull(new=True)
 
     def update_pull(self, new: bool = False) -> None:
```
PawelLipski: As discussed in one of the past PRs (the one about refactoring shell code), I'd avoid making switches based on the flag value. Pls extract separate `create_pull_request` and `update_pull_request` methods and delegate the shared parts of logic to a third method if needed

git_machete/tests/functional/test_machete.py
rifel123: hmm, actually i don't see point to have `ResposneWrapper`. Let's drag `read` method into `MockGithubAPIRequest`, make private everything which should not be visible outside and return `self` in `__call__`.

PawelLipski: Hmm I'd still insist on extracting a separate class for response... rather than mixing request-specific and response-specific functionality in a single class, that's gonna be very confusing (_why is there a `return_code` and `read()` on a Request class_)?

git_machete/tests/functional/test_machete.py
rifel123: It was used for update only where i had to replace given pull/issue anyway. Right now for possible future usage, ` get_pull/issue` doesn't remove item from collection.

git_machete/tests/functional/test_machete.py
```diff
@@ -47,122 +47,147 @@ def new_request(self) -> "MockGithubAPIRequest":
...
-            self.handle_get()
+            return self.handle_get()
         elif method == "PATCH":
-            self.handle_patch()
+            return self.handle_patch()
         elif method == "POST":
-            self.handle_post()
+            return self.handle_post()
+        else:
+            return self.make_reponse_object(HTTPStatus.NOT_FOUND, [])
```
PawelLipski: `response`

PawelLipski: Also, nit: real Github is likely to return 405 Method Not Allowed status in this case

PawelLipski (approved):
Looks ready now 🎉

git_machete/tests/functional/test_machete.py
```diff
@@ -26,6 +30,182 @@ def get_head_commit_hash() -> str:
...
+            return self.update_issue()
+        elif 'pulls' in self.parsed_url.path:
+            return self.update_pull_request()
+        else:
+            return self.make_response_object(HTTPStatus.NOT_FOUND, [])
+
+    def handle_post(self) -> "MockGithubAPIResponse":
+        assert not self.parsed_query
+        if self.parsed_url.path.endswith('issues'):
+            return self.update_issue()
```
PawelLipski, line 120: One thing, can't we just call `create_issue/pull_request` in `handle_post`? this way `if not pull_no/issue_no` could probably be removed from `update_...` methods

rifel123: Without further changes, we can't. This is because POST is used to add reviewers or assignees to pulls/issues while sending only a fraction of data which is necessary to create pull/issue, which is considered by put  `MockGihubAPIRequest` as an update.

PawelLipski: Ok ☹️ let's leave as is then

PR #215, opened 2021-09-07
Issue #206: Split current cmd.py into client.py and cli.py

### LINKED ISSUE #158: Consider migrating CLI from low-level `getopt` + custom code to `argparse`
https://docs.python.org/3/library/argparse.html

Also to be considered: https://click.palletsprojects.com/en/8.0.x/ although this would add a 3rd party dependency which I'd like to avoid for the sake of Sacred Simplicity of Distribution ™️

### LINKED ISSUE #206: Split current cmd.py into client.py and cli.py + update references to `git_machete.cmd`
client.py should contain just `MacheteClient` class (which could theoretically be used from a non-command line context as well).

cli.py should be strictly dedicated to command-line param parsing/handling etc.

BartoszSiwinski:
I'm trying to address this issue #206.

This is the scope:
- client.py should contain just MacheteClient class (which could theoretically be used from a non-command line context as well);
- cli.py should be strictly dedicated to command-line param parsing/handling etc.

Extras:
- I've also added a number of newlines to code lines so they fit (mostly) in 80-100 character range. I haven't cut all of them, since it's a quite lengthy (😆 ) process, but my goal is to add some more improvement like that with every PR, so repo is getting incrementally easier to read.
- I've added EscapeCodes class to encapsulate all these RED, YELLOW etc. in one structure.

@PawelLipski is there something else that needs to be done in here?

### DIFF
--- a/git_machete/cli.py
+++ b/git_machete/cli.py
@@ -1,0 +1,550 @@
+#!/usr/bin/env python
+
+import getopt
+import os
+import re
+import sys
+import textwrap
+from typing import Dict, List, Optional, Tuple, TypeVar
+
+from git_machete import __version__
+from git_machete import utils
+from git_machete.client import MacheteClient, allowed_directions
+from git_machete.constants import EscapeCodes
+from git_machete.docs import short_docs, long_docs
+from git_machete.exceptions import MacheteException, StopTraversal
+from git_machete.git_operations import GitContext
+from git_machete.options import CommandLineOptions
+from git_machete.utils import fmt, underline, excluding, warn
+
+T = TypeVar('T')
+
+initial_current_directory: Optional[str] = utils.get_current_directory_or_none() or os.getenv('PWD')
+
+alias_by_command: Dict[str, str] = {
+    "diff": "d",
+    "edit": "e",
+    "go": "g",
+    "log": "l",
+    "status": "s"
+}
+command_by_alias: Dict[str, str] = {v: k for k, v in alias_by_command.items()}
+
+command_groups: List[Tuple[str, List[str]]] = [
+    # 'is-managed' is mostly for scripting use and therefore skipped
+    ("General topics",
+     ["file", "help", "hooks", "version"]),
+    ("Build, display and modify the tree of branch dependencies",
+     ["add", "anno", "discover", "edit", "status"]),
+    ("List, check out and delete branches",
+     ["delete-unmanaged", "go", "list", "show"]),
+    ("Determine changes specific to the given branch",
+     ["diff", "fork-point", "log"]),
+    ("Update git history in accordance with the tree of branch dependencies",
+     ["advance", "reapply", "slide-out", "squash", "traverse", "update"])
+]
+
+
+def usage(command: str = None) -> None:
+    if command and command in command_by_alias:
+        command = command_by_alias[command]
+    if command and command in long_docs:
+        print(fmt(textwrap.dedent(long_docs[command])))
+    else:
+        print()
+        short_usage()
+        if command:
+            print(f"\nUnknown command: '{command}'")
+        print(fmt("\n<u>TL;DR tip</u>\n\n"
+                  "    Get familiar with the help for <b>format</b>, <b>edit</b>,"
+                  " <b>status</b> and <b>update</b>, in this order.\n"))
+        for hdr, cmds in command_groups:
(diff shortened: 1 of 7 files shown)


git_machete/cli.py
```diff
@@ -0,0 +1,540 @@
...
+
+def usage(command: str = None) -> None:
+    if command and command in command_by_alias:
+        command = command_by_alias[command]
+    if command and command in long_docs:
+        print(fmt(textwrap.dedent(long_docs[command])))
+    else:
+        print()
+        short_usage()
+        if command:
```
BartoszSiwinski, line 56: This is a bit modified in comparison to the content of its ancestor `cmd.py`.
Previously it was if `command and command not in long_docs`. I changed it, because in `else` itself we've already verified that it's either `command == None/''/False` or `command exists and  it is not in long_docs`. Based on this I think it's redundant to check that `command exists and is not in long doc`s. It should be sufficient to just check if it `exists`, because if `it exists and is in the long_docs `it wouldn't enter the `else` part.

git_machete/client.py
```diff
@@ -1,42 +1,44 @@
...
+    "log": "l",
+    "status": "s"
+}
 
 
-# Manipulation on definition file/tree of branches
+# Allowed parameter values for show/go command
+def allowed_directions(allow_current: bool) -> str:
+    current = "c[urrent]|" if allow_current else ""
+    return current + "d[own]|f[irst]|l[ast]|n[ext]|p[rev]|r[oot]|u[p]"
```
PawelLipski: These probably belong to cli.py too!

git_machete/client.py
```diff
@@ -551,15 +629,18 @@ def traverse(self) -> None:
...
+                # Avoid unnecessary fork point check if we already know that the
+                # branch qualifies for resetting to remote counterpart;
                 # neither rebase nor merge will be suggested in such case anyway.
                 needs_parent_sync = False
             elif self.__cli_opts.opt_merge:
-                needs_parent_sync = bool(upstream and not self.__git.is_ancestor_or_equal(upstream, branch))
+                needs_parent_sync = bool(
+                    upstream and not self.__git.is_ancestor_or_equal(upstream, branch))
             else:  # using rebase
                 needs_parent_sync = bool(upstream and not (self.__git.is_ancestor_or_equal(upstream, branch) and self.__git.get_commit_sha_by_revision(upstream) == self.fork_point(branch, use_overrides=True)))
```
PawelLipski: I'd split this one too ✂️

git_machete/cli.py
```diff
@@ -0,0 +1,540 @@
...
+T = TypeVar('T')
+
+initial_current_directory: Optional[str] = utils.get_current_directory_or_none() or os.getenv('PWD')
+
+command_by_alias: Dict[str, str] = {v: k for k, v in alias_by_command.items()}
+
+command_groups: List[Tuple[str, List[str]]] = [
+    # 'is-managed' is mostly for scripting use and therefore skipped
+    ("General topics", ["file", "help", "hooks", "version"]),
+    ("Build, display and modify the tree of branch dependencies",
```
PawelLipski, line 37: Nit: I'd merge this line (and 33/34) for clarity

git_machete/client.py
BartoszSiwinski: `alias_by_command` yes, I missed that. 

`allowed_directions` can't be put into `cli.py`, because `client.py` also uses that and I will get circular imports. It could be moved to third files, but there isn't one that would make sense from context point of view (even` utils.py` seems to be not a right place, because it is focusing on stuff unrelated to `cli`, `git` etc.). If you would like me to create a new file and put this method there let me know how you would like this file to be named.

Fact that we have this dilemma is probably a signal that files should be split more, responsibilities of particular classes and methods should be reconsidered and probably divided. For now, I would recommend leaving allowed_directions as is as it would be a massive task to redesign repo.

git_machete/cli.py
BartoszSiwinski: Instead I split this in such way that there is:
(str
 List),
(str
 List),
...
So each pair takes 2 lines in the same format and order.

git_machete/client.py
PawelLipski: Okay, let's leave as is for now, I missed that

PawelLipski (approved):

PR #219, opened 2021-09-08
get remote git url from config

rifel123:
Due to git 1.8 which does not support `get-url` function.

### DIFF
--- a/git_machete/git_operations.py
+++ b/git_machete/git_operations.py
@@ -161,7 +161,7 @@ class GitContext:
         return self.__remotes_cached
 
     def get_url_of_remote(self, remote: str) -> str:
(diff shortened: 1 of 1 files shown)


git_machete/git_operations.py
```diff
@@ -161,7 +161,7 @@ def remotes(self) -> List[str]:
         return self.__remotes_cached
 
     def get_url_of_remote(self, remote: str) -> str:
-        return self.popen_git("remote", "get-url", "--", remote).strip()
+        return self.popen_git("config", "--get", f"remote.{remote}.url").strip()
```
PawelLipski: Pls drop a comment that `git remote get-url` has only been added in git v2.5.1

PawelLipski (approved):

PR #229, opened 2021-09-13
Issue #158: Migrate CLI to argparse.

### LINKED ISSUE #158: Consider migrating CLI from low-level `getopt` + custom code to `argparse`

BartoszSiwinski:


git_machete/cli.py
```diff
@@ -99,6 +99,153 @@ def launch(orig_args: List[str]) -> None:
             f"Python {version_str} is no longer supported. Please switch to Python 3.6 or higher.\n")
         sys.exit(1)
 
+    def parse_options_argparse(in_args: List[str]) -> argparse.Namespace[str]:
```
rifel123: Remove `[str]` after `Namespace`, cause it breaks flow atm:
Screenshot 2021-09-13 at 14 50 11

git_machete/cli.py
```diff
@@ -99,6 +99,153 @@ def launch(orig_args: List[str]) -> None:
...
+        fork_point_parser.add_argument('--override-to')
+        fork_point_parser.add_argument('--override-to-inferred', action='store_true')
+        fork_point_parser.add_argument('--override-to-parent', action='store_true')
+        fork_point_parser.add_argument('--unset-override', action='store_true')
+
+        format_parser = subparsers.add_parser('format')
+
+        github_parser = subparsers.add_parser('github')
+        github_parser.add_argument('subcommand', choices=['anno-prs', 'create-pr', 'retarget-pr'])
+        github_parser.add_argument('--draft', action='store_true', default=argparse.SUPPRESS)
```
rifel123: I wonder how it should look like with `github checkout-pr N`. Probably another subparser has to be added, sth like:
```suggestion
       github_parser.add_argument('--draft', action='store_true', default=argparse.SUPPRESS)
       github_parser.add_argument('integer', metavar='N', type=int, help='an integer for the PR number')
```

git_machete/cli.py
BartoszSiwinski: As I mentioned in the chat, it's not something that is supposed to be run or used in any sense just yet. I'm still choosing the place where particular code should be, I'm still testing functionalities.

git_machete/cli.py
BartoszSiwinski: Another argument, not a subparser, but yes, overall github subcommand choices would have to be expanded and argument 'N' would have to be added.

git_machete/cli.py
```diff
@@ -75,38 +78,172 @@ def usage(command: str = None) -> None:
...
+def short_usage() -> str:
+    return (fmt("<b>Usage: git machete [--debug] [-h] [-v|--verbose] [--version] "
+                "<command> [command-specific options] [command-specific argument]</b>"))
 
 
 def version() -> None:
     print(f"git-machete version {__version__}")
 
 
+def create_cli_parser() -> argparse.ArgumentParser:
```
PawelLipski, line 115: Out of curiosity, is generation of shell completion scripts also supported by argparse? (Not that we strictly need that now since we've got hand-written bash&zsh completion)

git_machete/cli.py
```diff
@@ -75,38 +78,172 @@ def usage(command: str = None) -> None:
...
+    delete_unmanaged_parser.add_argument(
+        '-y', '--yes', action='store_true', default=argparse.SUPPRESS)
+
+    diff_full_parser = subparsers.add_parser('diff', aliases=['d'])
+    diff_full_parser.add_argument('branch', nargs='?')
+    diff_full_parser.add_argument(
+        '-s', '--stat', action='store_true', default=argparse.SUPPRESS)
+
+    discover_parser = subparsers.add_parser('discover', argument_default=argparse.SUPPRESS)
+    discover_parser.add_argument('-C', '--check-out-since')
```
PawelLipski: Hmmm how does argparse know that a parameter is required after `-C` and `-r` but not after `-l` or `-y`? or is it yet to be implemented?

git_machete/cli.py
```diff
@@ -75,38 +78,172 @@ def usage(command: str = None) -> None:
...
+    go_parser = subparsers.add_parser('go', aliases=['g'])
+    go_parser.add_argument(
+        'direction',
+        choices=['d', 'down', 'f', 'first', 'l', 'last', 'n', 'next',
+                 'p', 'prev', 'r', 'root', 'u', 'up']
+    )
+
+    help_parser = subparsers.add_parser('help')
+    help_parser.add_argument(
+        '--command', choices=list(short_docs.keys()), default=argparse.SUPPRESS)
```
PawelLipski: `--command` or just `command`?

PawelLipski: Coz `--command` looks like a flag 🤔

git_machete/cli.py
```diff
@@ -75,38 +78,172 @@ def usage(command: str = None) -> None:
...
+    list_parser.add_argument(
+        'category',
+        choices=[
+            'addable', 'managed', 'slidable', 'slidable-after', 'unmanaged',
+            'with-overridden-fork-point']
+    )
+    list_parser.add_argument('branch', nargs='?', default=argparse.SUPPRESS)
+
+    log_parser = subparsers.add_parser('log', aliases=['l'])
+    log_parser.add_argument('branch', nargs='?', default=argparse.SUPPRESS)
```
PawelLipski, line 272: Hmmm what does `nargs='?'` mean? is it "0 or more arguments"? or "0 or 1 argument"?

git_machete/cli.py
```diff
@@ -75,38 +78,172 @@ def usage(command: str = None) -> None:
...
+    traverse_parser = subparsers.add_parser('traverse', argument_default=argparse.SUPPRESS)
+    traverse_parser.add_argument('-F', '--fetch', action='store_true')
+    traverse_parser.add_argument('-l', '--list-commits', action='store_true')
+    traverse_parser.add_argument('-M', '--merge', action='store_true')
+    traverse_parser.add_argument('-n', action='store_true')
+    traverse_parser.add_argument('--no-edit-merge', action='store_true')
+    traverse_parser.add_argument('--no-interactive-rebase', action='store_true')
+    traverse_parser.add_argument('--no-detect-squash-merges', action='store_true')
+    traverse_parser.add_argument('--push', action='store_true')
+    traverse_parser.add_argument('--no-push', action='store_true')
```
PawelLipski, line 330: Hmm can it be sth like `store_false` but for `--push` flag? since that's basically the current logic

git_machete/cli.py
BartoszSiwinski: In `add_argument` method you can pass argument `action='store_true'`, which causes `parser` to put `my_optional_argument=True` into it's `argparse.Namespace` if added argument `is in` the parsed list. You can also do `'store_false'`.

git_machete/cli.py
BartoszSiwinski: I do have to check it.

git_machete/cli.py
BartoszSiwinski: 0 or 1, but if there is 0 use default. https://docs.python.org/3/library/argparse.html#nargs

git_machete/cli.py
BartoszSiwinski: Few lines up, there is for whole subparser default=argparse.SUPPRESS which causes particular argument not to be even in the argparse.Namespace if it was not in the parsed arguments. This way if I will pass this optional argument it will show as push=True, otherwise it will not show at all.

If I implement it with 'store_false' and pass this argument it will show as push=False, so it's opposite of what we want. If I won't pass it it will show as push=True.

https://docs.python.org/3/library/argparse.html#action

git_machete/cli.py
BartoszSiwinski: It doesn't seems so. There are some random libraries offering it, but not argparse itself from a quick research.

git_machete/cli.py
PawelLipski: Ok, looks that `+`, `?` and  `*` are used consistently with regex syntax here

git_machete/cli.py
```diff
@@ -142,7 +142,7 @@ def create_cli_parser() -> argparse.ArgumentParser:
 
     subparsers.add_parser('format')
 
-    github_parser = subparsers.add_parser('github', default=argparse.SUPPRESS)
+    github_parser = subparsers.add_parser('github', argument_default=argparse.SUPPRESS)
```
PawelLipski: Wooow what's the difference here? 😯

BartoszSiwinski: Top one doesn't work 😄  at all . 
`default` is a parameter of a `add_argument` method, whereas `argument_default` is an argument of a `add_parser` method. This change is basically a fix for an error I've made - there is no argument `default` in `add_parser` method.

git_machete/cli.py
```diff
@@ -354,192 +505,149 @@ def check_required_param(in_args: List[str], allowed_values_string: str) -> str:
...
-        else:
-            short_usage()
-            raise MacheteException(
-                f"\nUnknown command: `{cmd}`. Use `git machete help` to list possible commands")
 
-    except getopt.GetoptError as e:
-        short_usage()
+    except (argparse.ArgumentError, argparse.ArgumentTypeError) as e:
+        print(short_usage())
         sys.stderr.write(f"\n{e}\n")
```
PawelLipski, line 651: While you're here... pls remove the newline here, it usually look weird:
```suggestion
        sys.stderr.write(f"{e}\n")
```

git_machete/cli.py
```diff
@@ -354,192 +505,149 @@ def check_required_param(in_args: List[str], allowed_values_string: str) -> str:
...
+            if cli_opts.opt_start_from not in {"here", "root", "first-root"}:
                 raise MacheteException(
-                    "Invalid argument for `--start-from`. Valid arguments: `here|root|first-root`.")
+                    "Invalid argument for `--start-from`. "
+                    "Valid arguments: `here|root|first-root`.")
             if cli_opts.opt_return_to not in ("here", "nearest-remaining", "stay"):
                 raise MacheteException(
-                    "Invalid argument for `--return-to`. Valid arguments: here|nearest-remaining|stay.")
+                    "Invalid argument for `--return-to`. "
+                    "Valid arguments: here|nearest-remaining|stay.")
```
PawelLipski: Nit:
```suggestion
                    "Valid arguments: `here|nearest-remaining|stay`.")
```

PawelLipski: (oh, it was like that originally :trollface:)

git_machete/cli.py
```diff
@@ -75,274 +77,423 @@ def usage(command: str = None) -> None:
...
-def main() -> None:
-    launch(sys.argv[1:])
-
-
-def launch(orig_args: List[str]) -> None:
-    cli_opts = git_machete.options.CommandLineOptions()
-    git = GitContext(cli_opts)
-
-    if sys.version_info.major == 2 or (sys.version_info.major == 3 and sys.version_info.minor < 6):
+class MacheteHelpAction(argparse.Action):
```
PawelLipski, line 87: AFAICS `-h` still awkwardly shows up in error messages

```
$ git machete  show foo
usage: git-machete show [-h]
                        {c,current,d,down,f,first,l,last,n,next,p,prev,r,root,u,up}
                        [branch]
git-machete show: error: argument direction: invalid choice: 'foo' (choose from 'c', 'current', 'd', 'down', 'f', 'first', 'l', 'last', 'n', 'next', 'p', 'prev', 'r', 'root', 'u', 'up')
```

Any chance to get rid of that?.... can be a separate issue ofc as I assume it either requires low-level hacking or is non-trivial to do properly

git_machete/cli.py
```diff
@@ -75,274 +77,423 @@ def usage(command: str = None) -> None:
...
-                    with open(machete_client.definition_file_path, "a"):
-                        pass
-                elif os.path.isdir(machete_client.definition_file_path):
-                    # Extremely unlikely case, basically checking if anybody tampered with the repository.
-                    raise MacheteException(
-                        f"{machete_client.definition_file_path} is a directory rather than a regular file, aborting")
+
+        cli_parser = create_cli_parser()
+        parsed_cli = cli_parser.parse_args(orig_args)
+        parsed_cli_as_dict = vars(parsed_cli)
```
PawelLipski: Nit: pls explicitly type these vars for clarity if easily doable

git_machete/cli.py
```diff
@@ -75,274 +77,423 @@ def usage(command: str = None) -> None:
...
+        '-s', '--stat', action='store_true', default=argparse.SUPPRESS)
+    diff_full_parser.add_argument('-h', '--help', action=MacheteHelpAction)
+
+    discover_parser = subparsers.add_parser(
+        'discover', argument_default=argparse.SUPPRESS, add_help=False)
+    discover_parser.add_argument('-C', '--check-out-since')
+    discover_parser.add_argument('-l', '--list-commits', action='store_true')
+    discover_parser.add_argument('-r', '--roots')
+    discover_parser.add_argument('-y', '--yes', action='store_true')
+    discover_parser.add_argument('-h', '--help', action=MacheteHelpAction)
```
PawelLipski: Nit: is it possible to add `--help` argument to all subparses in one go/via some loop at the very end?

git_machete/cli.py
```diff
@@ -75,274 +77,423 @@ def usage(command: str = None) -> None:
...
+
+    diff_full_parser = subparsers.add_parser('diff', aliases=['d'], add_help=False)
+    diff_full_parser.add_argument('branch', nargs='?')
+    diff_full_parser.add_argument(
+        '-s', '--stat', action='store_true', default=argparse.SUPPRESS)
+    diff_full_parser.add_argument('-h', '--help', action=MacheteHelpAction)
+
+    discover_parser = subparsers.add_parser(
+        'discover', argument_default=argparse.SUPPRESS, add_help=False)
+    discover_parser.add_argument('-C', '--check-out-since')
```
PawelLipski: `checked-out-since` ‼️

git_machete/cli.py
```diff
@@ -75,274 +77,423 @@ def usage(command: str = None) -> None:
...
+
+    format_parser = subparsers.add_parser('format', add_help=False)
+    format_parser.add_argument('-h', '--help', action=MacheteHelpAction)
+
+    github_parser = subparsers.add_parser(
+        'github', argument_default=argparse.SUPPRESS, add_help=False)
+    github_parser.add_argument(
+        'subcommand', choices=['anno-prs', 'checkout-prs', 'create-pr', 'retarget-pr'])
+    github_parser.add_argument('pr_no', nargs='?')
+    github_parser.add_argument('--draft', action='store_true')
```
PawelLipski, line 225: Nit: `git machete github anno-prs --draft` now passes... it theoretically shouldn't, `--draft` only makes sense for `create-pr`

git_machete/cli.py
```diff
@@ -75,274 +77,423 @@ def usage(command: str = None) -> None:
...
+            cli_opts.opt_down_fork_point = arg
+        elif opt == "debug":
+            git_machete.options.CommandLineOptions.opt_debug = True
+        elif opt == "draft":
+            cli_opts.opt_draft = True
+        elif opt == "fetch":
+            cli_opts.opt_fetch = True
+        elif opt == "fork_point":
+            cli_opts.opt_fork_point = arg
+        elif opt == "sync_github_prs":
```
PawelLipski: Nit: pls keep abc order (it used to be abc order but including short options)

git_machete/cli.py
```diff
@@ -75,274 +77,423 @@ def usage(command: str = None) -> None:
...
+        elif opt == "return_to":
+            cli_opts.opt_return_to = arg
+        elif opt == "stat":
+            cli_opts.opt_stat = True
+        elif opt == "start_from":
+            cli_opts.opt_start_from = arg
+        elif opt == "unset_override":
+            cli_opts.opt_unset_override = True
+        elif opt == "verbose":
+            git_machete.options.CommandLineOptions.opt_verbose = True
```
PawelLipski: Hmmm does it still need to be a class attribute? given that you call `set_utils_variables`, which assigns `opt_verbose`/`opt_debug` to global vars anyway

git_machete/cli.py
```diff
@@ -75,274 +77,423 @@ def usage(command: str = None) -> None:
...
+    diff_full_parser.add_argument('-h', '--help', action=MacheteHelpAction)
+
+    discover_parser = subparsers.add_parser(
+        'discover', argument_default=argparse.SUPPRESS, add_help=False)
+    discover_parser.add_argument('-C', '--check-out-since')
+    discover_parser.add_argument('-l', '--list-commits', action='store_true')
+    discover_parser.add_argument('-r', '--roots')
+    discover_parser.add_argument('-y', '--yes', action='store_true')
+    discover_parser.add_argument('-h', '--help', action=MacheteHelpAction)
+
```
PawelLipski, line 185: Not-so-nit: same goes for `-v` and other common flags, as for now they can't be passed at the end:

```
$ git machete github anno-prs --debug
usage: git-machete [--debug] [-h] [--version] [-v]
                   {add,advance,anno,delete-unmanaged,diff,d,discover,edit,e,file,fork-point,format,github,go,g,help,hooks,is-managed,list,log,l,reapply,show,slide-out,squash,status,s,traverse,update,version}
                   ...
git-machete: error: unrecognized arguments: --debug
``

File: git_machete/cli.py (hunk 11 above)
Comment (BartoszSiwinski) [reply git_machete/cli.py:87]: There few things we can do and I will try to present their consequences in the following format show_parser definition -> result when you call `git machete show foo` .

Option A (leave as is):
`show_parser = subparsers.add_parser('show', add_help=False)`
result:
```
➜  git-machete git:(improve/parser) ✗ git machete show foo   
usage: git-machete show [-h] {c,current,d,down,f,first,l,last,n,next,p,prev,r,root,u,up} [branch]
git-machete show: error: argument direction: invalid choice: 'foo' (choose from 'c', 'current', 'd', 'down', 'f', 'first', 'l', 'last', 'n', 'next', 'p', 'prev', 'r', 'root', 'u', 'up')
```

Option B (we remove usage from displaying):
`show_parser = subparsers.add_parser('show', usage=argparse.SUPPRESS, add_help=False)`
result:
```
➜  git-machete git:(improve/parser) ✗ git machete show foo
git-machete show: error: argument direction: invalid choice: 'foo' (choose from 'c', 'current', 'd', 'down', 'f', 'first', 'l', 'last', 'n', 'next', 'p', 'prev', 'r', 'root', 'u', 'up')
```

Option C (we change usage into long_docs(command), by calling get_help_description/usage(command))
Please notice ugle double usage as the begging of the message. This can be hidden in some cases, but for stuff liek `format` it won't work:
`show_parser = subparsers.add_parser('show', usage=get_help_description('show'), add_help=False)`
result:
```
➜  git-machete git:(improve/parser) ✗ git machete show foo
usage: 
Usage: git machete show <direction> [<branch>]
where <direction> is one of: c[urrent], d[own], f[irst], l[ast], n[ext], p[rev], r[oot], u[p]
displayed relative to given <branch>, or the current checked out branch if <branch> is unspecified.

Outputs name of the branch (or possibly multiple branches, in case of down) that is:

* current: the current branch; exits with a non-zero status if none (detached HEAD)
* down:    the direct children/downstream branch of the given branch.
* first:   the first downstream of the root branch of the given branch (like root followed by next), or the root branch itself if the root has no downstream branches.
* last:    the last branch in the definition file that has the same root as the given branch; can be the root branch itself if the root has no downstream branches.
* next:    the direct successor of the given branch in the definition file.
* prev:    the direct predecessor of the given branch in the definition file.
* root:    the root of the tree where the given branch is located. Note: this will typically be something like develop or master, since all branches are usually meant to be ultimately merged to one of those.
* up:      the direct parent/upstream branch of the given branch.
git-machete show: error: argument direction: invalid choice: 'foo' (choose from 'c', 'current', 'd', 'down', 'f', 'first', 'l', 'last', 'n', 'next', 'p', 'prev', 'r', 'root', 'u', 'up')
```

If any of these options is satisfying for you, please let me know. I recommend option B, so silence it and if someone wants a help for it they should execute `git machete <command> --help`.

Comment (PawelLipski) [reply git_machete/cli.py:87]: Ok! let's go for B

>  if someone wants a help for it they should execute git machete <command> --help.

It'd be probably even more consistent since currently:
* for `github`, `go`, `list`, `show` &mdash; if no param is provided, help is displayed
* for all other commands &mdash; invocation with no params is a valid one

So once we go for B, `help` will be provided either from `git machete help <command>` or `git machete <command> --help` and never via `git machete <command>`

File: git_machete/cli.py (hunk 18 above)
Comment (BartoszSiwinski) [reply git_machete/cli.py:185]: We may have a problem, because this is not be possible -> https://docs.python.org/3/library/argparse.html#intermixed-parsing 
I'm using subparsers and mutually exclusive groups. Mutually exclusive groups could be probably dropped and replaced with something primitive, but dropping subparsers invalidates the idea of using argparse.

I will try few more things with that issue and see if it helps.

Comment (PawelLipski) [reply git_machete/cli.py:185]: Hmmm okay but even without this intermixed parsing... how about just adding `<subcommand>_parser.add_argument('-v', '--verbose', action='store_true')` to every single subparser (probably in some loop), similarly as it's done for `--help`?

Comment (BartoszSiwinski) [reply git_machete/cli.py:185]: I found a counterintuitive hack. So there will be now a new parser, with just these common optional args like -v -h --debug that will be kind of joined with all other parsers (kind of like inheritance). It will work like that:

```
➜  git-machete git:(improve/parser) git machete --debug show up
>>> git symbolic-ref --quiet HEAD
<stdout>:
refs/heads/improve/parser

enchance/checkout-pr
➜  git-machete git:(improve/parser) git machete show up --debug
>>> git symbolic-ref --quiet HEAD
<stdout>:
refs/heads/improve/parser

enchance/checkout-pr
➜  git-machete git:(improve/parser) git machete show --debug up        
>>> git symbolic-ref --quiet HEAD
<stdout>:
refs/heads/improve/parser
```

File: git_machete/cli.py (hunk 13 above)
Comment (BartoszSiwinski) [reply git_machete/cli.py]: I've done it via adding new common args parser and inheriting args from it. It should work now.

File: git_machete/cli.py (hunk 18 above)
Comment (BartoszSiwinski) [reply git_machete/cli.py:185]: > Hmmm okay but even without this intermixed parsing... how about just adding `<subcommand>_parser.add_argument('-v', '--verbose', action='store_true')` to every single subparser (probably in some loop), similarly as it's done for `--help`?

Sorry, my gh was not refreshed and I missed that comment. Responding: it's no longer needed. This intermixed stuff is confusing, but seems to be not a problem, when there is a separate generic_optional_args_parser, so no need to add help or version to each subparser separately.

**Comment by PawelLipski** [comment]
There's quite a lot of stray commits from Piotr's PRs on this branch (when comparing to develop) 🤔

**Comment by BartoszSiwinski** [comment]
> There's quite a lot of stray commits from Piotr's PRs on this branch (when comparing to develop) 🤔

I think I messed up rebase. I rebased to develop without first rebasing enchance/checkout-pr so it blew up.

**Comment by PawelLipski** [comment]
Yeah... if you could `git rebase --onto` or `git machete update --fork-point`

File: git_machete/cli.py [hunk 19]
```diff
@@ -45,304 +44,517 @@
...
+        'category',
+        choices=[
+            'addable', 'managed', 'slidable', 'slidable-after', 'unmanaged',
+            'with-overridden-fork-point']
+    )
+    list_parser.add_argument('branch', nargs='?', default=argparse.SUPPRESS)
+
+    log_parser = subparsers.add_parser(
+        'log',
+        aliases=['l'],
```
PawelLipski, line 268: Whoops `git machete l` just does nothing and exits 😯 same for `git machete e`, `git machete g <something>`, `git machete s`

git_machete/cli.py
```diff
@@ -45,304 +44,517 @@
...
+        """[1:]))
+    return usage_str
 
 
-def short_usage() -> None:
-    print(
-        fmt("<b>Usage: git machete [--debug] [-h] [-v|--verbose] [--version] "
-            "<command> [command-specific options] [command-specific argument]</b>"))
+def get_short_general_usage() -> str:
+    return (fmt("<b>Usage: git machete [--debug] [-h] [-v|--verbose] [--version] "
```
PawelLipski, line 79: Nit: in error messages, the program name is `git-machete`, can be changed to `git machete`:
```
$ git machete a
usage: git-machete [--debug] [-h] [--version] [-v]
```

PawelLipski: Although not necessary if problematic, since both `git machete` and `git-machete` are both valid invocations; the former is just nicer-looking xD

git_machete/cli.py
```diff
@@ -354,192 +566,154 @@ def check_required_param(in_args: List[str], allowed_values_string: str) -> str:
...
-                raise MacheteException(f"`git machete github` expects argument(s): {github_allowed_subcommands}")
-
-            param = args[0]
+            github_subcommand = parsed_cli.subcommand
             machete_client.read_definition_file()
-            if param == "anno-prs":
+            if (github_subcommand in {'anno-prs', 'checkout-prs', 'retarget-pr'} and
+                    'draft' in parsed_cli):
+                raise MacheteException(
+                    "--draft option is NOT valid with 'anno-prs', 'checkout-prs',"
```
PawelLipski: Also, `pr_no` could be banned for `'anno-prs', 'checkout-prs', 'retarget-pr'`, as for now it's just ignored AFAICS:

```
$ git machete github anno-prs 12345
<OK>
```

PawelLipski (commented):
Looks almost ready to approve 🥳

PawelLipski (approved):
🎉

PR #230, opened 2021-09-13
add check if head and base branch are equal

### LINKED ISSUE #193: Unnecessary attempt to create pull request when child branch is equal to parent branch
This attempt end up with error, cause there is no difference between those two branches.
there is check for this issue:
Screenshot 2021-08-26 at 10 56 33

but it passes when current branch has no new commit. It's because `filtered_reflog` method excludes every branch creation from reflog.

Maybe there is need to make dedicated check in this case?

rifel123:


git_machete/client.py
```diff
@@ -1906,8 +1906,9 @@ def __sync_before_creating_pr(self) -> None:
         if current_branch not in self.managed_branches:
             self.add(current_branch)
 
-        if self.__is_merged_to_upstream(current_branch):
-            return
+        if self.__git.is_ancestor_or_equal(current_branch, self.up_branch.get(current_branch), equal_only=True):
+            raise MacheteException(
+                'Head Branch for this pull request is equal to base branch.\nCannot create pull request.')
```
PawelLipski, line 1911: Whoops that's not exactly the same as the `if` condition above :/

git_machete/client.py
```diff
@@ -1906,8 +1906,9 @@ def __sync_before_creating_pr(self) -> None:
         if current_branch not in self.managed_branches:
             self.add(current_branch)
 
-        if self.__is_merged_to_upstream(current_branch):
-            return
+        if self.__git.is_ancestor_or_equal(current_branch, self.up_branch.get(current_branch), equal_only=True):
```
PawelLipski: Hmmm okay but if current branch points to an **ancestor** commit of parent branch... will Github agree to open a PR from current branch to parent branch?

git_machete/client.py
```suggestion
                'Head branch for this pull request is equal to base branch.\nCannot create pull request.')
```

git_machete/client.py
rifel123: Yes, you were right, Github doesn't allow to open pr under such circumstances. So, i changed the condition now, opening a PR is allowed only if head branch has at least one commit which is not included in base branch.

git_machete/client.py
```diff
@@ -1906,8 +1906,9 @@ def __sync_before_creating_pr(self) -> None:
         if current_branch not in self.managed_branches:
             self.add(current_branch)
 
-        if self.__is_merged_to_upstream(current_branch):
-            return
+        if self.__git.is_head_in_base_branch(current_branch, self.up_branch.get(current_branch)):
```
PawelLipski: Pls note that this check can be implemented with just `is_ancestor_or_equal`... no new code in git_operations.py needed 😅

PawelLipski: Lmk if a further clarification is needed on why _all commits from X are found in Y_ is equivalent to _X is ancestor of or equal to Y_

git_machete/client.py
```diff
@@ -1906,8 +1906,9 @@ def __sync_before_creating_pr(self) -> None:
         if current_branch not in self.managed_branches:
             self.add(current_branch)
 
-        if self.__is_merged_to_upstream(current_branch):
-            return
+        if self.__git.is_ancestor_or_equal(current_branch, self.up_branch.get(current_branch)):
```
PawelLipski: Pls add a test that would cover this scenario

git_machete/client.py
```diff
@@ -1906,8 +1906,9 @@ def __sync_before_creating_pr(self) -> None:
...
+            raise MacheteException(
+                'All commits in head branch for this pull request are already included in base branch.\nCannot create pull request.')
```
PawelLipski: I'd include head and base branch names explicitly in the error message for clarity

git_machete/client.py
```diff
@@ -1906,9 +1906,10 @@ def __sync_before_creating_pr(self) -> None:
         if current_branch not in self.managed_branches:
             self.add(current_branch)
 
-        if self.__git.is_ancestor_or_equal(current_branch, self.up_branch.get(current_branch)):
+        up_branch = self.up_branch.get(current_branch)
```
PawelLipski: What if `up_branch` is `None` (i.e. `current_branch` is a root branch)? The check whether parent branch exists happens in `create_github_pr` only after `__sync_before_creating_pr` is called. If there's indeed a bug here, pls add a regression test for this case as well (opening a PR for a root branch).

git_machete/client.py
```diff
@@ -1906,8 +1901,15 @@ def __sync_before_creating_pr(self) -> None:
...
-            return
+        up_branch: Optional[str] = self.up_branch.get(current_branch)
+        if not up_branch:
+            raise MacheteException(
+                f'Branch `{current_branch}` does not have a parent branch (it is a root), '
+                'base branch for the PR cannot be established.')
+
+        if self.__git.is_ancestor_or_equal(current_branch, up_branch):
+            raise MacheteException(
+                f'All commits in `{current_branch}` branch  are already included in `{up_branch}` branch.\nCannot create pull request.')
```
PawelLipski, line 1908:
```suggestion
                f'All commits in `{current_branch}` branch are already included in `{up_branch}` branch.\nCannot create pull request.')
```

git_machete/client.py
```diff
@@ -1768,11 +1768,6 @@ def create_github_pr(self, head: str, draft: bool) -> None:
             return
```
PawelLipski, line 1765: Hmmm why is the `except MacheteException as e:` needed above? Shouldn't a `MacheteException` coming from `__sync_before_creating_pr` be caught in top-level handler (`def launch`) as any other `MacheteException`?

git_machete/tests/functional/test_machete.py
```diff
@@ -1790,10 +1790,23 @@ def test_github_create_pr(self) -> None:
             .push()
         )
         self.launch_command('discover')
-        self.assert_command(['github', 'create-pr'], "All commits in `testing/endpoints` branch  are already included in `develop` branch.\nCannot create pull request.\n", strip_indentation=False)
+
+        machete_client = MacheteClient(cli_opts, git)
+        machete_client.read_definition_file()
+        expected_error_message = "All commits in `testing/endpoints` branch  are already included in `develop` branch.\nCannot create pull request."
```
PawelLipski, line 1823: Oh I see why that `print(e)` was needed... anyway, the new approach is better (more consistent with how exceptions are handled everywhere else)

PawelLipski (approved):

PR #241, opened 2021-09-20
Separate mypy checks into own tox env

### LINKED ISSUE #196: No way to run mypy checks without running tests as well
There should be a separate env in tox.ini for mypy, just as it is for pep8 checks

ronnie-llamado:
Fixes #196 

Moved mypy check into own tox env. Allows mypy checks to be run with ```tox -e mypy```.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,5 +1,5 @@
 [tox]
-envlist = pep8,py36,py37,py38,py39,coverage,docs
+envlist = pep8,py36,py37,py38,py39,coverage,docs,mypy
 minversion = 2.3.2
 skipsdist = True
 
@@ -13,9 +13,7 @@ install_command =
 deps =
   -r{toxinidir}/requirements.txt
   -r{toxinidir}/test-requirements.txt
-whitelist_externals = mypy
 commands =
-  mypy --config-file mypy.ini git_machete
   stestr run {posargs}
 
 [testenv:pep8]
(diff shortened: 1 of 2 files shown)


tox.ini
```diff
@@ -50,3 +48,11 @@ deps =
   sphinx-rtd-theme==0.5.2
 commands =
   sphinx-build -b html docs/source docs/html
+
+[testenv:mypy]
```
PawelLipski, line 52: Tested locally, works as expected :)

Please also update `.circleci/config.yml` so that mypy check is also executed under CI!

PawelLipski: See https://app.circleci.com/pipelines/github/VirtusLab/git-machete (log in with Github)

ronnie-llamado: Added to CI! 

A little more rework was needed in `tox.ini` to handle running `mypy` on all supported versions (for local testing convenience) or a specific version (for CI tests).

tox.ini
```diff
@@ -1,5 +1,5 @@
 [tox]
-envlist = pep8,py36,py37,py38,py39,coverage,docs
+envlist = pep8,mypy-py{36,37,38,39},py{36,37,38,39},coverage,docs
```
PawelLipski, line 2: Nice, didn't now shell-like brace expansion works in .ini files 😯

PawelLipski (approved):

Issue #242, opened 2021-09-20
`git machete github checkout-prs` cannot check out branches from yet-unadded forks


PawelLipski:
`Creating out of develop` is the problem here :/

```
➜ 21:31 ~/git-machete improve/parser PR #229 (BartoszSiwinski) $ git machete github checkout-prs 241
Fetching origin...
remote: Enumerating objects: 15, done.
remote: Counting objects: 100% (15/15), done.
remote: Compressing objects: 100% (6/6), done.
remote: Total 12 (delta 9), reused 9 (delta 6), pack-reused 0
Unpacking objects: 100% (12/12), 1.32 KiB | 79.00 KiB/s, done.
From github.com:VirtusLab/git-machete
   f88bbc7..8a8b25b  improve/parser -> origin/improve/parser
A local branch feature/add-mypy-env-to-tox does not exist. Creating out of develop
Switched to a new branch 'feature/add-mypy-env-to-tox'
Added branch feature/add-mypy-env-to-tox onto develop
Annotating feature/add-mypy-env-to-tox as PR #241 (ronnie-llamado)
Pull request #241 checked out at local branch feature/add-mypy-env-to-tox
```

PawelLipski:
Maybe checking out the `refs/pulls` remote branches wasn't that bad idea after all 🤔 to be weighed against the alternative (just adding the fork as yet another remote)

rifel123:
Adding forked repo as another remote, may end up having a lot of remotes, which probably leads to  much longer execution time when `fetch` is done for every remote, like it is in `traverse` for example. On the other hand it might be difficult to determine base branch for `refs/pull`. Maybe add branch taken with `refs/pull` as new root?

PawelLipski:
Huh good question :goncern: Actually, for clarity: _ it might be difficult to determine base branch for `refs/pull`_ what do you mean specifically? 🤔

rifel123:
Like for PR #73 there would no base ATM, same thing may happen for closed PRs from forks.

PawelLipski:
Aww okay, the same problem you've mentioned on PR... I think it's unrelated to either handling forks or even whether we use `refs/pulls` or regular remote branches, right? I think it merits a separate issue, in fact

rifel123:
closed with #254

PR #248, opened 2021-09-23
Allow to checkout closed pull requests

### LINKED ISSUE #227: Check first if given pull request is closed in `github checkout-pr`
Github Api returns all opened pull requests after `GET` on [pulls endpoint](https://docs.github.com/en/rest/reference/pulls).

During `github checkout-pr` on pull request which is already closed, git-machete ends up with message:
```
$ git machete github checkout-pr 204

PR number 204 is not found in repository git-machete
```
Scope of this issue is to make distinction in the error message when pull request is closed or never created.

rifel123:
Closes #227 

Unfortunately Github API does not provide `merged` state, there is only `open` or `closed`.

### DIFF
--- a/git_machete/client.py
+++ b/git_machete/client.py
@@ -1684,22 +1684,21 @@ class MacheteClient:
         debug('checkout_github_pr()', f'organization is {org}, repository is {repo}')
         print(f"Fetching {remote}...")
         self.__git.fetch_remote(remote)
-        if self.__git.remotes():
-            self.flush_caches()
+        self.flush_caches()
 
-        pr = get_pull_request_by_number_or_none(str(pr_no), org, repo)
+        pr = get_pull_request_by_number_or_none(pr_no, org, repo)
         if not pr:
             raise MacheteException(f"PR #{pr_no} is not found in repository `{org}/{repo}`")
         if '/'.join([remote, pr.head]) not in self.__git.get_remote_branches():
-            raise MacheteException(f"Could not checkout PR #{pr_no} because it's head branch `{pr.head}` is already deleted from `{remote}`.")
+            raise MacheteException(f"Could not check out PR #{pr_no} because its head branch `{pr.head}` is already deleted from `{remote}`.")
         if pr.state == 'closed':
             warn(f'Pull request #{pr_no} is already closed.')
         debug('checkout_github_pr()', f'found {pr}')
 
-        all_prs: List[GitHubPullRequest] = derive_pull_requests(org, repo)
+        all_open_prs: List[GitHubPullRequest] = derive_pull_requests(org, repo)
 
         self.__cli_opts.opt_yes = True  # TODO (#161): pass only needed options to methods
-        path: List[str] = self.__get_path_from_pr_chain(pr, all_prs)
+        path: List[str] = self.__get_path_from_pr_chain(pr, all_open_prs)
         reversed_path: List[str] = path[::-1]  # need to add from root downwards
         for index, branch in enumerate(reversed_path):
             if branch not in self.managed_branches:
@@ -1715,16 +1714,16 @@ class MacheteClient:
         current_user: Optional[str] = git_machete.github.derive_current_user_login()
         debug('checkout_github_pr()',
               'Current GitHub user is ' + (current_user or '<none>'))
-        self.__sync_annotations_to_definition_file(all_prs, current_user)
+        self.__sync_annotations_to_definition_file(all_open_prs, current_user)
 
         self.__git.checkout(pr.head)
         print(fmt(f"Pull request `#{pr.number}` checked out at local branch `{pr.head}`"))
 
-    def __get_path_from_pr_chain(self, current_pr: GitHubPullRequest, all_prs: List[GitHubPullRequest]) -> List[str]:
+    def __get_path_from_pr_chain(self, current_pr: GitHubPullRequest, all_open_prs: List[GitHubPullRequest]) -> List[str]:
         path: List[str] = [current_pr.head]
         while current_pr:
             path.append(current_pr.base)
-            current_pr = utils.find_or_none(lambda x: x.head == current_pr.base, all_prs)
+            current_pr = utils.find_or_none(lambda x: x.head == current_pr.base, all_open_prs)
         return path
 
     def retarget_github_pr(self, head: str) -> None:
--- a/git_machete/client.py
+++ b/git_machete/client.py
@@ -1688,7 +1688,7 @@ class MacheteClient:
             raise MacheteException(f"PR #{pr_no} is not found in repository `{org}/{repo}`")
 
         if pr.state == 'closed':
-            warn(f'Pull request #{str(pr_no)} is already closed.')
+            warn(f'Pull request #{pr_no} is already closed.')
         debug('checkout_github_pr()', f'found {pr}')
 
         all_prs: List[GitHubPullRequest] = derive_pull_requests(org, repo)
--- a/git_machete/client.py
+++ b/git_machete/client.py
@@ -18,7 +18,7 @@ from git_machete.git_operations import GitContext
 from git_machete.github import (
     add_assignees_to_pull_request, add_reviewers_to_pull_request,
     create_pull_request, derive_pull_request_by_head, derive_pull_requests,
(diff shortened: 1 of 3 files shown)


git_machete/client.py
```diff
@@ -1683,13 +1683,15 @@ def checkout_github_prs(self, pr_no: int) -> None:
...
-
+        pr = get_pull_request_by_number(str(pr_no), org, repo)
         if not pr:
             raise MacheteException(f"PR #{pr_no} is not found in repository `{org}/{repo}`")
+
+        if pr.state == 'closed':
+            warn(f'Pull request #{pr_no} is already closed.')
         debug('checkout_github_pr()', f'found {pr}')
 
+        all_prs: List[GitHubPullRequest] = derive_pull_requests(org, repo)
```
PawelLipski: Do `all_prs` include closed PRs as well? 🤔

git_machete/client.py
```diff
@@ -1683,13 +1683,15 @@ def checkout_github_prs(self, pr_no: int) -> None:
         remote, (org, repo) = self.__derive_remote_and_github_org_and_repo()
         debug('checkout_github_pr()', f'organization is {org}, repository is {repo}')
 
-        all_prs: List[GitHubPullRequest] = derive_pull_requests(org, repo)
-        pr = utils.find_or_none(lambda pr: pr.number == pr_no, all_prs)
-
+        pr = get_pull_request_by_number(str(pr_no), org, repo)
```
PawelLipski: Whoops try to check out a PR which is not only closed but also the branch is gone from the remote, e.g. `git machete github checkout-pr 247` ☹️

PawelLipski: I'd reconsider if an attempt to check out the PR is really a good approach, probably an error is a better thing to do... pls research pros and cons

PawelLipski: I know there was a loose suggestion on my part on the issue to allow checking out closed PRs and just display a warning... but as usual with new features in any project, the role of the implementer is to capture caveats that might require the idea to change

git_machete/github.py
```diff
@@ -136,6 +137,9 @@ def __fire_github_api_request(method: str, path: str, token: Optional[str], requ
                     first_line + f'You might not have the required permissions for this repository. '
                                  f'Provide a GitHub API token with `repo` access via <b>{GITHUB_TOKEN_ENV_VAR}</b> env var or `gh` or `hub`.\n'
                                  'Visit `https://github.com/settings/tokens` to generate a new one.')
+        elif err.code == http.HTTPStatus.NOT_FOUND:
```
PawelLipski, line 140: Under what circumstances can 404 happen now? 🤔

git_machete/client.py
rifel123: i did try but without pruning remote i got false positives ;/

So, right now it is allowed to check out closed PR but only when head branch is available on remote, otherwise exception is raised.

git_machete/client.py
rifel123: No. Initially i wanted to grab all PRs here, but in case of git-machete repository it extended execution time for about 0.2-0.3s. In case of bigger/older repositories execution time would much longer, so i dropped this idea. Instead i am asking API for certain PR number in `get_pull_request_by_number_or_none`.

rifel123: Plus there would changes required in `anno-prs` if `all_prs` holds not only open PRs.

git_machete/github.py
rifel123: `get_pull_request_by_number_or_none` call for certain pull request number, hence I expect 404 code when given pull requests was never created.

git_machete/client.py
```diff
@@ -1682,20 +1682,21 @@ def checkout_github_prs(self, pr_no: int) -> None:
...
+        if self.__git.remotes():
+            self.flush_caches()
 
-        pr = get_pull_request_by_number(str(pr_no), org, repo)
+        pr = get_pull_request_by_number_or_none(str(pr_no), org, repo)
         if not pr:
             raise MacheteException(f"PR #{pr_no} is not found in repository `{org}/{repo}`")
-
+        if '/'.join([remote, pr.head]) not in self.__git.get_remote_branches():
+            raise MacheteException(f"Could not checkout PR #{pr_no} because it's head branch `{pr.head}` is already deleted from `{remote}`.")
```
```suggestion
            raise MacheteException(f"Could not check out PR #{pr_no} because its head branch `{pr.head}` is already deleted from `{remote}`.")
```

git_machete/client.py
PawelLipski: Ok! pls rename this var accordingly then to `all_open_prs`

git_machete/client.py
```diff
@@ -1682,18 +1682,21 @@ def checkout_github_prs(self, pr_no: int) -> None:
         remote: str
         remote, (org, repo) = self.__derive_remote_and_github_org_and_repo()
         debug('checkout_github_pr()', f'organization is {org}, repository is {repo}')
+        print(f"Fetching {remote}...")
+        self.__git.fetch_remote(remote)
+        if self.__git.remotes():
```
PawelLipski: Hmm actually I've just noticed now... since we've just fetched `remote` (which corresponds to a Github-hosted remote), there is at least one remote... so this check is redundant

git_machete/client.py
```diff
@@ -1682,18 +1682,21 @@ def checkout_github_prs(self, pr_no: int) -> None:
...
+            self.flush_caches()
 
-        all_prs: List[GitHubPullRequest] = derive_pull_requests(org, repo)
-        pr = utils.find_or_none(lambda pr: pr.number == pr_no, all_prs)
-
+        pr = get_pull_request_by_number_or_none(str(pr_no), org, repo)
```
PawelLipski: I'd rather expect the `get_pull_request_by_number_or_none`'s first param to be an `int` as well (since you're expecting a PR number here... not just any string), the conversion to `str` is an implementation detail

git_machete/github.py
PawelLipski: Oh! ok I see

git_machete/client.py
PawelLipski: Okay! looks fine now 🎉

git_machete/tests/functional/test_machete.py
```diff
@@ -45,10 +46,13 @@ def __init__(self) -> None:
...
 class MockGithubAPIState:
-    def __init__(self, pulls: List[Dict[str, Any]]) -> None:
+    def __init__(self, pulls: List[Dict[str, Any]], issues: List[Dict[str, Any]] = None) -> None:
         self.pulls: List[Dict[str, Any]] = pulls
         self.user: Dict[str, str] = {'login': 'other_user', 'type': 'User', 'company': 'VirtusLab'}  # login must be different from the one used in pull requests, otherwise pull request author will not be annotated
-        self.issues: List[Dict[str, Any]] = []
+        if issues:
+            self.issues: List[Dict[str, Any]] = issues
+        else:
+            self.issues = []
```
PawelLipski:
```suggestion
        self.issues: List[Dict[str, Any]] = issues or []
```

git_machete/client.py
```diff
@@ -1682,21 +1682,23 @@ def checkout_github_prs(self, pr_no: int) -> None:
...
+        self.flush_caches()
 
-        all_prs: List[GitHubPullRequest] = derive_pull_requests(org, repo)
-        pr = utils.find_or_none(lambda pr: pr.number == pr_no, all_prs)
-
+        pr = get_pull_request_by_number_or_none(pr_no, org, repo)
         if not pr:
             raise MacheteException(f"PR #{pr_no} is not found in repository `{org}/{repo}`")
+        if '/'.join([remote, pr.head]) not in self.__git.get_remote_branches():
+            raise MacheteException(f"Could not check out PR #{pr_no} because its head branch `{pr.head}` is already deleted from `{remote}`.")
```
rifel123, line 1693: TBD: Here we could `fetch pull/$ID/head` (also for #242) and return here instead of rising an exception. This would also allow to checkout pull requests from forks, but it might be difficult to find base branch (for example: very old PR #73 can't point to any branch at the moment cause its outdated, but #241 could point to develop)

PawelLipski: Good spot with this corner case (#73)... I'll address on #242 🤔

PawelLipski (approved):

PR #249, opened 2021-09-23
Issue #237: Improve cache management

### LINKED ISSUE #237: Side-effecting operations on `GitContext` should automatically invalidate relevant caches
For instance, `fetch_remotes` will likely make `__remote_branches_cached` and `__commit_sha_by_revision_cached` invalid. As for now, we're just selectively doing `flush_caches`... but that's not enough since there're more methods that modify the state of the underlying git repo, while not invalidating any caches.

I'm afraid some bugs might be lurking even in the command-line client, and there would surely be bugs from the PoV of programatic access.

BartoszSiwinski:
It's mostly method names correction in places where I found it confusing like remotes, was actually a method and not a property/instance variable.

In a `TODO` comment I placed pseudo justification for a `flush_caches()` in that place. In some cases is just an example of a cache that would be no longer valid.

### DIFF
--- a/git_machete/git_operations.py
+++ b/git_machete/git_operations.py
@@ -415,6 +415,22 @@ class GitContext:
         self.__remote_branches_cached = None
         self.__remotes_cached = None
 
+        # TODO: Temporary paste, to make it easier to compare stuff
+        self.__config_cached = None
+        self.__remotes_cached = None
+        self.__counterparts_for_fetching_cached = None
+        self.__short_commit_sha_by_revision_cached = {}
+        self.__tree_sha_by_commit_sha_cached = None
+        self.__commit_sha_by_revision_cached = None
+        self.__committer_unix_timestamp_by_revision_cached = None
+        self.__local_branches_cached = None
+        self.__remote_branches_cached = None
+        self.__initial_log_shas_cached = {}
+        self.__remaining_log_shas_cached = {}
+        self.__reflogs_cached = None
+        self.__merge_base_cached = {}
+        self.__contains_equivalent_tree_cached = {}
+
     def get_revision_repr(self, revision: str) -> str:
         short_sha = self.get_short_commit_sha_by_revision(revision)
         if self.is_full_sha(revision) or revision == short_sha:
--- a/git_machete/git_operations.py
+++ b/git_machete/git_operations.py
@@ -167,10 +167,11 @@ class GitContext:
         if remote not in self.__fetch_done_for:
             self.run_git("fetch", remote)
             self.__fetch_done_for.add(remote)
-            # TODO: BS: flush, because now there is more remotes
+            # TODO: BS: flush, because now there is more remote_branches_cached, counterparts_for_fetching_cached
 
     def set_upstream_to(self, remote_branch: str) -> None:
         self.run_git("branch", "--set-upstream-to", remote_branch)
+        # TODO: BS: flush, because now there is more counterparts_for_fetching_cached
 
     def reset_keep(self, to_revision: str) -> None:
         try:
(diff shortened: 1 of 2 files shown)


git_machete/git_operations.py
```diff
@@ -64,7 +64,7 @@ def get_default_editor(self) -> Optional[str]:
         proposed_editor_funs: List[Tuple[str, Callable[[], Optional[str]]]] = [
             ("$" + git_machete_editor_var, lambda: os.environ.get(git_machete_editor_var)),
```
PawelLipski, line 65: https://github.com/VirtusLab/git-machete/issues/251 ... something to consider while you're at this PR. Not to be done right now, but just to confirm that the existing `run_git` and `popen_git` calls from client.py are safe from the PoV of cache validity.

git_machete/git_operations.py
```diff
@@ -167,9 +167,13 @@ def fetch_remote(self, remote: str) -> None:
         if remote not in self.__fetch_done_for:
             self.run_git("fetch", remote)
             self.__fetch_done_for.add(remote)
+            # TODO (#237): flush, remote_branches_cached and others
```
PawelLipski: For clarity... what is supposed to happen for these TODOs before the PR is merged? 🤔

BartoszSiwinski: They're to be deleted - whole. I don't think it makes sense to keep them as a comment or smth like that. I created them for myself as a debug duck and for reviewers to help them with the understanding of the decisiones. After the review is completed/initially approved they no longer have any purpose.

git_machete/git_operations.py
BartoszSiwinski: I looked into it and they're not. For example `delete_unmanaged` in my opinion should be flushing the cache. Cache clearing should be done on the side of the actual executor of the command that invalidates the cache. `delete_unmanaged` is defined in `client.py` and does that amongst other lines: 
```
if ans in ('y', 'yes'):
    self.__git.run_git("branch", "-d" if is_merged_to_remote else "-D", branch)
```
Ideally we would have here a call to a wrapper around `git branch -d/-D,` and this wrapper would call `run_git` and also invalidate/clear the cache. I would say that the issue #251 should be pulled into `v3.4.0 `release and as a followup/part of this story cache management should be reconsidered.

git_machete/git_operations.py
```diff
@@ -573,14 +586,14 @@ def merged_local_branches(self) -> List[str]:
         ))
 
     def get_hook_path(self, hook_name: str) -> str:
```
PawelLipski, line 585: While you're here... `merged_local_branches` -> `get_merged_local_branches`

git_machete/git_operations.py
```diff
@@ -405,15 +413,20 @@ def create_branch(self, branch: str, out_of_revision: str) -> None:
         self.flush_caches()  # the repository state has changed because of a successful branch creation, let's defensively flush all the caches
```
PawelLipski, line 410: One nuance... `checkout` AFAICS can technically invalidate `__reflogs_cached` and all the `__..._by_revision_cached`... but under the single key,  `HEAD` 🤔

PawelLipski: The question is... are any of this caches used for the `HEAD` key?

PawelLipski: Even if they aren't currently... there's (probably?) currently nothing in place that would forbid such usage

git_machete/git_operations.py
PawelLipski: Okay! into v3.4.0 it goes then

git_machete/git_operations.py
BartoszSiwinski: > One nuance... `checkout` AFAICS can technically invalidate `__reflogs_cached` and all the `__..._by_revision_cached`... but under the single key, `HEAD` 🤔

How can it invalidate that cache?

PawelLipski: Just `flush_caches()` as well, let's not over-optimize with just deleting a single key

BartoszSiwinski: Ok, I've added a `self.flush_caches` at the end of `GitContext.checkout` method

PawelLipski (approved):

PR #255, opened 2021-09-27
Issue #251: Remove direct run_git and popen_git calls from client.py.

### LINKED ISSUE #251: There should be no calls to `run_git` or `popen_git` from client.py
Instead, `MacheteClient` should use dedicated methods from `GitContext` and never interact with `git` binary directly.

This is important not only for clean-code reasons, but also for correctness: `GitContext` maintains caches which might be invalidated by side-effecting `git` operations (like `checkout` or `reset` &mdash; anything that modifies the repo).

BartoszSiwinski:
Things done:
1. Rename `run_git` to `_run_git` to discourage user from using this method outside of `GitContext` class
2. Rename `popen_git` to `_popen_git` to discourage user from using this method outside of `GitContext` class
3. Add methods to replace direct call of `run_git` in `client.py`. These methods are named `run_<git_command>`.
4. Add methods to replace direct call of `popen_git` in `client.py`. These methods are named `get_<git_command>`.

It looks a bit artificial, because now we have `get_log` and `run_log` methods. Is there any particular reason why we don't use always `popen_git` instead of `run_git`?

PR as draft until question above is answered, because decision of changing also that will have a major influence over the PR.

### DIFF
--- a/git_machete/client.py
+++ b/git_machete/client.py
@@ -1178,15 +1178,11 @@ class MacheteClient:
             return
 
         earliest_sha, earliest_short_sha, earliest_subject = commits[0]
-        earliest_full_body = self.__git.popen_git(
-            "log", "-1", "--format=%B", earliest_sha).strip()
+        earliest_full_body = self.__git.get_log("-1", "--format=%B", earliest_sha).strip()
         # %ai for ISO-8601 format; %aE/%aN for respecting .mailmap; see `git rev-list --help`
-        earliest_author_date = self.__git.popen_git(
-            "log", "-1", "--format=%ai", earliest_sha).strip()
-        earliest_author_email = self.__git.popen_git(
-            "log", "-1", "--format=%aE", earliest_sha).strip()
-        earliest_author_name = self.__git.popen_git(
-            "log", "-1", "--format=%aN", earliest_sha).strip()
+        earliest_author_date = self.__git.get_log("-1", "--format=%ai", earliest_sha).strip()
+        earliest_author_email = self.__git.get_log("-1", "--format=%aE", earliest_sha).strip()
+        earliest_author_name = self.__git.get_log("-1", "--format=%aN", earliest_sha).strip()
 
         # Following the convention of `git cherry-pick`, `git commit --amend`, `git rebase` etc.,
         # let's retain the original author (only committer will be overwritten).
@@ -1199,8 +1195,8 @@ class MacheteClient:
         # The tree (HEAD^{tree}) argument must be passed as first,
         # otherwise the entire `commit-tree` will fail on some ancient supported
         # versions of git (at least on v1.7.10).
-        squashed_sha = self.__git.popen_git(
-            "commit-tree", "HEAD^{tree}", "-p", fork_commit, "-m", earliest_full_body,
+        squashed_sha = self.__git.get_commit_tree(
+            "HEAD^{tree}", "-p", fork_commit, "-m", earliest_full_body,
             env=author_env).strip()
 
         # This can't be done with `git reset` since it doesn't allow for a custom reflog message.
@@ -1790,7 +1786,7 @@ class MacheteClient:
         debug(f'create_github_pr({head})', f'organization is {org}, repository is {repo}')
         debug(f'create_github_pr({head})', 'current GitHub user is ' + (current_user or '<none>'))
 
-        title: str = self.__git.popen_git("log", "-1", "--format=%s").strip()
+        title: str = self.__git.get_log("-1", "--format=%s").strip()
         description_path = self.__git.get_git_subpath('info', 'description')
         description: str = utils.slurp_file_or_empty(description_path)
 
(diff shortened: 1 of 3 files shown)


git_machete/client.py
```diff
@@ -946,7 +946,7 @@ def delete_unmanaged(self) -> None:
                 opt_yes_msg = f"Deleting branch {msg_core}"
                 ans = self.ask_if(msg, opt_yes_msg)
                 if ans in ('y', 'yes'):
-                    self.__git.run_git("branch", "-d" if is_merged_to_remote else "-D", branch)
+                    self.__git.run_branch("-d" if is_merged_to_remote else "-D", branch)
```
PawelLipski: Maybe `delete_branch`? with a boolean to param of some adequate name to choose b/w `-D`/`-d` (the flag itself should be an implementation detail, should not leak into the API of `GitContext`)

git_machete/client.py
```diff
@@ -1040,11 +1040,10 @@ def diff(self, branch: Optional[str]) -> None:
             [fp] + \
             ([f"refs/heads/{branch}"] if branch else []) + \
             ["--"]
-        self.__git.run_git("diff", *params)
+        self.__git.run_diff(*params)
 
     def log(self, branch: str) -> None:
-        self.__git.run_git(
-            "log",
+        self.__git.run_log(
```
PawelLipski: Distinction b/w `run_log` and `get_log` is very needed indeed, the former **displays** the log (probably using a pager like `less` for scrolling), while the latter serves for just retrieving the output

git_machete/client.py
```diff
@@ -1199,15 +1194,15 @@ def squash(self, current_branch: str, fork_commit: str) -> None:
         # The tree (HEAD^{tree}) argument must be passed as first,
         # otherwise the entire `commit-tree` will fail on some ancient supported
         # versions of git (at least on v1.7.10).
-        squashed_sha = self.__git.popen_git(
-            "commit-tree", "HEAD^{tree}", "-p", fork_commit, "-m", earliest_full_body,
+        squashed_sha = self.__git.get_commit_tree(
```
PawelLipski: As with other methods, I'd abstract away the flags wherever possible. The API of `GitContext` should not expose the details about the underlying `git`-CLI-based implementation

git_machete/git_operations.py
```diff
@@ -40,14 +40,14 @@ def __init__(self, cli_opts: CommandLineOptions) -> None:
         self.__contains_equivalent_tree_cached: Dict[Tuple[str, str], bool] = {}
 
     @staticmethod
-    def run_git(git_cmd: str, *args: str, **kwargs: Dict[str, str]) -> int:
+    def _run_git(git_cmd: str, *args: str, **kwargs: Dict[str, str]) -> int:
```
PawelLipski, line 44: Hmmm so should it be single or double underscore in private method names? 🤔 is there a difference?

git_machete/git_operations.py
```diff
@@ -40,14 +40,14 @@ def __init__(self, cli_opts: CommandLineOptions) -> None:
         self.__contains_equivalent_tree_cached: Dict[Tuple[str, str], bool] = {}
 
     @staticmethod
```
PawelLipski, line 43: I'd also take a look at public attrs like `git_dir` while you're here, they can probably be made private as well

git_machete/git_operations.py
BartoszSiwinski: https://docs.python.org/3/tutorial/classes.html#private-variables
As per link to doc above __'s used to avoid clashes between subclasses. I personally treat it as private and superprivate. So one is 'I don't recommend using, because it's not meant to be use outside of the class' where double is 'If you touch it, it can break the whole class and everyone you love'.

git_machete/client.py
BartoszSiwinski: I'm not sure if I get you, I tried to do something about this issue. I will upload it with next push and you can decide for yourself if it's sufficient.

PawelLipski: ☹️

PawelLipski: You can rework on a subsequent PR... but generally since the rework is done here, let's implement it already in a solid manner.

PawelLipski: In general, the user of the API, be it `get_commit_tree` or `update_ref`, should **not** be concerned with the details like flags passed to the underlying command. Although yes, technically the original issue is resolved even with exposing a suboptimal API like the current one :trollface:

BartoszSiwinski: Ok, I get you now, you would like user to just call `new_method()` or maybe `new_method(only_one_arg)` and have it done all the rest of the work. I suspect that it will lead to very weird code, because we will be trying to name with words what is actually represented by git commands `options`. What is more, whenever we need slightly different set of options we will have to create new method (which also will cause naming difficulties). 

I will do it, but I suspect that in terms of maintenance we might be looking for troubles in the future. If we would  just call `branch(*args, **kwargs)` [although `kwargs` are debatable] it would be flexible for different methods and yet good enough to manage things like `flushing cache`. There would be a problem how to differentiate between run_git and popen_git, but it is a step forward.

PawelLipski (approved):

PR #257, opened 2021-09-28
enforce correct shebags in CI

### LINKED ISSUE #190: Sort out correct shebang for python
Right now, while building container for rpm package, ambiguous shebang breaks build:
Screenshot 2021-08-25 at 09 13 54

Clarification here is needed

rifel123:
Initially i wanted to replace incorrect shebangs with `sed` in CI but there would be still unchanged ones left in repo, so maybe better to break CI when found?

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -37,6 +37,7 @@ jobs:
             set -x
             enforce-indent-two-spaces-outside-python.sh
             enforce-issue-number-for-todos.sh
+            enforce-correct-shebangs.sh
             ensure-completion-scripts-correct.sh
             ensure-links-correct.sh
             ensure-version-bumped.sh
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -35,9 +35,9 @@ jobs:
 
             PATH=./ci/checks:$PATH
             set -x
+            enforce-correct-shebangs.sh
             enforce-indent-two-spaces-outside-python.sh
             enforce-issue-number-for-todos.sh
-            enforce-correct-shebangs.sh
             ensure-completion-scripts-correct.sh
             ensure-links-correct.sh
             ensure-version-bumped.sh
--- a/ci/checks/enforce-correct-shebangs.sh
+++ b/ci/checks/enforce-correct-shebangs.sh
@@ -1,0 +1,5 @@
+#!/usr/bin/env bash
+
+set -e -o pipefail -u
+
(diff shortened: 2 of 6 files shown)


PawelLipski:
Definitely! no replacing **a n y t h i n g** in the CI... generally, no committing anything back to the same branch from the CI, that's a nightmare to manage. Just break the build and call it a day

ci/checks/enforce-correct-shebangs.sh
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
+
+set -e -o pipefail -u
+
+if grep -r -e '^#!\/usr\/bin\/env python$' -e '^#!\/usr\/bin\/env python2$' ../../git_machete; then
```
PawelLipski: No need to escape `/` in regexes

ci/checks/enforce-correct-shebangs.sh
```diff
@@ -0,0 +1,8 @@
...
+  echo "Ambigous python shebang"
```
PawelLipski: `Ambiguous`; also, as a rule, include the desired form in the error message so that the developer knows how to address the error

ci/checks/enforce-correct-shebangs.sh
PawelLipski: Also, should it be `../../git_machete` or just `git_machete`? 🤔  since AFAICS this script is executed from the top level folder in the CI

.circleci/config.yml
```diff
@@ -37,6 +37,7 @@ jobs:
             set -x
             enforce-indent-two-spaces-outside-python.sh
             enforce-issue-number-for-todos.sh
+            enforce-correct-shebangs.sh
```
PawelLipski: Nit: move up for abc order

ci/checks/enforce-correct-shebangs.sh
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
+
+set -e -o pipefail -u
+
+if grep -r -e '^#!/usr/bin/env python$' -e '^#!/usr/bin/env python2$' git_machete; then
```
PawelLipski: Or just:
```suggestion
if grep -r -e '^#!/usr/bin/env python2?$'; then
```

ci/checks/enforce-correct-shebangs.sh
```diff
@@ -0,0 +1,8 @@
...
+  echo "Ambigouos python shebang, please declare shebangs that point to python3:  #!/usr/bin/env python3"
```
PawelLipski:
```suggestion
  echo "Ambiguous python shebang, please declare shebangs that point to python3:  #!/usr/bin/env python3"
```
;)

`-ouos` suffix is non-existent

PawelLipski (approved):

PR #260, opened 2021-09-29
Remove messages returned by git in tests

### LINKED ISSUE #207: Investigate how to hide output from `git` commands launched by the tested git-machete code in tests
See https://github.com/VirtusLab/git-machete/pull/204#issuecomment-911668526 and the following comment

### LINKED ISSUE #228: `Unclosed file` reported for `os.popen` calls in tests
I've at least observed that when running a single test from IntelliJ .

Probably using `with` should fix the problem here.

```
/usr/lib64/python3.7/subprocess.py:883: ResourceWarning: subprocess 552800 is still running
  ResourceWarning, source=self)
ResourceWarning: Enable tracemalloc to get the object allocation traceback
/home/plipski/git-machete/git_machete/tests/functional/test_machete.py:190: ResourceWarning: unclosed file <_io.TextIOWrapper name=3 encoding='UTF-8'>
  self.remote_path = os.popen("mktemp -d").read().strip()
ResourceWarning: Enable tracemalloc to get the object allocation traceback
```

rifel123:
Closes #207

But the current solution has pretty big con: any output from git will be removed from git_machete everywhere, so in normal flow the user will not able to see for egz: `Switched to branch 'develop'`. Those messages will be outputted when returncode is not 0.

Was there a specific reason to use `subprocess.call` in `run_cmd`? AFAICS `call` i equivalent to  `subprocess.run(...).returncode` and `run` calls `Popen` underneath anyway.

Another way to remove spam would be  just to  mock `run_cmd` from utils module.

So, main question is to capture all stdout/err or just mock it in tests?

### DIFF
--- a/git_machete/tests/functional/test_machete.py
+++ b/git_machete/tests/functional/test_machete.py
@@ -400,7 +400,7 @@ class MacheteTester(unittest.TestCase):
             """,
         )
 
-    @mock.patch('git_machete.utils.run_cmd', mock_run_cmd)
+    @mock.patch('git_machete.utils.run_cmd', mock_run_cmd)  # to hide git outputs in tests
     def test_branch_reappers_in_definition(self) -> None:
         body: str = \
             """master
@@ -420,7 +420,7 @@ class MacheteTester(unittest.TestCase):
             if e.parameter != expected_error_msg:
                 self.fail(f'Actual Exception message: {e} \nis not equal to expected message: {expected_error_msg}')
 
-    @mock.patch('git_machete.utils.run_cmd', mock_run_cmd)
+    @mock.patch('git_machete.utils.run_cmd', mock_run_cmd)  # to hide git outputs in tests
     def test_show(self) -> None:
         self.setup_discover_standard_tree()
 
@@ -445,7 +445,7 @@ class MacheteTester(unittest.TestCase):
             "ignore-trailing"
         )
 
-    @mock.patch('git_machete.utils.run_cmd', mock_run_cmd)
+    @mock.patch('git_machete.utils.run_cmd', mock_run_cmd)  # to hide git outputs in tests
     def test_traverse_no_push(self) -> None:
         self.setup_discover_standard_tree()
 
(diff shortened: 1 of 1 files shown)


rifel123:
Seems like it also covers #228

git_machete/utils.py
```diff
@@ -168,7 +168,12 @@ def run_cmd(cmd: str, *args: str, **kwargs: Any) -> int:
     elif verbose_mode:
         sys.stderr.write(flat_cmd + "\n")
 
-    exit_code: int = subprocess.call([cmd] + list(args), **kwargs)
+    completed_process: subprocess.CompletedProcess[bytes] = subprocess.run([cmd] + list(args), stdout=subprocess.PIPE, stderr=subprocess.PIPE, **kwargs)
+    exit_code: int = completed_process.returncode
+
+    if exit_code != 0:
+        sys.stdout.writelines(completed_process.stdout.decode('utf-8'))
```
PawelLipski: Hmmm this is a change in logic now... since AFAICS that the entire output will only be printed **once the process is finished**, and not while the process is running. I can't think now of any specific command that will be affected (maybe `git rebase`?) but I doubt this will NOT have any consequence at all

PawelLipski: Oh okay, I wish I'd read the PR description more thoroughly first 😅

PawelLipski: > Another way to remove spam would be just to mock run_cmd from utils module.

Hmmm yeah let's try mocking `run_cmd` or `subprocess.call` in the tests. Don't proceed however if it requires too much hacking/"engineering", we can live with #207 unresolved (#228 would be still nice to fix, tho)

PawelLipski: > So, main question is to capture all stdout/err or just mock it in tests?

Generally,  **no change to logic in production code** should happen, only to test code

PawelLipski:
Nit: some stray output still left in one test:

image

just curious if it's a sign of some underlying inconsistency in test/mock design, not that I care about these specific two lines 🤔

rifel123:
those messages comes from `print` from`checkout-prs` and `create-pr`, where commands are called directly from class, so `launch_command` does't catch them.

PawelLipski (approved):

PR #262, opened 2021-10-01
Check PEP8 correctness for one python version in CI

### LINKED ISSUE #184: Is it possible for PEP8 check to fail against one Python version if it passed against another?
We're currently running pep8 check in the CI for each supported Python version ([3.6, 3.9]).
But is it really needed? Maybe just running against one version is enough?

~Same point applies to `black` check, which is to be implemented within the scope of #140 (and will probably replace pep8 check).~ EDIT, we resigned from `black`, too opinionated

### LINKED ISSUE #220: Include git version (not just Python version) in CircleCI job names
Since e.g. this job https://app.circleci.com/pipelines/github/VirtusLab/git-machete/661/workflows/668b9317-7743-476a-a04a-a50ea32df7d8/jobs/1391 is displayed as failing `tests_py36` in Github... while the failure here was related NOT to Python version, but rather to the old git version, this might be confusing. Also, maybe use more friendly format like `test python 3.6, git 1.8.10` if possible

rifel123:
closes #184 and very small issue #220

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -12,7 +12,7 @@ mandatory_jobs: &mandatory_jobs
   - test against python 3_6 and git 1_8_10
   - test against python 3_7 and git 2_7_6
   - test against python 3_8 and git 2_25_0 with PEP8 checks
-  - test against python 3_9 and git 2_30_1 with corevarage report and docs creation
+  - test against python 3_9 and git 2_30_1 with coverage report and docs creation
 
 jobs:
   "initial checks":
@@ -74,7 +74,7 @@ jobs:
           name: Run tests against Python 3.8 and git 2.25.0 (check PEP8 correctness)
           command: PYTHON_VERSION=3.8 GIT_VERSION=2.25.0 PEP8=true bash ci/tox/ci-run.sh
 
-  "test against python 3_9 and git 2_30_1 with corevarage report and docs creation":
+  "test against python 3_9 and git 2_30_1 with coverage report and docs creation":
     executor: ubuntu_executor
     steps:
       - checkout
@@ -159,7 +159,7 @@ workflows:
       - test against python 3_6 and git 1_8_10
       - test against python 3_7 and git 2_7_6
       - test against python 3_8 and git 2_25_0 with PEP8 checks
-      - test against python 3_9 and git 2_30_1 with corevarage report and docs creation
+      - test against python 3_9 and git 2_30_1 with coverage report and docs creation
       - dry_run_deploy_from_develop:
           filters:
             branches:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -44,7 +44,7 @@ jobs:
             prohibit-tab-character.sh
             prohibit-trailing-whitespace.sh
 
-  tests_py36:
+  tests_py36_git_1.8.0:
     executor: ubuntu_executor
     steps:
       - checkout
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -44,31 +44,31 @@ jobs:
...
-  tests_py38:
+  tests_py38_git_2_25_0_pep8:
     executor: ubuntu_executor
     steps:
       - checkout
       - run:
-          name: Run tests against Python 3.8 and git 2.25.0
-          command: PYTHON_VERSION=3.8 GIT_VERSION=2.25.0 bash ci/tox/ci-run.sh
+          name: Run tests against Python 3.8 and git 2.25.0 (check PEP8 correctness)
+          command: PYTHON_VERSION=3.8 GIT_VERSION=2.25.0 FLAKE=true bash ci/tox/ci-run.sh
```
PawelLipski: I'd use `pep8`/`PEP8` name everywhere... the fact that `flake8` is used underneath is an impl detail

.circleci/config.yml
```diff
@@ -44,31 +44,31 @@ jobs:
             prohibit-tab-character.sh
             prohibit-trailing-whitespace.sh
 
-  tests_py36:
+  tests_py36_git_1_8_0:
```
PawelLipski: I've granted you admin access to this repo, pls change the required checks to pass for master/develop under https://github.com/VirtusLab/git-machete/settings/branches

.circleci/config.yml
```diff
@@ -150,27 +150,27 @@ workflows:
...
+            - tests_py36_git_1_8_0
+            - tests_py37_git_2_7_6
+            - tests_py38_git_2_25_0_pep8
+            - tests_py39_git_2_30_1_coverage_docs
       - deploy_from_master:
           filters:
             branches:
               only: "master"
           requires:
             - initial_checks
```
PawelLipski: Nit: you can use a YAML anchor/alias here

.circleci/config.yml
PawelLipski: Actually! as suggested in the issue, would it be possible to use `test against python 3.6 and git 1.8.10` job name format? probably some quoting would be need in yaml for that

rifel123: Unfortunately it's not possible as far as can see. It's connected to how circleCI validates jobs names: every job must meet regex pattern: `^[A-Za-z][A-Za-z\s\d_-]*$` where dots are not allowed. But we could use something similar to `test against python 3_6 and git 1_8_10` only.

PawelLipski: Ok, let's go for the space/underscore version

.circleci/config.yml
```diff
@@ -149,28 +155,20 @@ workflows:
...
-      - initial_checks
-      - tests_py36
-      - tests_py37
-      - tests_py38
-      - tests_py39_coverage_docs
+      - initial checks
+      - test against python 3_6 and git 1_8_10
+      - test against python 3_7 and git 2_7_6
+      - test against python 3_8 and git 2_25_0 with PEP8 checks
+      - test against python 3_9 and git 2_30_1 with coverage report and docs creation
```
rifel123, line 163: this needs to stay, as yaml cannot extend arrays, here is further explanation:
https://github.com/yaml/yaml/issues/35

PawelLipski: Wow didn't know about this limitation 😯

ci/tox/Dockerfile
```diff
@@ -37,6 +37,11 @@ ENV CHECK_COVERAGE=${check_coverage}
 ARG build_docs
 ENV BUILD_DOCS=${build_docs}
 
+# Pass build argument to flake8_check environment variable that is used to
+# determine where to run tox enviornment that runs check against PEP8 standard.
```
PawelLipski:
```suggestion
# Pass build argument to PEP8 environment variable that is used to
# determine where to run tox environment that runs check against PEP8 standard.
```

PawelLipski (approved):

PR #254, opened 2021-09-27
Allow to checkout pull requests from forks

rifel123:

PR #263, opened 2021-10-01
Issue #169: Enforce that fork-point options values have to be ancestors of branch.

### LINKED ISSUE #169: `--(down)-fork-point` options should probably require that the provided commit is ancestor of the branch to rebase/squash
A problem detected by @Sarunoi.

Let's assume we're on branch B. Machete parent of B is branch A.
When the commit X passed as argument to `git machete update --fork-point X` is NOT ancestor commit of B,
then the underlying operation (`git rebase --onto A X B`) probably rebases the history from common ancestor commit of X and B (exclusive) to B (inclusive) &mdash; instead of from X (excl.) to B (incl.).

I can't see much use in allowing for such case, and probably it'd be better if we fail with an error in case X is NOT ancestor commit of B.

TBD what to do when X == B &mdash; we can as well fail `git machete update` in such case instead of running a no-op rebase (?)

BartoszSiwinski:
I'm not in love with this solution, because `slide-out` is handling this validation in different place than `reapply`, `squash` and `update`.

It makes sense at the current state of the repo that this validation would be taken care of in the `cli.py`, because there are already some other validation like that there. Unfortunately for `slide-out` I need to know what child branches are connected with the last passed branch and to do that I'm using "superprivate" (`double _`) `MacheteClient` variable that stores this information (`__down_branches: dict`).

### DIFF
--- a/git_machete/cli.py
+++ b/git_machete/cli.py
@@ -655,6 +655,9 @@ def launch(orig_args: List[str]) -> None:
             machete_client.read_definition_file()
             git.expect_no_operation_in_progress()
             current_branch = git.get_current_branch()
+            if "fork_point" in parsed_cli:
+                git.check_that_forkpoint_is_ancestor_or_equal_to_tip_of_branch(
+                    forkpoint_sha=cli_opts.opt_fork_point, branch=current_branch)
             git.rebase_onto_ancestor_commit(
                 current_branch,
                 cli_opts.opt_fork_point or machete_client.fork_point(
@@ -683,6 +686,9 @@ def launch(orig_args: List[str]) -> None:
             machete_client.read_definition_file()
             git.expect_no_operation_in_progress()
             current_branch = git.get_current_branch()
+            if "fork_point" in parsed_cli:
+                git.check_that_forkpoint_is_ancestor_or_equal_to_tip_of_branch(
+                    forkpoint_sha=cli_opts.opt_fork_point, branch=current_branch)
             machete_client.squash(
                 current_branch,
                 cli_opts.opt_fork_point or machete_client.fork_point(
@@ -706,6 +712,9 @@ def launch(orig_args: List[str]) -> None:
         elif cmd == "update":
             machete_client.read_definition_file()
             git.expect_no_operation_in_progress()
+            if "fork_point" in parsed_cli:
+                git.check_that_forkpoint_is_ancestor_or_equal_to_tip_of_branch(
+                    forkpoint_sha=cli_opts.opt_fork_point, branch=git.get_current_branch())
             machete_client.update()
         elif cmd == "version":
             version()
--- a/git_machete/client.py
+++ b/git_machete/client.py
@@ -474,6 +474,20 @@ class MacheteClient:
             if not new_upstream:
                 raise MacheteException(f"No upstream branch defined for `{branch}`, cannot slide out")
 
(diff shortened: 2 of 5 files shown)


git_machete/client.py
```diff
@@ -474,6 +474,21 @@ def slide_out(self, branches_to_slide_out: List[str]) -> None:
...
 
+        if self.__cli_opts.opt_down_fork_point:
+            last_branch_to_slide_out = branches_to_slide_out[-1]
+            children_of_the_last_branch_to_slide_out = self.__down_branches.get(
+                last_branch_to_slide_out)
+
+            if len(children_of_the_last_branch_to_slide_out) > 1:
+                raise MacheteException(
+                    "Last branch to slide out can't have more than one child branch "
+                    "if option --down-fork-point is passed.")
```
PawelLipski: Nit: 
```suggestion
                    "if option `--down-fork-point` is passed.")
```

git_machete/client.py
```diff
@@ -474,6 +474,21 @@ def slide_out(self, branches_to_slide_out: List[str]) -> None:
             if not new_upstream:
                 raise MacheteException(f"No upstream branch defined for `{branch}`, cannot slide out")
 
+        if self.__cli_opts.opt_down_fork_point:
+            last_branch_to_slide_out = branches_to_slide_out[-1]
+            children_of_the_last_branch_to_slide_out = self.__down_branches.get(
+                last_branch_to_slide_out)
+
+            if len(children_of_the_last_branch_to_slide_out) > 1:
+                raise MacheteException(
```
PawelLipski, line 483: Pls also cover this case in functional tests

git_machete/git_operations.py
```diff
@@ -719,3 +723,13 @@ def get_latest_checkout_timestamps(self) -> Dict[str, int]:  # TODO (#110): defa
...
         return result
+
+    def check_that_forkpoint_is_ancestor_or_equal_to_head_of_branch(
+            self, forkpoint_sha: str, branch: str) -> None:
+        if not self.is_ancestor_or_equal(
+                earlier_revision=forkpoint_sha,
+                earlier_prefix='',
+                later_revision=branch):
+            raise MacheteException(
+                f"Forkpoint {forkpoint_sha} is not ancestor of or HEAD "
```
```suggestion
                f"Fork point {forkpoint_sha} is not ancestor or tip "
```

`HEAD` written in upper case never refers to a tip of a branch. Sometimes `head` in lower case is used in that context by git docs... but this is also confusing. `Tip` is more unambiguous... although given that branches are pointers anyway... `tip of branch X` is kinda equivalent to `branch X` anyway

git_machete/cli.py
```diff
@@ -655,6 +655,9 @@ def strip_first_fragment(remote_branch: str) -> str:
             machete_client.read_definition_file()
             git.expect_no_operation_in_progress()
             current_branch = git.get_current_branch()
+            if "fork_point" in parsed_cli:
```
PawelLipski, line 658: Pls add some test case (at least for one of these commands) where an error is raised

BartoszSiwinski:
After getting approve I will deal with the conflicts and rebase the branch.

PawelLipski (approved):
:benissimo:

PawelLipski:
Pls refrain with merging tho until v3.4.0 is released (which should happen soon after #254 is merged... unless some new issues are discovered in the meantime ofc)

PR #266, opened 2021-10-06
Issue #233: Provide support for 4th location of GitHub token

### LINKED ISSUE #233: Provide support for 4th location for GitHub token
_We already have 3 possible ways_ (`GITHUB_TOKEN` env var, `gh` config file, `hub` config file), _why add yet another?_

Well, the users might not use `gh` or `hub` (in fact I'm myself not using any of them at all since `git machete github` is available in alpha version). Also, having `GITHUB_TOKEN` exported in .zshrc or .bashrc is also questionable (e.g. since then this secret is passed in environment to every single process executed from the shell).

Hence, a 4th alternative should be consider: maybe `~/.github-token` with just token (no yaml, no json - just plain token)? Or let's see if there's some established convention out there how such file can be called.

BartoszSiwinski:


git_machete/__init__.py
```diff
@@ -1 +1,5 @@
+import os
+
 __version__ = '3.4.0'
+
+ROOT_DIR = os.path.abspath(__file__).replace('git_machete/__init__.py', '')
```
rifel123: I would suggest not to keep this file in repository, cause it would have to be added to `gitignore` in every case and if someone accidentally push this file to github then token gets immediately invalid.  
As suggested in issue #233, i think it's better to keep it in users directory of `~/` which makes token available for every repository that user may have. 

if it has to be kept in repository for some reason, maybe put this file next to definition file under `.git` directory?

BartoszSiwinski: Hm... It shouldn't be pushed, because it's added to `.gitignore`, so that shouldn't be an issue. Regarding putting it in `~` instead. I'm open to change. I will wait for @PawelLipski to review and unless he will disagree I will move it there.

PawelLipski: I strongly second @rifel123's suggestion, for the sake of:
1. security: to prevent people from ever pushing the token to GitHub in case they forget to add the file to `.gitignore` (or `.git/info/exclude`
2. convenience: GitHub tokens are NOT scoped to the individual repos, they're account-wide; it doesn't make sense to keep separate tokens on per-repository basis locally, you just have one token for all GitHub repos on your local machine

`~/.github-token` FTW

git_machete/github.py
```diff
@@ -100,7 +101,20 @@ def get_token_from_hub() -> Optional[str]:
     def get_token_from_env() -> Optional[str]:
         return os.environ.get(GITHUB_TOKEN_ENV_VAR)
 
-    return get_token_from_env() or get_token_from_gh() or get_token_from_hub()
+    def get_token_from_file_in_root() -> Optional[str]:
+        required_file_name = '.github-token'
+        file_full_path = f'{ROOT_DIR}/{required_file_name}'
```
PawelLipski: Whooops isn't `ROOT_DIR` just the installation directory of `git-machete` on the given machine (rather than root directory of the current project)? I'd rather expect `git_machete.git_operations.GitContext.get_root_dir` for this purpose

PawelLipski:
Updated milestone to v3.5.0 btw.

README.md
```diff
@@ -176,6 +176,7 @@ This will be resolved from the first of:
 1. The `GITHUB_TOKEN` env var.
 2. The auth token from the current [`gh`](https://cli.github.com/) configuration.
 3. The auth token from the current [`hub`](https://github.com/github/hub) configuration.
+4. The content of the .github-token file in the home directory (`~`). This file has to be manually created by the user.
```
PawelLipski: A matter of taste, but I'd move this to point two, after env var, before `gh`/`hub` (in the sense that peeking into the token used by other tools is more of a hack/fallback)

git_machete/github.py
```diff
@@ -100,7 +100,20 @@ def get_token_from_hub() -> Optional[str]:
     def get_token_from_env() -> Optional[str]:
         return os.environ.get(GITHUB_TOKEN_ENV_VAR)
 
-    return get_token_from_env() or get_token_from_gh() or get_token_from_hub()
+    def get_token_from_file_in_home_directory() -> Optional[str]:
+        required_file_name = '.github-token'
+        file_full_path = f'~/{required_file_name}'
```
PawelLipski: Hmmm does Python do tilde expansion by default? https://stackoverflow.com/questions/2057045/pythons-os-makedirs-doesnt-understand-in-my-path

PawelLipski (approved):

PawelLipski:
Btw! pls update version in __init__ and add an entry to RELEASE_NOTES.md

PR #273, opened 2021-10-11
Prerelease fixes: Update tox envs dependencies and release notes

BartoszSiwinski:
Add mypy dependency to tox environment mypy-py{36,37,38,39} and join release notes from 3.3.1 and 3.4.0 since there was no release 3.3.1.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -54,6 +54,7 @@ whitelist_externals = tox
 commands = tox -e "mypy-py{36,37,38,39}"
 
 [testenv:mypy-py{36,37,38,39}]
+deps = mypy
 whitelist_externals = mypy
 commands =
   mypy --config-file mypy.ini git_machete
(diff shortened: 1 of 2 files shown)


tox.ini
```diff
@@ -54,6 +54,7 @@ whitelist_externals = tox
 commands = tox -e "mypy-py{36,37,38,39}"
 
 [testenv:mypy-py{36,37,38,39}]
+deps = mypy
 whitelist_externals = mypy
```
PawelLipski: Hmm is now this `whitelist_externals` needed? Coz e.g. there's no `whitelist_externals` for `sphinx-build` or `coverage` in the envs above

PawelLipski: I'd assume that a command that's provided by `deps` is no longer considered an _external_

PawelLipski (approved):

PR #281, opened 2021-10-13
Issue #161: Decouple MacheteClient and GitContext from CommandLineOptions class.

### LINKED ISSUE #161: Decouple `MacheteClient` and `GitContext` API from command-line options

BartoszSiwinski:


git_machete/cli.py
```diff
@@ -508,13 +508,17 @@ def launch(orig_args: List[str]) -> None:
         if cmd == "add":
             machete_client.read_definition_file()
             branch = get_branch_arg_or_current_branch(cli_opts, git)
-            machete_client.add(branch)
+            machete_client.add(
+                branch=branch,
+                opt_onto=cli_opts.opt_onto,
```
PawelLipski, line 513: Okay before I go further in the review: so finally you went for the option A from our Slack conversation (keep API param names in sync with CLI options)... any thoughts on why not B (API param names agnostic of CLI)?

BartoszSiwinski: Ok, I got you wrong on Slack. I've mentioned that I prefer B, you mentioned that you like second one and I understood that second to my so A, and you meant second overall which is B. Well that's awkward.

I will reiterate the pros and cons and you can decide here:


Let's consider command `add`. It has following options:
`git machete add [-o|--onto=<target-upstream-branch>] [-R|--as-root] [-y|--yes] [<branch>]`

Corresponding method from MacheteClient could have one of two signatures:
A) `add(branch: str, opt_onto: str = None, opt_as_root: bool = False, opt_yes: bool = False)`
pros: 
- straightforward rewriting from cli_opt, so code is easy to maintain and debug, 
- API is consistent with CLI so it's easier to understand what is going on with these argument knowing the CLI side of the project (which should usually the case)
- it's clear which arguments are optional and which aren't (opt_ prefix is rather logical after a moment in the repo/knowing script writing convention)
- code is resembling connection with CLI which is overall theme of the repo (given the fact that we have few places where we ask user about something using this MacheteClient as lib is rather unrealistic - we would have to create always_yes mechanism that would override any questions to user and assume approval)
- some of the options have negative type of a name `no_edit_merge`, if we would like to change API to be more `lib`-like we should create argument `edit_merge` and assign it value `not cli_opt.no_edit_merge`, which would cause a little headache for someone looking at the code if every second arg is `arg = not_opposite_arg`
- it's already done this way

B) `add(branch: str, parent_branch: str = None, as_root: bool = False, skip_user_confirmation: bool = False)`
pros: 
- API from dev perspective is more descriptive about what is doing what. For example opt_yes means nothing, in a few places is even misleading. Here it's replaced with skip_user_confirmation that is basically what is does in this context. Not all of the options have this problem, but there are few that could probably be replaced with something more meaningful in terms of code.
- a step closer to make git-machete project into real library

PawelLipski: Okay! an excellent breakdown, I'm convinced now to the option A (esp. the point _given the fact that we have few places where we ask user about something using this MacheteClient as lib is rather unrealistic_)

Let's proceed with the existing approach, I'll review the code as is now...

git_machete/cli.py
```diff
@@ -633,7 +653,7 @@ def strip_first_fragment(remote_branch: str) -> str:
             elif category == "managed":
                 res = machete_client.managed_branches
             elif category == "slidable":
-                res = machete_client.slidable()
+                res = machete_client.get_slidable_branches()
```
PawelLipski, line 656: Right, good spot

git_machete/client.py
```diff
@@ -438,23 +450,27 @@ def get_root_of(branch: str) -> str:
             self.managed_branches = excluding(self.managed_branches, merged_branches_to_skip)
             for branch in merged_branches_to_skip:
                 upstream = self.up_branch[branch]
-                self.__down_branches[upstream] = excluding(self.__down_branches[upstream], [branch])
+                self.__down_branches[upstream] = excluding(self.__down_branches.get(upstream), [branch])
```
PawelLipski: Hmmm why this (slight, but still) change in logic? 🤔

git_machete/client.py
```diff
@@ -1513,9 +1608,26 @@ def set_fork_point_override(self, branch: str, to_revision: str) -> None:
         self.__git.set_config_attr(while_descendant_of_key, b_sha)
 
         sys.stdout.write(
-            fmt(f"Fork point for <b>{branch}</b> is overridden to <b>{self.__git.get_revision_repr(to_revision)}</b>.\n", f"This applies as long as {branch} points to (or is descendant of) its current head (commit {self.__git.get_short_commit_sha_by_revision(b_sha)}).\n\n", f"This information is stored under git config keys:\n  * `{to_key}`\n  * `{while_descendant_of_key}`\n\n", f"To unset this override, use:\n  `git machete fork-point --unset-override {branch}`\n"))
```
PawelLipski, line 1516: Whooops that was long 😅

git_machete/client.py
```diff
@@ -1719,19 +1869,22 @@ def checkout_github_prs(self, pr_no: int) -> None:
 
         all_open_prs: List[GitHubPullRequest] = derive_pull_requests(org, repo)
 
-        self.__cli_opts.opt_yes = True  # TODO (#161): pass only needed options to methods
         path: List[str] = self.__get_path_from_pr_chain(pr, all_open_prs)
         reversed_path: List[str] = path[::-1]  # need to add from root downwards
         for index, branch in enumerate(reversed_path):
             if branch not in self.managed_branches:
                 if index == 0:
-                    self.__cli_opts.opt_as_root = True
```
PawelLipski, line 1728: 😌

PawelLipski (approved):
🧹

git_machete/client.py
BartoszSiwinski: It was workaround for type checking, that I actually don't need, because I later found a real issue which was inaccurate type for __down_branches. I will restore it.

PR #287, opened 2021-10-15
Issue #265: Create exit_script method that will process script exiting.

### LINKED ISSUE #265: Create exit_script(status_code=None) method and use it as the only exit point of the script.
Currently we have multiple `sys.exit` in a few places in the cli.py with different exit codes. It would be good to move all of these sys.exit to one method and process leaving the script there. This way it would be easy to mock exiting from script in tests which in turn would enable us to catch actual errors like MacheteException.

This change would affect for example this code:
```
    except (argparse.ArgumentError, argparse.ArgumentTypeError) as e:
        print(get_short_general_usage())
        sys.stderr.write(f"\n{e}\n")
        sys.exit(2)
    except MacheteException as e:
        sys.stderr.write(f"\n{e}\n")
        sys.exit(1)
    except KeyboardInterrupt:
        sys.stderr.write("\nInterrupted by the user")
        sys.exit(1)
    except StopInteraction:
        pass
```
After the issue
```
    except (argparse.ArgumentError, argparse.ArgumentTypeError) as e:
        print(get_short_general_usage())
        sys.stderr.write(f"\n{e}\n")
        exit_script(status_code=2)
    except MacheteException as e:
        sys.stderr.write(f"\n{e}\n")
        exit_script(status_code=1)
    except KeyboardInterrupt:
        sys.stderr.write("\nInterrupted by the user")
        exit_script(status_code=1)
    except StopInteraction:
        pass
```

BartoszSiwinski:


git_machete/client.py
```diff
@@ -1225,8 +1225,8 @@ def up(self, branch: str, prompt_if_inferred_msg: Optional[str],
                             opt_yes=False
                     ) in ('y', 'yes'):
                         return upstream
-                    else:
-                        sys.exit(1)
+                    raise MacheteException(
```
PawelLipski: This seems pretty verbose... maybe just `Aborting.`?

git_machete/client.py
```diff
@@ -1265,9 +1265,8 @@ def __run_post_slide_out_hook(self, new_upstream: str, slid_out_branch: str,
             exit_code = utils.run_cmd(hook_path, new_upstream, slid_out_branch, *new_downstreams,
                                       cwd=self.__git.get_root_dir())
             if exit_code != 0:
-                sys.stderr.write(
+                raise MacheteException(
                     f"The machete-post-slide-out hook exited with {exit_code}, aborting.\n")
-                sys.exit(exit_code)
```
PawelLipski, line 1270: This is a minuscule change in logic wrt. exit code, pls just see if docs for `hooks` require an update

git_machete/git_operations.py
```diff
@@ -665,8 +665,8 @@ def get_all_lines_fixed() -> Iterator[str]:
             if exit_code == 0:
                 do_rebase()
             else:
-                sys.stderr.write("The machete-pre-rebase hook refused to rebase.\n")
-                sys.exit(exit_code)
+                raise MacheteException(
+                    f"The machete-pre-rebase hook refused to rebase. Error code: {exit_code}\n")
```
PawelLipski, line 669: Same here, change in logic wrt. exit code, can stay as long as it's reflected in docs

git_machete/client.py
```diff
@@ -1416,25 +1415,25 @@ def __sync_annotations_to_definition_file(self, prs: List[GitHubPullRequest], cu
...
             return self.prev_branch(branch)
-        elif param in ("r", "root"):
+        elif param in {"r", "root"}:
             return self.root_branch(branch, if_unmanaged=PICK_FIRST_ROOT)
-        elif param in ("u", "up"):
+        elif param in {"u", "up"}:
             return self.up(branch, prompt_if_inferred_msg=None, prompt_if_inferred_yes_opt_msg=None)
-        else:
-            raise MacheteException(f"Invalid direction: `{param}` expected: {allowed_directions(allow_current)}")
+        raise MacheteException(f"Specified direction '{param}' is not supported.")
```
PawelLipski, line 1435: I think including `allowed_directions` in the error message was pretty useful

git_machete/cli.py
```diff
@@ -460,18 +460,22 @@ def validate_python_version() -> None:
...
-        sys.exit(1)
+        exit_script(1)
 
 
 def get_branch_arg_or_current_branch(
         cli_opts: git_machete.options.CommandLineOptions, git_context: GitContext) -> str:
     return cli_opts.opt_branch or git_context.get_current_branch()
 
 
+def exit_script(status_code: Optional[int] = None) -> None:
```
PawelLipski, line 462: Pls mention in a comment why that's useful so that _future generations_ don't scratch their heads

git_machete/client.py
BartoszSiwinski: To be honest I'm not sure about that. `direction` arg for `go` and `show` commands is already restricted via argparse internal functionalities (see [choices](https://docs.python.org/3/library/argparse.html#choices)). 
This exception is raised here only in case there will be discrepancy between what `parser` is permitting and what `parse_direction` method can process which would be implementation bug and not user entering invalid args.

If you still would like me to add this information please put a comment here and I will introduce it as requested. At the moment I think that it will be just an unnecessary bias in the error message (this error is rather expected to pop out for a developer than user).

rifel123: Yes, since migration to argparse, this error is not likely to show, cause argparse will do argument validation at the very beginning. Also, if wrong argument is passed, error is raised with similar message to error filled with `allowed_directions`:
Screenshot 2021-10-21 at 09 25 36

Hence, I'd leave as it is, cause we have this checked earlier.

git_machete/client.py
BartoszSiwinski: It's about ok, but I will do small update to be more precise/straightforward about was is happening with this hook.

git_machete/git_operations.py
BartoszSiwinski: It's about ok, but I will update it to be more precise.

git_machete/client.py
PawelLipski: Ok! if so, then let's leave as is

PawelLipski (approved):

Issue #118, opened 2021-02-12
Turn on strict optional (mypy)


mkondratek:
This IIUC will enable None-correctness checks similar to Kotlin or [NullnessChecker](https://checkerframework.org/manual/#nullness-checker) in Java

PawelLipski:
For anybody who takes this task: pls first assess the scope of changes this will cause in the repo, maybe the extra soundness is not worth the extra complication/loss of readability

PawelLipski:
Wow it's done, I forgot to close it ‼️

PR #294, opened 2021-10-18
Add custom types for AnyBranch, LocalBranch, RemoteBranch, Commit

### LINKED ISSUE #172: Distinguish b/w branches and commit hashes using type system
Currently, we use `str` for pretty much everything... which is error-prone (one can e.g. accidentally pass commit hash where branch name is expected). The type system should ideally distinguish b/w:
* any branch (maybe local, maybe remote)
* local branch
* remote branch
* commit hash

rifel123:
Currently everywhere `str` is used, so mypy cannot precisely check types compatibility.


git_machete/client.py
```diff
@@ -1424,23 +1427,23 @@ def __infer_upstream(self, branch: str, condition: Callable[[str], bool] = lambd
...
                 self.__git.get_config_attr_or_none(self.config_key_for_override_fork_point_while_descendant_of(branch))) is not None
 
-    def __get_fork_point_override_data(self, branch: str) -> Optional[Tuple[str, str]]:
+    def __get_fork_point_override_data(self, branch: AnyBranch) -> Optional[Tuple[Commit, Commit]]:
         to_key = self.config_key_for_override_fork_point_to(branch)
-        to = self.__git.get_config_attr_or_none(to_key)
+        to = Commit(self.__git.get_config_attr_or_none(to_key) or "")
         while_descendant_of_key = self.config_key_for_override_fork_point_while_descendant_of(branch)
-        while_descendant_of = self.__git.get_config_attr_or_none(while_descendant_of_key)
+        while_descendant_of = Commit(self.__git.get_config_attr_or_none(while_descendant_of_key) or "")
```
rifel123: `or ""` is necessary here, because python allows to use `str(None)` which returns "None" (non empty string).
This causes `if str(None)` to be aways true.

git_machete/cli.py
```diff
@@ -547,7 +548,7 @@ def launch(orig_args: List[str]) -> None:
             if cli_opts.opt_inferred:
                 print(machete_client.fork_point(branch, use_overrides=False))
             elif cli_opts.opt_override_to:
-                machete_client.set_fork_point_override(branch, cli_opts.opt_override_to)
+                machete_client.set_fork_point_override(branch, Commit(cli_opts.opt_override_to))
```
PawelLipski: `or ""` would probably be needed here as well...

git_machete/utils.py
```diff
@@ -47,7 +47,7 @@ def tupled(f: Callable[..., T]) -> Callable[[Any], T]:
     return lambda tple: f(*tple)
 
 
-def get_second(pair: Tuple[str, str]) -> str:
+def get_second(pair: Tuple[Any, str]) -> str:
```
PawelLipski: Maybe just rework to generic argument(s) while you're here

git_machete/cli.py
PawelLipski: Oh! okay, no, coz we already know that `opt_override_to` is non-None

git_machete/client.py
PawelLipski: Hmmm this is a pretty brittle construct... would it make sense to do sth like:

1. declare a factory method in `Commit`:
`@staticmethod def of(value: str): Optional[Commit] = Commit(value) if value else None`

2. never use `Commit(...)` directly but always `Commit.of` for (relative) safety

Same for other classes as well for consistency... so as to make sure that out of `None` string, we get a `None`, and not a wrapped empty string 🤔

PawelLipski:
Pls note down on the original issues what were the problems with alternatives, like newtypes and simple wrapper classes

rifel123:
I can see three options here but every solution has its pros and cons:
1.  Use `NewType` from `typing` module  to make distinction between str and NewTypes.
 pros:
  - No massive changes in code
  - Each new type is in fact still a string so we have all it's functionalities and interface
  - Faster that classes, as `NewType` returns only function to limit overhead ([explanation under NewType section](https://mypy.readthedocs.io/en/stable/more_types.html))
 cons:
  - No way to make `AnyBranch` type which could accept  both `LocalBranch` and `RemoteBranch`. Each type is completely separate.
  - Since NewTypes are functions there is no possibility to  make class reserved checks (egz `isinstance()`, currently not used anywhere in git-machete) 
  - weird type casting, as NewTypes can accept only base class (well, can accept anything but mypy shouts!)
  
2.  Declare types as classes:
 pros:
   - `AnyBranch` can accept all (`AnyBranch`, `LocalBranch`, `RemoteBranch`)
   - full separation from string class
  cons:
   - massive changes in code, cause there is not string interface, for example, even if used string as class variable, there are still changes required when performing sort on collection of Branches, or doing simple checks like `if LocalBranch: pass`
   - slower than `NewType`
   - problem with casting `None` as described above
3.  Declare types as classes but each class is a subclass of `string` (solution in this PR)
 pros:
   - `AnyBranch` can accept all (`AnyBranch`, `LocalBranch`, `RemoteBranch`)
   - no massive changes, cause string interface is here 
 cons:
   - if variable is typed to be `str` it can accept all of the new types (`LocalBranch`, `AnyBranch`, `RemoteBranch`, `Commit` )
   - slower than `NewType`
   - problem with casting `None` as described above 

TBD which option to take.

PawelLipski:
I'm in favor of option 3:
* _if variable is typed to be str it can accept all of the new types_ is rather an expected behavior and not a major impediment.
* performance overhead is probably also minuscule, compared to initial import of Python packages + invocation of git commands
* problem with `Commit(None)` can potentially be worked around with `Commit.of` factory method, as suggested in my inline comment

@BartoszSiwinski if you could also take a look

BartoszSiwinski:
Option 3 seems to be the easiest way of getting what we actually want. 

- It may slowdown, but honestly I don't really think of git-machete as something that is performance oriented, so this issue it rather miscellaneous.

- var typed as str can accept all the new types - it's ok. It's supposed to help us create cleaner code and suggest what is expected in branch argument or local variable, so we as developers have to make conscious decision to update it. Wherever we add it it will be improvement, wherever it will be forgotten it will stay as is, so code quality will not take a hit.

I vote for 3.

The only real issue is this None casting new type on None.

Can we have something like:
```
class Commit(str):
   def __init__(string, *args, **kwargs):
      string_ = string. # not to work on the input args in fear of changing something outside of the class
      if string_ is None:
         string_ = "" 
      super().__init__(string_, *args, **kwargs)
```
Or maybe even
```
class Commit(str):
   def __init__(string, *args, **kwargs):
      super().__init__(string if string is not None else "", *args, **kwargs)
```
?

PawelLipski:
> Or maybe even

Or `Commit.of` static factory method, as suggested earlier on this PR!

git_machete/cli.py
```diff
@@ -16,6 +16,7 @@
 from git_machete.exceptions import MacheteException, StopInteraction
 from git_machete.git_operations import GitContext
 from git_machete.utils import fmt, underline, excluding, warn
+from git_machete.custom_types import Commit, LocalBranch, RemoteBranch
```
PawelLipski: I'd say logically these types belong to `git_operations`... unless they needed to be extracted to a separate module for the sake of avoiding dependency cycles?

git_machete/cli.py
```diff
@@ -631,14 +632,14 @@ def launch(orig_args: List[str]) -> None:
...
+                    return LocalBranch.of(re.sub("^[^/]+/", "", remote_branch))
 
                 remote_counterparts_of_local_branches = utils.map_truthy_only(
-                    lambda _branch: git.get_combined_counterpart_for_fetching_of_branch(_branch),
+                    lambda _branch: LocalBranch.of(git.get_combined_counterpart_for_fetching_of_branch(_branch)),
                     git.get_local_branches())
-                qualifying_remote_branches = excluding(git.get_remote_branches(),
-                                                       remote_counterparts_of_local_branches)
+                qualifying_remote_branches: List[RemoteBranch] = excluding(git.get_remote_branches(),
+                                                                           [RemoteBranch.of(b) for b in remote_counterparts_of_local_branches])  # cast to RemoteBranch only to satisfy mypy
```
PawelLipski: I'd say it's not even just to satisfy mypy (in the sense that _an error here is a false positive on the part of mypy_)... I'd say it's pretty justified that mypy raises an error here. Calling `excluding` with a `iterable` being `List[RemoteBranch]` and `s` being `List[str]` would violate the contract, as both `Iterable`s are expected to have the same type param `T`.

Hence, I'd remove the comment

git_machete/client.py
```diff
@@ -39,37 +40,37 @@ class MacheteClient:
...
         return self._managed_branches
 
     @managed_branches.setter
-    def managed_branches(self, val: List[str]) -> None:
+    def managed_branches(self, val: List[LocalBranch]) -> None:
         self._managed_branches = val
 
     @property
-    def up_branch(self) -> Dict[str, Optional[str]]:
+    def up_branch(self) -> Dict[LocalBranch, LocalBranch]:
```
PawelLipski: Hmmm shouldn't `Optional` be retained everywhere where it used to be? 🤔

git_machete/github.py
```diff
@@ -216,7 +217,7 @@ def set_milestone_of_pull_request(org: str, repo: str, number: int, milestone: s
     __fire_github_api_request('PATCH', f'/repos/{org}/{repo}/issues/{number}', token, request_body)
 
 
-def derive_pull_request_by_head(org: str, repo: str, head: str) -> Optional[GitHubPullRequest]:
+def derive_pull_request_by_head(org: str, repo: str, head: AnyBranch) -> Optional[GitHubPullRequest]:
```
PawelLipski: Shouldn't it be `LocalBranch` here?

git_machete/utils.py
```diff
@@ -47,7 +47,7 @@ def tupled(f: Callable[..., T]) -> Callable[[Any], T]:
     return lambda tple: f(*tple)
 
 
-def get_second(pair: Tuple[str, str]) -> str:
+def get_second(pair: Tuple[T, T]) -> T:
```
PawelLipski: The first argument doesn't need to be `T`, can be anything (not sure if Python has a placeholder for type param like `?` in Java... but just `S` can be used and later never referenced - it's almost like a placeholder

git_machete/git_operations.py
```diff
@@ -24,18 +24,18 @@ def __init__(self) -> None:
         self.__fetch_done_for: Set[str] = set()
         self.__config_cached: Optional[Dict[str, str]] = None
         self.__remotes_cached: Optional[List[str]] = None
-        self.__counterparts_for_fetching_cached: Optional[Dict[str, Optional[str]]] = None  # TODO (#110): default dict with None
-        self.__short_commit_sha_by_revision_cached: Dict[str, str] = {}
-        self.__tree_sha_by_commit_sha_cached: Optional[Dict[str, Optional[str]]] = None  # TODO (#110): default dict with None
-        self.__commit_sha_by_revision_cached: Optional[Dict[str, Optional[str]]] = None  # TODO (#110): default dict with None
+        self.__counterparts_for_fetching_cached: Optional[Dict[AnyBranch, Optional[RemoteBranch]]] = None  # TODO (#110): default dict with None
```
PawelLipski: I'd rather see `LocalBranch` as first type param, it's supposed to be a mapping local->remote branch

git_machete/git_operations.py
```diff
@@ -24,18 +24,18 @@ def __init__(self) -> None:
...
-        self.__local_branches_cached: Optional[List[str]] = None
-        self.__remote_branches_cached: Optional[List[str]] = None
-        self.__initial_log_shas_cached: Dict[str, List[str]] = {}
-        self.__remaining_log_shas_cached: Dict[str, List[str]] = {}
-        self.__reflogs_cached: Optional[Dict[str, Optional[List[REFLOG_ENTRY]]]] = None
-        self.__merge_base_cached: Dict[Tuple[str, str], str] = {}
-        self.__contains_equivalent_tree_cached: Dict[Tuple[str, str], bool] = {}
+        self.__local_branches_cached: Optional[List[LocalBranch]] = None
+        self.__remote_branches_cached: Optional[List[RemoteBranch]] = None
+        self.__initial_log_shas_cached: Dict[LocalBranch, List[str]] = {}
```
PawelLipski: Hmmm there's inconsistency on how commit SHAs are kept... here it's `str` but in `__commit_sha_by_revision_cached` it's `Commit`. In fact, I'd be inclined to rename `Commit` to `AnyRevision` (coz this type is NOT commit object, it's just a descriptor like `master` or `13245abd112e`), and add a subtype `FullCommitHash`

PawelLipski: Then e.g. `__initial_log_shas_cached` would be of `Dict[LocalBranch, List[FullCommitHash]]` type,
and `__commit_sha_by_revision_cached` would be `Optional[Dict[AnyRevision, Optional[FullCommitSha]]]

git_machete/custom_types.py
```diff
@@ -0,0 +1,25 @@
...
+        return LocalBranch(value) if value else None
+
+
+class RemoteBranch(AnyBranch):
+    @staticmethod
+    def of(value: str) -> Optional["RemoteBranch"]:
+        return RemoteBranch(value) if value else None
+
+
+class Commit(str):
```
PawelLipski: See the remarks to class hierarchy elsewhere, I'd see `AnyRevision` at the top, with `AnyBranch` and `FullCommitSha` as subtypes. `LocalBranch` and `RemoteBranch` would remain subtypes of `AnyRevision`. `Commit` would be changed into `FullCommitSha`.

PawelLipski (commented):
Okay, not delving deeper so far since there're already a few major comments which may still change the code quite a lot.

BartoszSiwinski:
> > Or maybe even
> 
> Or `Commit.of` static factory method, as suggested earlier on this PR!

Is there any reason why you would like to choose factory `of` over handling this in `__ini__`?
I'm getting the feeling that it will be difficult to name the factory in a non-confusing manner. We could handle the case of `None` being passed as `init arg` in the user code, so we as a new type only accept str or we could handle it in the class itself.

With this `of` fabric it will not be like we have `LocalBranch` type that can store values of `None` or `str`, but the fabric of one type will either return value of its own type or completely different type (`None`), so we will never get `LocalBranch == None` for example.

git_machete/utils.py
rifel123: Ok, let it be `Any` as we do not care about this type at all.

git_machete/custom_types.py
rifel123: Given class structure was applied with `ShortCommitSha`(subtype of `AnyRevision`)  to make distinction between full and short shas.

git_machete/cli.py
```diff
@@ -631,14 +631,14 @@ def launch(orig_args: List[str]) -> None:
...
+                    return LocalBranch.of(re.sub("^[^/]+/", "", remote_branch))
 
                 remote_counterparts_of_local_branches = utils.map_truthy_only(
-                    lambda _branch: git.get_combined_counterpart_for_fetching_of_branch(_branch),
+                    lambda _branch: LocalBranch.of(git.get_combined_counterpart_for_fetching_of_branch(_branch)),
                     git.get_local_branches())
-                qualifying_remote_branches = excluding(git.get_remote_branches(),
-                                                       remote_counterparts_of_local_branches)
+                qualifying_remote_branches: List[RemoteBranch] = excluding(git.get_remote_branches(),
+                                                                           [RemoteBranch.of(b) for b in remote_counterparts_of_local_branches])
```
BartoszSiwinski:
```suggestion
                                                                           {RemoteBranch.of(b) for b in remote_counterparts_of_local_branches})
```
Maybe `set()` instead of `list()` in here? This way we don't have to iterate through whole list with each remote to be compared and we're avoiding checking multiple times `None` that can pop in the list

git_machete/tests/functional/test_machete.py
```diff
@@ -37,8 +38,8 @@ def get_current_commit_hash() -> str:
 
 
 def mock_fetch_ref(cls: Any, remote: str, ref: str) -> None:
-    branch: str = ref[ref.index(':') + 1:]
-    git.create_branch(branch, get_current_commit_hash())
+    branch: LocalBranch = LocalBranch.of(ref[ref.index(':') + 1:])
+    git.create_branch(branch, RemoteBranch.of(get_current_commit_hash()))
```
BartoszSiwinski:
```suggestion
    git.create_branch(branch, AnyRevision(get_current_commit_hash()))
```
or (preferred)
```suggestion
    git.create_branch(branch, FullCommitSha(get_current_commit_hash()))
```

git_machete/tests/functional/test_machete.py
```diff
@@ -2041,7 +2042,7 @@ def test_github_create_pr(self) -> None:
         expected_error_message = "Pull request for branch hotfix/add-trigger is already created under link www.github.com!\nPR details: PR #6 by github_user: hotfix/add-trigger -> master"
         machete_client.read_definition_file()
         with self.assertRaises(MacheteException) as e:
-            machete_client.create_github_pr(head='hotfix/add-trigger', opt_draft=False, opt_onto=None)
+            machete_client.create_github_pr(head=LocalBranch.of('hotfix/add-trigger'), opt_draft=False, opt_onto=None)
```
BartoszSiwinski:
```suggestion
            machete_client.create_github_pr(head=LocalBranch('hotfix/add-trigger'), opt_draft=False, opt_onto=None)
```
In this case we don't have to worry about `None` option

git_machete/tests/functional/test_machete.py
```diff
@@ -2058,15 +2059,15 @@ def test_github_create_pr(self) -> None:
         machete_client.read_definition_file()
         expected_error_message = "All commits in `testing/endpoints` branch  are already included in `develop` branch.\nCannot create pull request."
         with self.assertRaises(MacheteException) as e:
-            machete_client.create_github_pr(head='testing/endpoints', opt_draft=False, opt_onto=None)
+            machete_client.create_github_pr(head=LocalBranch.of('testing/endpoints'), opt_draft=False, opt_onto=None)
```
BartoszSiwinski:
```suggestion
            machete_client.create_github_pr(head=LocalBranch('testing/endpoints'), opt_draft=False, opt_onto=None)
```

git_machete/tests/functional/test_machete.py
```diff
@@ -2058,15 +2059,15 @@ def test_github_create_pr(self) -> None:
...
         if e:
             self.assertEqual(e.exception.parameter, expected_error_message,
                              'Verify that expected error message has appeared when head branch is equal or ancestor of base branch.')
 
         self.repo_sandbox.check_out('develop')
         expected_error_message = "Branch `develop` does not have a parent branch (it is a root), base branch for the PR cannot be established."
         with self.assertRaises(MacheteException) as e:
-            machete_client.create_github_pr(head='develop', opt_draft=False, opt_onto=None)
+            machete_client.create_github_pr(head=LocalBranch.of('develop'), opt_draft=False, opt_onto=None)
```
BartoszSiwinski:
```suggestion
            machete_client.create_github_pr(head=LocalBranch('develop'), opt_draft=False, opt_onto=None)
```

rifel123: There are a lot of places where using `.of` is not necessary at all. But it was written this way to unify casting to New Types across all files. Also imagine someone in the future changes this piece of code that you actually might receive `None` here, it would cause really hard to track down bug.

PawelLipski: Yeah, let's leave `.of` even if it's redundant sometimes, for the sake of consistency

git_machete/cli.py
```diff
@@ -549,7 +549,7 @@ def launch(orig_args: List[str]) -> None:
                     use_overrides=False,
                     opt_no_detect_squash_merges=cli_opts.opt_no_detect_squash_merges))
             elif cli_opts.opt_override_to:
-                machete_client.set_fork_point_override(branch, cli_opts.opt_override_to)
+                machete_client.set_fork_point_override(branch, FullCommitSha.of(cli_opts.opt_override_to))
```
PawelLipski: Hmmm the problem is that `opt_override_to` doesn't need to be full commit SHA (in fact, it rarely will in practice), it might be a branch name or shortened hash

PawelLipski: So, probably needs to be changed `AnyRevision` here + anywhere where needed, to propagate the change

git_machete/cli.py
```diff
@@ -631,14 +631,14 @@ def launch(orig_args: List[str]) -> None:
             machete_client.read_definition_file()
             res = []
             if category == "addable":
-                def strip_first_fragment(remote_branch: str) -> str:
-                    return re.sub("^[^/]+/", "", remote_branch)
+                def strip_first_fragment(remote_branch: RemoteBranch) -> LocalBranch:
```
PawelLipski: Maybe rename to `strip_remote_name` for clarity, while you're here

git_machete/options.py
```diff
@@ -1,20 +1,21 @@
...
+        self.opt_branch: Optional[LocalBranch] = None
         self.opt_checked_out_since: Optional[str] = None
         self.opt_color: str = "auto"
         self.opt_debug: bool = False
-        self.opt_down_fork_point: Optional[str] = None
+        self.opt_down_fork_point: Optional[FullCommitSha] = None
         self.opt_draft: bool = False
         self.opt_fetch: bool = False
-        self.opt_fork_point: Optional[str] = None
+        self.opt_fork_point: Optional[FullCommitSha] = None
```
PawelLipski: Yeah, as mentioned elsewhere, both `opt_down_fork_point` and this one should probably be `AnyRevision` instead

git_machete/client.py
```diff
@@ -94,15 +94,15 @@ def read_definition_file(self, verify_branches: bool = True) -> None:
...
+        invalid_branches: List[LocalBranch] = []
         for index, line in enumerate(lines):
             if line == "" or line.isspace():
                 continue
             prefix = "".join(itertools.takewhile(str.isspace, line))
             if prefix and not self.__indent:
                 self.__indent = prefix
 
-            b_a: List[str] = line.strip().split(" ", 1)
+            b_a: List[LocalBranch] = [LocalBranch.of(entry) for entry in line.strip().split(" ", 1)]
```
PawelLipski: Pls rename to `branch_and_maybe_annotation` while you're here

git_machete/client.py
```diff
@@ -221,21 +221,21 @@ def add(self, *, branch: str, opt_onto: Optional[str], opt_as_root: bool, opt_ye
...
+            remote_branch: Optional[RemoteBranch] = self.__git.get_sole_remote_branch(branch)
             if remote_branch:
                 common_line = (
                     f"A local branch `{branch}` does not exist, but a remote "
                     f"branch `{remote_branch}` exists.\n")
                 msg = common_line + f"Check out `{branch}` locally?" + get_pretty_choices('y', 'N')
                 opt_yes_msg = common_line + f"Checking out `{branch}` locally..."
                 if self.ask_if(msg, opt_yes_msg, opt_yes=opt_yes) in ('y', 'yes'):
-                    self.__git.create_branch(branch, f"refs/remotes/{remote_branch}")
+                    self.__git.create_branch(branch, RemoteBranch.of(f"refs/remotes/{remote_branch}"))
```
PawelLipski: Hmmm if I'd rather introduce the convention that `RemoteBranch` does NOT include the `refs/remotes/` prefix, same for `refs/heads/` and `LocalBranch`

PawelLipski: Of course the prefix still needs to be used in certain contexts for unambiguity (esp. given that git allows for having a branch and tag of the same name). Maybe a method `full_name` should be added in `LocalBranch` and `RemoteBranch`?

git_machete/client.py
```diff
@@ -221,21 +221,21 @@ def add(self, *, branch: str, opt_onto: Optional[str], opt_as_root: bool, opt_ye
...
                 else:
                     return
                 # Not dealing with `onto` here. If it hasn't been explicitly
                 # specified via `--onto`, we'll try to infer it now.
             else:
-                out_of = f"refs/heads/{opt_onto}" if opt_onto else "HEAD"
+                out_of = RemoteBranch.of(f"refs/heads/{opt_onto}") if opt_onto else RemoteBranch.of("HEAD")
```
PawelLipski: Whooops `RemoteBranch.of("HEAD")` is wrong... `"HEAD"` can only be reasonably passed to `AnyRevision`

PawelLipski: Maybe `HEAD` should even be a constant in `git_operations.py`? like `HEAD = AnyRevision.of("HEAD")`

PawelLipski: `RemoteBranch.of(f"refs/heads/...")` is clearly wrong as well ‼️

git_machete/client.py
```diff
@@ -324,7 +324,7 @@ def update(
                     "Branch `%s` not found in the tree of branch dependencies. "
                     "Rebasing onto the inferred upstream `%s`..."))
             self.__git.rebase(
-                f"refs/heads/{onto_branch}",
+                LocalBranch.of(f"refs/heads/{onto_branch}"),
```
PawelLipski: Ditto, passing prefix is questionable

git_machete/client.py
```diff
@@ -767,7 +767,7 @@ def traverse(
                             return
                     else:
                         self.__git.rebase(
-                            f"refs/heads/{upstream}", self.fork_point(
+                            AnyBranch.of(f"refs/heads/{upstream}"), self.fork_point(
```
PawelLipski: Can it be `LocalBranch`? let's use a more specific type wherever possible

git_machete/client.py
```diff
@@ -873,12 +873,12 @@ def prefix_dfs(u_: str, accumulated_path_: List[Optional[str]]) -> None:
             prefix_dfs(upstream, accumulated_path_=[])
 
         out = io.StringIO()
-        edge_color: Dict[str, str] = {}
-        fp_sha_cached: Dict[str, Optional[str]] = {}  # TODO (#110): default dict with None
-        fp_branches_cached: Dict[str, List[BRANCH_DEF]] = {}
+        edge_color: Dict[LocalBranch, str] = {}
+        fp_sha_cached: Dict[Union[LocalBranch, FullCommitSha], Optional[Union[LocalBranch, FullCommitSha]]] = {}  # TODO (#110): default dict with None
```
PawelLipski: 🤔 at first glance I'd say it'd be just `Dict[LocalBranch, Optional[FullCommitSha]]`... so, where is `FullCommitSha` used a key, and where is `LocalBranch` used as a value?

git_machete/git_operations.py
```diff
@@ -12,7 +12,43 @@
...
+        return LocalBranch(value) if value else None
+
+
+class RemoteBranch(AnyBranch):
+    @staticmethod
+    def of(value: str) -> Optional["RemoteBranch"]:
+        return RemoteBranch(value) if value else None
+
+
+class FullCommitSha(AnyRevision):
```
PawelLipski: Nit: I'd use `Hash` instead of `Sha`... no strong feelings tho

git_machete/client.py
PawelLipski: Or maybe instead... splitting `LocalBranch` into `LocalBranchName` and `LocalBranchFullName`? maybe an overkill tho

git_machete/client.py
```diff
@@ -912,7 +912,7 @@ def fp_sha(branch_: str) -> Optional[str]:
         hook_path = self.__git.get_hook_path("machete-status-branch")
         hook_executable = self.__git.check_hook_executable(hook_path)
 
-        def print_line_prefix(b_: str, suffix: str) -> None:
+        def print_line_prefix(b_: LocalBranch, suffix: str) -> None:
```
PawelLipski: Pls expand all `b_`s into `branch_` while you're here

git_machete/client.py
```diff
@@ -1126,67 +1126,67 @@ def __fork_point_and_containing_branch_defs(self, branch: str, use_overrides: bo
...
 
-    def log(self, branch: str) -> None:
-        full_branch_name = f"refs/heads{branch}"
+    def log(self, branch: LocalBranch) -> None:
+        full_branch_name = LocalBranch.of(f"refs/heads{branch}")
         forkpoint = self.fork_point(branch, use_overrides=True, opt_no_detect_squash_merges=False)
         self.__git.display_branch_history_from_forkpoint(full_branch_name, forkpoint)
 
-    def down(self, branch: str, pick_mode: bool) -> str:
+    def down(self, branch: LocalBranch, pick_mode: bool) -> str:
```
PawelLipski: Maybe return `List[LocalBranch]` here (possibly a single-element list), and only `join` when needed in cli.py

PawelLipski: (For consistency with the following methods)

git_machete/client.py
```diff
@@ -1237,39 +1237,40 @@ def up(self, branch: str, prompt_if_inferred_msg: Optional[str],
                     f"Branch `{branch}` not found in the tree of branch "
                     f"dependencies and its upstream could not be inferred")
 
-    def get_slidable_branches(self) -> List[str]:
+    def get_slidable_branches(self) -> List[LocalBranch]:
         return [branch for branch in self.managed_branches if branch in self.up_branch]
 
-    def slidable_after(self, branch: str) -> List[str]:
+    def slidable_after(self, branch: LocalBranch) -> List[LocalBranch]:
```
PawelLipski: While you're here, rename to `get_slidable_after`

git_machete/client.py
```diff
@@ -1284,11 +1285,11 @@ def squash(self, *, current_branch: str, opt_fork_point: str) -> None:
             return
 
         earliest_sha, earliest_short_sha, earliest_subject = commits[0]
-        earliest_full_body = self.__git.get_commit_information("raw body", earliest_sha).strip()
+        earliest_full_body = self.__git.get_commit_information("raw body", FullCommitSha(earliest_sha)).strip()
```
PawelLipski: `.of`? pls Ctrl+Shift+F the entire project for outliers like this one

git_machete/client.py
```diff
@@ -1519,23 +1520,23 @@ def __infer_upstream(self, branch: str, condition: Callable[[str], bool] = lambd
         return None
 
     @staticmethod
-    def config_key_for_override_fork_point_to(branch: str) -> str:
+    def config_key_for_override_fork_point_to(branch: AnyBranch) -> str:
```
PawelLipski: I'd rather expect `LocalBranch`. It doesn't ever make sense to override fork point (or ever compute fork point, for that matter) for remote branches, as of the current git-machete domain model 🤔

git_machete/client.py
```diff
@@ -1519,23 +1520,23 @@ def __infer_upstream(self, branch: str, condition: Callable[[str], bool] = lambd
...
         return f"machete.overrideForkPoint.{branch}.whileDescendantOf"
 
     # Also includes config that is incomplete (only one entry out of two) or otherwise invalid.
-    def has_any_fork_point_override_config(self, branch: str) -> bool:
+    def has_any_fork_point_override_config(self, branch: AnyBranch) -> bool:
         return (self.__git.get_config_attr_or_none(self.config_key_for_override_fork_point_to(branch)) or
                 self.__git.get_config_attr_or_none(self.config_key_for_override_fork_point_while_descendant_of(branch))) is not None
 
-    def __get_fork_point_override_data(self, branch: str) -> Optional[Tuple[str, str]]:
+    def __get_fork_point_override_data(self, branch: AnyBranch) -> Optional[Tuple[FullCommitSha, FullCommitSha]]:
```
PawelLipski: Pls open a task to create a dedicated type here (`ForkPointOverrideData`) instead of a tuple

git_machete/client.py
```diff
@@ -1586,19 +1587,19 @@ def __get_overridden_fork_point(self, branch: str) -> Optional[str]:
...
+    def set_fork_point_override(self, branch: AnyBranch, to_revision: AnyRevision) -> None:
         if branch not in self.__git.get_local_branches():
             raise MacheteException(f"`{branch}` is not a local branch")
         to_sha = self.__git.get_commit_sha_by_revision(to_revision, prefix="")
         if not to_sha:
             raise MacheteException(f"Cannot find revision {to_revision}")
         if not self.__git.is_ancestor_or_equal(to_sha, branch, earlier_prefix=""):
             raise MacheteException(
-                f"Cannot override fork point: {self.__git.get_revision_repr(to_revision)} is not an ancestor of {branch}")
+                f"Cannot override fork point: {self.__git.get_revision_repr(FullCommitSha.of(to_revision))} is not an ancestor of {branch}")
```
PawelLipski: Hmmm you can't pass an `AnyRevision` to `FullCommitSha.of`, you don't have a guarantee that this revision is a full hash indeed ‼️

PawelLipski: That's not even needed, since `get_revision_repr` expects just `AnyRevision`

git_machete/client.py
```diff
@@ -1610,7 +1611,7 @@ def set_fork_point_override(self, branch: str, to_revision: str) -> None:
         sys.stdout.write(
             fmt(
                 f"Fork point for <b>{branch}</b> is overridden to <b>"
-                f"{self.__git.get_revision_repr(to_revision)}</b>.\n",
+                f"{self.__git.get_revision_repr(FullCommitSha.of(to_revision))}</b>.\n",
```
PawelLipski: Ditto, wrapping unnecessary

git_machete/client.py
```diff
@@ -1782,15 +1783,15 @@ def handle_untracked_branch(
...
+    def is_merged_to(self, branch: LocalBranch, target: AnyBranch, *, opt_no_detect_squash_merges: bool) -> bool:
         if self.__git.is_ancestor_or_equal(branch, target):
             # If branch is ancestor of or equal to the target, we need to distinguish between the
             # case of branch being "recently" created from the target and the case of
             # branch being fast-forward-merged to the target.
             # The applied heuristics is to check if the filtered reflog of the branch
             # (reflog stripped of trivial events like branch creation, reset etc.)
             # is non-empty.
-            return bool(self.filtered_reflog(branch, prefix="refs/heads/"))
+            return bool(self.filtered_reflog(str(branch), prefix="refs/heads/"))
```
PawelLipski: Hmmm `filtered_reflog` shouldn't accept `str`, rather `LocalBranch`

PawelLipski: Oh! there's a problem with the prefix hmmmm

PawelLipski: So maybe `filtered_reflog` should accept a single param of `AnyBranch` type, and use `full_name` method of this single param instead of an extra `prefix` param

git_machete/client.py
```diff
@@ -1988,7 +1989,7 @@ def create_github_pr(
 
         ok_str = ' <green><b>OK</b></green>'
         print(fmt(f'Creating a {"draft " if opt_draft else ""}PR from `{head}` to `{base}`...'), end='', flush=True)
-        pr: GitHubPullRequest = create_pull_request(org, repo, head=head, base=base, title=title,
+        pr: GitHubPullRequest = create_pull_request(org, repo, head=str(head), base=str(base), title=title,
```
PawelLipski: Hmmm actually... isn't it possible to just pass a `LocalBranch` where a `str` is expected (with no `str(...)` cast)? after all, a `LocalBranch` _is_ a `str`, right?

git_machete/git_operations.py
```diff
@@ -142,7 +178,7 @@ def get_config_attr_or_none(self, key: str) -> Optional[str]:
         self.__ensure_config_loaded()
         return self.__config_cached.get(key.lower())
 
-    def set_config_attr(self, key: str, value: str) -> None:
+    def set_config_attr(self, key: str, value: FullCommitSha) -> None:
```
PawelLipski: Hmmm I don't think it should be `FullCommitSha` for `value`, as in general git config entries can keep any strings, not only commit hashes (even though we might be using this method only with commit hashes as for now)

git_machete/git_operations.py
```diff
@@ -197,93 +233,93 @@ def push(self, remote: str, branch: str, force_with_lease: bool = False) -> None
...
         # Without ^{commit}, 'git rev-parse --verify' will not only accept references to other kinds of objects (like trees and blobs),
         # but just echo the argument (and exit successfully) even if the argument doesn't match anything in the object store.
         try:
-            return self._popen_git("rev-parse", "--verify", "--quiet", revision + "^{commit}").rstrip()
+            return FullCommitSha.of(self._popen_git("rev-parse", "--verify", "--quiet", revision + "^{commit}").rstrip())
         except MacheteException:
             return None
 
-    def get_commit_sha_by_revision(self, revision: str, prefix: str = "refs/heads/") -> Optional[str]:
+    def get_commit_sha_by_revision(self, revision: AnyRevision, prefix: str = "refs/heads/") -> Optional[FullCommitSha]:
```
PawelLipski: Whoops problematic place again, since a non-empty `prefix` can't be prepended to just any revision, only to a branch... again, `full_name` method (that would maybe return a `{Local,Remote}BranchFullName` type?) for the rescue

git_machete/git_operations.py
```diff
@@ -197,93 +233,93 @@ def push(self, remote: str, branch: str, force_with_lease: bool = False) -> None
...
+            return self.get_commit_sha_by_revision(FullCommitSha.of(revision), prefix)
 
-    def get_committer_unix_timestamp_by_revision(self, revision: str, prefix: str = "refs/heads/") -> int:
+    def get_committer_unix_timestamp_by_revision(self, revision: AnyBranch, prefix: str = "refs/heads/") -> int:
         if self.__committer_unix_timestamp_by_revision_cached is None:
             self.__load_branches()
         return self.__committer_unix_timestamp_by_revision_cached.get(prefix + revision, 0)
 
-    def get_inferred_remote_for_fetching_of_branch(self, branch: str) -> Optional[str]:
+    def get_inferred_remote_for_fetching_of_branch(self, branch: AnyBranch) -> Optional[str]:
```
PawelLipski: I'd expect `branch` to be a `LocalBranch`; it doesn't make sense to call this method for remote branches, since their remote can be basically extracted from their name, like `origin/foo` => `origin`

git_machete/git_operations.py
```diff
@@ -197,93 +233,93 @@ def push(self, remote: str, branch: str, force_with_lease: bool = False) -> None
...
         # Since many people don't use '--set-upstream' flag of 'push', we try to infer the remote instead.
         for remote in self.get_remotes():
             if f"{remote}/{branch}" in self.get_remote_branches():
                 return remote
         return None
 
-    def get_strict_remote_for_fetching_of_branch(self, branch: str) -> Optional[str]:
+    def get_strict_remote_for_fetching_of_branch(self, branch: AnyBranch) -> Optional[str]:
```
PawelLipski: Ditto here and below

git_machete/git_operations.py
```diff
@@ -329,10 +365,10 @@ def __load_branches(self) -> None:
             if len(values) != 4:
                 continue  # invalid, shouldn't happen
             branch, commit_sha, tree_sha, committer_unix_timestamp_and_time_zone = values
-            b_stripped = re.sub("^refs/remotes/", "", branch)
-            self.__remote_branches_cached += [b_stripped]
-            self.__commit_sha_by_revision_cached[branch] = commit_sha
-            self.__tree_sha_by_commit_sha_cached[commit_sha] = tree_sha
+            b_stripped = AnyBranch.of(re.sub("^refs/remotes/", "", branch))
```
PawelLipski: Nit: `AnyBranch.of` feels redundant here

git_machete/git_operations.py
```diff
@@ -329,10 +365,10 @@ def __load_branches(self) -> None:
...
+            self.__remote_branches_cached += [RemoteBranch.of(b_stripped)]
+            self.__commit_sha_by_revision_cached[FullCommitSha.of(branch)] = FullCommitSha.of(commit_sha)
```
PawelLipski: Whoops `branch` is not a `FullCommitSha`

PawelLipski: Maybe `FullCommitSha.of` (and maybe other factory methods as well) should do some basic verification? like, `FullCommitSha.of`'s input should have length of 40. This would help spotting mistakes like this one much easier (at the expense of higher risk of runtime errors if a bug isn't covered by tests)

git_machete/git_operations.py
```diff
@@ -342,32 +378,32 @@ def __load_branches(self) -> None:
...
                 continue  # invalid, shouldn't happen
             branch, commit_sha, tree_sha, committer_unix_timestamp_and_time_zone, fetch_counterpart = values
-            b_stripped = re.sub("^refs/heads/", "", branch)
+            b_stripped = LocalBranch.of(re.sub("^refs/heads/", "", branch))
             fetch_counterpart_stripped = re.sub("^refs/remotes/", "", fetch_counterpart)
-            self.__local_branches_cached += [b_stripped]
-            self.__commit_sha_by_revision_cached[branch] = commit_sha
-            self.__tree_sha_by_commit_sha_cached[commit_sha] = tree_sha
+            self.__local_branches_cached += [LocalBranch.of(b_stripped)]
+            self.__commit_sha_by_revision_cached[FullCommitSha.of(branch)] = FullCommitSha.of(commit_sha)
```
PawelLipski: Ditto, incorrect key type

git_machete/git_operations.py
```diff
@@ -388,10 +424,10 @@ def __load_all_reflogs(self) -> None:
                 continue
             branch, pos = branch_and_pos
             if branch not in self.__reflogs_cached:
-                self.__reflogs_cached[branch] = []
-            self.__reflogs_cached[branch] += [(sha, subject)]
+                self.__reflogs_cached[LocalBranch.of(branch)] = []
```
PawelLipski: As you can see from lines 412-415, reflogs are fetched for both local **and remote** branches... so the type of `__reflogs_cached` should be `AnyBranch`

git_machete/git_operations.py
```diff
@@ -388,10 +424,10 @@ def __load_all_reflogs(self) -> None:
...
+            self.__reflogs_cached[LocalBranch.of(branch)] += [(FullCommitSha.of(sha), subject)]
 
-    def get_reflog(self, branch: str) -> List[REFLOG_ENTRY]:
+    def get_reflog(self, branch: LocalBranch) -> List[REFLOG_ENTRY]:
```
PawelLipski: Same here, probably `AnyBranch` (also, `filtered_reflog` to be tidied up)

git_machete/git_operations.py
```diff
@@ -489,16 +525,16 @@ def expect_no_operation_in_progress(self) -> None:
...
 
-    def get_current_branch(self) -> str:
+    def get_current_branch(self) -> LocalBranch:
         result = self.get_current_branch_or_none()
         if not result:
             raise MacheteException("Not currently on any branch")
         return result
 
-    def __get_merge_base(self, sha1: str, sha2: str) -> str:
+    def __get_merge_base(self, sha1: AnyRevision, sha2: AnyRevision) -> FullCommitSha:
```
PawelLipski: Params can be `FullCommitHash`, judging by the invocation

git_machete/git_operations.py
```diff
@@ -509,15 +545,15 @@ def __get_merge_base(self, sha1: str, sha2: str) -> str:
             #   then there is exactly one merge-base - the ancestor,
             # * if neither of sha1, sha2 is an ancestor of another,
             #   then none of the (possibly more than one) merge-bases is equal to either of sha1/sha2 anyway.
-            self.__merge_base_cached[sha1, sha2] = self._popen_git("merge-base", sha1, sha2).rstrip()
+            self.__merge_base_cached[sha1, sha2] = FullCommitSha.of(self._popen_git("merge-base", sha1, sha2).rstrip())
```
PawelLipski: ...and hence the key of `__merge_base_cached` can be a pair of commit hashes

git_machete/git_operations.py
```diff
@@ -670,18 +706,18 @@ def get_all_lines_fixed() -> Iterator[str]:
         else:
             do_rebase()
 
-    def rebase_onto_ancestor_commit(self, branch: str, ancestor_commit: str, opt_no_interactive_rebase: bool) -> None:
+    def rebase_onto_ancestor_commit(self, branch: LocalBranch, ancestor_commit: FullCommitSha, opt_no_interactive_rebase: bool) -> None:
         self.rebase(ancestor_commit, ancestor_commit, branch, opt_no_interactive_rebase)
 
-    def get_commits_between(self, earliest_exclusive: str, latest_inclusive: str) -> List[Tuple[str, str, str]]:
+    def get_commits_between(self, earliest_exclusive: AnyRevision, latest_inclusive: LocalBranch) -> List[Tuple[str, str, str]]:
```
PawelLipski: Nit: both params can be `AnyRevision` in fact (no particular reason to expect `latest_inclusive` to be a local branch)

git_machete/git_operations.py
```diff
@@ -730,7 +766,7 @@ def get_latest_checkout_timestamps(self) -> Dict[str, int]:  # TODO (#110): defa
                     result[to_branch] = int(match.group(1))
         return result
 
-    def get_commit_information(self, information: str, commit: str = '') -> str:
+    def get_commit_information(self, information: str, commit: FullCommitSha = None) -> str:
         if information not in GIT_FORMAT_PATTERNS:
             raise MacheteException(
                 f"Retriving {information} from commit is not supported by project"
```
PawelLipski: `Retrieving`

git_machete/git_operations.py
```diff
@@ -730,7 +766,7 @@ def get_latest_checkout_timestamps(self) -> Dict[str, int]:  # TODO (#110): defa
                     result[to_branch] = int(match.group(1))
         return result
 
-    def get_commit_information(self, information: str, commit: str = '') -> str:
+    def get_commit_information(self, information: str, commit: FullCommitSha = None) -> str:
```
PawelLipski: `commit` should be `Optional[FullCommitSha]` (not enforced tho, until https://github.com/VirtusLab/git-machete/issues/118 is done)

PawelLipski: ...but can be relaxed to `Optional[AnyRevision]`, on the other hand

git_machete/git_operations.py
```diff
@@ -742,21 +778,21 @@ def get_commit_information(self, information: str, commit: str = '') -> str:
             params.append(commit)
         return self._popen_git(*params)
 
-    def display_branch_history_from_forkpoint(self, branch: str, forkpoint: str) -> int:
+    def display_branch_history_from_forkpoint(self, branch: LocalBranch, forkpoint: FullCommitSha) -> int:
```
PawelLipski: `fork_point` (and everywhere else), for consistency

git_machete/git_operations.py
```diff
@@ -742,21 +778,21 @@ def get_commit_information(self, information: str, commit: str = '') -> str:
...
         return self._run_git("log", f"^{forkpoint}", branch)
 
     def squash_commits_with_msg_and_new_env(
-            self, fork_commit: str, msg: str, env: Dict[str, str]) -> str:
+            self, fork_commit: FullCommitSha, msg: str, env: Dict[str, str]) -> FullCommitSha:
```
PawelLipski: Replace `fork_commit` with `fork_point` everywhere, for consistency

git_machete/git_operations.py
```diff
@@ -742,21 +778,21 @@ def get_commit_information(self, information: str, commit: str = '') -> str:
             params.append(commit)
         return self._popen_git(*params)
 
-    def display_branch_history_from_forkpoint(self, branch: str, forkpoint: str) -> int:
+    def display_branch_history_from_forkpoint(self, branch: LocalBranch, forkpoint: FullCommitSha) -> int:
         return self._run_git("log", f"^{forkpoint}", branch)
 
     def squash_commits_with_msg_and_new_env(
```
```suggestion
    def commit_tree_with_given_parent_and_message_and_env(
```

git_machete/git_operations.py
PawelLipski: Here the param should rather be called `parent_commit` tho

git_machete/git_operations.py
```diff
@@ -742,21 +778,21 @@ def get_commit_information(self, information: str, commit: str = '') -> str:
...
+            "commit-tree", "HEAD^{tree}", "-p", fork_commit, "-m", msg, env=env))
 
-    def delete_branch(self, branch_name: str, force: bool = False) -> int:
+    def delete_branch(self, branch_name: LocalBranch, force: bool = False) -> int:
         self.flush_caches()
         delete_option = '-D' if force else '-d'
         return self._run_git("branch", delete_option, branch_name)
 
-    def display_diff(self, forkpoint: str, format_with_stat: bool, branch: str = None) -> int:
+    def display_diff(self, forkpoint: Union[LocalBranch, FullCommitSha], format_with_stat: bool, branch: LocalBranch = None) -> int:
```
PawelLipski: `fork_point` can be `AnyRevision`?

git_machete/git_operations.py
```diff
@@ -767,12 +803,12 @@ def display_diff(self, forkpoint: str, format_with_stat: bool, branch: str = Non
 
         return self._run_git(*params)
 
-    def update_head_ref_to_new_hash_with_msg(self, hash: str, msg: str) -> int:
+    def update_head_ref_to_new_hash_with_msg(self, hash: FullCommitSha, msg: str) -> int:
```
```suggestion
    def update_head_ref_to_new_hash_with_reflog_subject(self, hash: FullCommitSha, reflog_subject: str) -> int:
```

git_machete/git_operations.py
```diff
@@ -767,12 +803,12 @@ def display_diff(self, forkpoint: str, format_with_stat: bool, branch: str = Non
...
         self.flush_caches()
         return self._run_git("update-ref", "HEAD", hash, "-m", msg)
 
     def check_that_forkpoint_is_ancestor_or_equal_to_tip_of_branch(
```
PawelLipski: I'd move this method to client.py. git_operations.py shouldn't know anything about fork point (it's not strictly a machete-only term, since there's `git rebase --fork-point`... but we never use fork point in this git-native sense, only in the machete-redefined sense)

git_machete/client.py
```diff
@@ -2163,7 +2164,7 @@ def __sync_before_creating_pr(self, *, opt_onto: Optional[str], opt_yes: bool) -
                     "`discover` or `edit` or agree on adding the branch to the "
                     "definition file during the execution of `github create-pr` command.")
 
-        up_branch: Optional[str] = self.up_branch.get(current_branch)
+        up_branch: Optional[LocalBranch] = self.up_branch.get(current_branch)
         if not up_branch:
             raise MacheteException(
                 f'Branch `{current_branch}` does not have a parent branch (it is a root), '
```
PawelLipski, line 2175: While you're here, remove double space from `branch  are` below (line 2175)

git_machete/client.py
```diff
@@ -1396,31 +1397,31 @@ def sync_annotations_to_github_prs(self) -> None:
...
                     anno += f' ({pr.user})'
-                upstream: Optional[str] = self.up_branch.get(pr.head)
+                upstream: Optional[LocalBranch] = self.up_branch.get(LocalBranch.of(pr.head))
                 if pr.base != upstream:
                     warn(f'branch `{pr.head}` has a different base in PR #{pr.number} (`{pr.base}`) '
                          f'than in machete file (`{upstream or "<none, is a root>"}`)')
                     anno += f" WRONG PR BASE or MACHETE PARENT? PR has '{pr.base}'"
-                if self.__annotations.get(pr.head) != anno:
+                if self.__annotations.get(LocalBranch.of(pr.head)) != anno:
                     print(fmt(f'Annotating <b>{pr.head}</b> as `{anno}`'))
```
PawelLipski: While you're here:
```suggestion
                    print(fmt(f'Annotating `{pr.head}` as `{anno}`'))
```
(for more consistent formatting)

git_machete/client.py
```diff
@@ -1130,36 +1130,35 @@ def fork_point(
...
             branch if branch else self.__git.get_current_branch(),
             use_overrides=True, opt_no_detect_squash_merges=False)
         self.__git.display_diff(
             branch=branch,
-            forkpoint=fp,
+            fork_point=fp,
             format_with_stat=opt_stat)
 
     def log(self, branch: LocalBranch) -> None:
-        full_branch_name = LocalBranch.of(f"refs/heads{branch}")
```
PawelLipski: This fixes #314 btw (although I'm not closing that issue since adding a regression test would still be a good idea)

git_machete/client.py
```diff
@@ -1409,51 +1408,51 @@ def __sync_annotations_to_definition_file(self, prs: List[GitHubPullRequest], cu
...
-            return self.prev_branch(branch)
+            return [self.prev_branch(branch)]
         elif param in ("r", "root"):
-            return self.root_branch(branch, if_unmanaged=PICK_FIRST_ROOT)
+            return [self.root_branch(branch, if_unmanaged=PICK_FIRST_ROOT)]
         elif param in ("u", "up"):
-            return self.up(branch, prompt_if_inferred_msg=None, prompt_if_inferred_yes_opt_msg=None)
+            return [self.up(branch, prompt_if_inferred_msg=None, prompt_if_inferred_yes_opt_msg=None)]
         else:
             raise MacheteException(f"Invalid direction: `{param}` expected: {allowed_directions(allow_current)}")
```
```suggestion
            raise MacheteException(f"Invalid direction: `{param}`; expected: {allowed_directions(allow_current)}")
```

git_machete/client.py
```diff
@@ -1834,6 +1833,16 @@ def flush_caches(self) -> None:
         self.__branch_defs_by_sha_in_reflog = None
         self.__git.flush_caches()
 
+    def check_that_forkpoint_is_ancestor_or_equal_to_tip_of_branch(
+            self, forkpoint_sha: AnyRevision, branch: AnyBranch) -> None:
+        if not self.__git.is_ancestor_or_equal(
+                earlier_revision=forkpoint_sha,
+                earlier_prefix='',
```
PawelLipski: Hmmm I'd say prefixes shouldn't be used &nbsp; **a n y w h e r e** &nbsp; now that `full_name` is in place, pls rework if so

git_machete/git_operations.py
```diff
@@ -15,40 +15,83 @@
...
+                return ShortCommitHash(value)
+            else:
+                raise TypeError(f'ShortCommitHash requires length of 7. Given value: {value}, has length: {len(value)}.')
+        return None
+
+    def full_name(self) -> "ShortCommitHash":
+        return self
+
+
+REFLOG_ENTRY = Tuple[FullCommitHash, str]
```
PawelLipski: While you're here, AFAICS the convention for type aliases is to use CamelCase: https://stackoverflow.com/a/33045252/2750812, pls rename this one (and also `BRANCH_DEF` in client.py)

git_machete/git_operations.py
```diff
@@ -15,40 +15,83 @@
...
+
+HEAD = AnyRevision.of("HEAD")
```
PawelLipski, line 148: `HEAD` is okay in all caps coz it's a constant, not a type alias

git_machete/git_operations.py
```diff
@@ -15,40 +15,83 @@
...
     @staticmethod
-    def of(value: str) -> Optional["ShortCommitSha"]:
-        return ShortCommitSha(value) if value else None
+    def of(value: str) -> Optional["FullCommitHash"]:
+        value = value.strip()
+        if value:
+            if len(value) == 40:
+                return FullCommitHash(value)
+            else:
+                raise TypeError(f'FullCommitHash requires length of 40. Given value: {value}, has length: {len(value)}.')
```
PawelLipski: TBD if these checks stay or go... did they happen to capture any mistake?

git_machete/git_operations.py
```diff
@@ -15,40 +15,83 @@
...
 class AnyBranch(AnyRevision):
     @staticmethod
     def of(value: str) -> Optional["AnyBranch"]:
         return AnyBranch(value) if value else None
 
+    def full_name(self) -> "AnyBranch":
+        return self
+
 
 class LocalBranch(AnyBranch):
```
PawelLipski: I'd use `LocalBranchName`, `RemoteBranchName` for consistency

git_machete/git_operations.py
```diff
@@ -15,40 +15,83 @@
...
     @staticmethod
     def of(value: str) -> Optional["LocalBranch"]:
         return LocalBranch(value) if value else None
 
+    def full_name(self) -> Optional["LocalBranchFullName"]:
+        return LocalBranchFullName.of(self)
+
+
+class LocalBranchFullName(LocalBranch):
```
PawelLipski: Hmm I'm now thinking if it makes sense for `LocalBranchFullName` to inherit from (i.e. **be a special case of**) `LocalBranch`, same for remote branches

git_machete/git_operations.py
PawelLipski: (plus `AnyBranchName`)

git_machete/git_operations.py
PawelLipski: Since now if you've got an instance of `LocalBranch`, you've got no guarantee whether it's a long name or short name... how about making `LocalBranch(Short?)Name` and `LocalBranchFullName` both inherit directly from `AnyBranchName`? would it complicate things a lot?

PawelLipski: This way e.g. if we have `managed_branches` of type `List[LocalBranchShortName]`, we know that this list only contains branch names with**out** prefixes.

git_machete/git_operations.py
```diff
@@ -15,40 +15,83 @@
...
+        return RemoteBranchFullName.of(self)
+
 
-class FullCommitSha(AnyRevision):
+class RemoteBranchFullName(RemoteBranch):
     @staticmethod
-    def of(value: str) -> Optional["FullCommitSha"]:
-        return FullCommitSha(value) if value else None
+    def of(value: str) -> Optional["RemoteBranchFullName"]:
+        return RemoteBranchFullName(f"refs/remotes/{value}") if value else None
```
PawelLipski: Hmmm this is confusing (since now `of` requires a short name, but returns a long one)... I'd rather make a distinction like:
```suggestion
    def of(value: str) -> Optional["RemoteBranchFullName"]:
        return RemoteBranchFullName(value) if value else None
    @staticmethod
    def from_short_name(value: RemoteBranchShortName) -> "RemoteBranchFullName":
        return self.of(f"refs/remotes/{value}")
```

PawelLipski: Similarly for the local branches

git_machete/git_operations.py
```diff
@@ -283,29 +326,29 @@ def is_full_sha(revision: AnyRevision) -> Optional[Match[str]]:
...
         else:
-            return self.get_commit_sha_by_revision(FullCommitSha.of(revision), prefix)
+            return self.get_commit_sha_by_revision(revision)
 
-    def get_committer_unix_timestamp_by_revision(self, revision: AnyBranch, prefix: str = "refs/heads/") -> int:
+    def get_committer_unix_timestamp_by_revision(self, revision: AnyBranch) -> int:
         if self.__committer_unix_timestamp_by_revision_cached is None:
             self.__load_branches()
-        return self.__committer_unix_timestamp_by_revision_cached.get(prefix + revision, 0)
+        return self.__committer_unix_timestamp_by_revision_cached.get(revision.full_name(), 0)
```
PawelLipski, line 391: Pls refine the key type of `__committer_unix_timestamp_by_revision_cached`

git_machete/git_operations.py
```diff
@@ -365,10 +408,10 @@ def __load_branches(self) -> None:
             if len(values) != 4:
                 continue  # invalid, shouldn't happen
             branch, commit_sha, tree_sha, committer_unix_timestamp_and_time_zone = values
-            b_stripped = AnyBranch.of(re.sub("^refs/remotes/", "", branch))
+            b_stripped = re.sub("^refs/remotes/", "", branch)
             self.__remote_branches_cached += [RemoteBranch.of(b_stripped)]
-            self.__commit_sha_by_revision_cached[FullCommitSha.of(branch)] = FullCommitSha.of(commit_sha)
-            self.__tree_sha_by_commit_sha_cached[FullCommitSha.of(commit_sha)] = FullCommitSha.of(tree_sha)
+            self.__commit_sha_by_revision_cached[RemoteBranch.of(branch)] = FullCommitHash.of(commit_sha)
+            self.__tree_sha_by_commit_sha_cached[FullCommitHash.of(commit_sha)] = FullCommitHash.of(tree_sha)
```
PawelLipski: Slight nuance... tree hash is NOT a commit hash. I mean, both are SHA-1 hashes... but corresponding to different types of objects, tree vs commit. So maybe introducing yet another type, `FullTreeHash` would be good idea for the sake of clarity.

Pls note that a tree hash is NOT even a revision in any sense ("revision" typically refers to something that identifies a commit, like a commit hash or branch name or tag; tree hash does NOT identify a commit, but a tree), so this `FullTreeHash` should just extend `str` directly and not e.g. `AnyRevision`.

git_machete/git_operations.py
rifel123: So far i didn't observe anything, but I'd like to keep it at least till end of this PR.

git_machete/git_operations.py
```diff
@@ -44,6 +44,9 @@ class LocalBranchFullName(LocalBranch):
     def of(value: str) -> Optional["LocalBranchFullName"]:
         return LocalBranchFullName(f"refs/heads/{value}") if value else None
 
+    def full_name(self) -> Optional["LocalBranchFullName"]:
+        return LocalBranchFullName.of(self)
```
PawelLipski: Just `self`?

git_machete/git_operations.py
rifel123: Yes, good point, it should work.

git_machete/client.py
```diff
@@ -24,7 +24,7 @@
     colored, underline, dim, get_second)
 
 
-BRANCH_DEF = Tuple[LocalBranch, str]
+Branch_Def = Tuple[LocalBranchShortName, str]
```
PawelLipski: Just `BranchDef`; the underscores in `Hash_ShortHash_Message` serve as "And"

git_machete/client.py
```diff
@@ -926,11 +926,11 @@ def print_line_prefix(branch_: LocalBranch, suffix: str) -> None:
                 print_line_prefix(branch, f"{utils.get_vertical_bar()} \n")
                 if opt_list_commits:
                     if edge_color[branch] in (EscapeCodes.RED, EscapeCodes.DIM):
-                        commits: List[Hash_ShortHash_Message] = self.__git.get_commits_between(fp_sha(branch), LocalBranch.of(f"refs/heads/{branch}")) if fp_sha(branch) else []
+                        commits: List[Hash_ShortHash_Message] = self.__git.get_commits_between(fp_sha(branch), LocalBranchShortName.of(f"refs/heads/{branch}")) if fp_sha(branch) else []
```
PawelLipski: Just `full_name()`? the current form is incorrect

PawelLipski: Pls Ctrl+Shift+F for some stray leftovers of `refs/heads/` and `refs/remotes/`, I think they should only be present in `full_name()` implementations and (almost) nowhere else (maybe some invocations of `git for-each-ref` can also retain them)

git_machete/client.py
```diff
@@ -1519,19 +1519,19 @@ def __infer_upstream(self, branch: LocalBranch, condition: Callable[[LocalBranch
         return None
 
     @staticmethod
-    def config_key_for_override_fork_point_to(branch: LocalBranch) -> str:
+    def config_key_for_override_fork_point_to(branch: LocalBranchShortName) -> str:
         return f"machete.overrideForkPoint.{branch}.to"
 
     @staticmethod
-    def config_key_for_override_fork_point_while_descendant_of(branch: AnyBranch) -> str:
+    def config_key_for_override_fork_point_while_descendant_of(branch: AnyBranchName) -> str:
```
PawelLipski: Param should be be `LocalBranchShortName`

git_machete/git_operations.py
```diff
@@ -94,7 +103,16 @@ def full_name(self) -> "ShortCommitHash":
...
+class FullTreeHash(str):
+    @staticmethod
+    def of(value: str) -> Optional["FullTreeHash"]:
+        return FullTreeHash(value) if value else None
+
+    def full_name(self) -> "FullTreeHash":
+        return self
+
+
+Reflog_Entry = Tuple[FullCommitHash, str]
```
PawelLipski: `ReflogEntry`

git_machete/git_operations.py
```diff
@@ -94,7 +103,16 @@ def full_name(self) -> "ShortCommitHash":
         return self
 
 
-REFLOG_ENTRY = Tuple[FullCommitHash, str]
+class FullTreeHash(str):
+    @staticmethod
+    def of(value: str) -> Optional["FullTreeHash"]:
+        return FullTreeHash(value) if value else None
+
+    def full_name(self) -> "FullTreeHash":
```
PawelLipski: `full_name` is probably not needed here (at least I wouldn't expect it to ever be called... is it currently?)

git_machete/cli.py
```diff
@@ -791,4 +791,4 @@ def main() -> None:
 
 
 if __name__ == "__main__":
-    main()
+    launch(['status'])
```
PawelLipski: Whoops

PawelLipski: I suppose it's due to draft status of the PR ;) Btw, while currently this is not captured by any of our tests, with #321 it would be (FYI @BartoszSiwinski)

rifel123: yes, i had to debug couple of things in the meantime :) 
Tests don't capture it because this line is never reached then, in fact this line never used until someone wants to run command directly from IDE.

git_machete/git_operations.py
rifel123: Yes, since `FullTreeHash` is not ancestor of `AnyRevision` there is no point to have it.

PawelLipski: *descendant aka subclass, not ancestor ;) yes

git_machete/git_operations.py
```diff
@@ -430,11 +455,11 @@ def __load_branches(self) -> None:
             if len(values) != 4:
                 continue  # invalid, shouldn't happen
             branch, commit_sha, tree_sha, committer_unix_timestamp_and_time_zone = values
-            b_stripped = re.sub("^refs/remotes/", "", branch)
-            self.__remote_branches_cached += [RemoteBranchShortName.of(b_stripped)]
-            self.__commit_sha_by_revision_cached[RemoteBranchShortName.of(branch)] = FullCommitHash.of(commit_sha)
+            b_stripped_remote = RemoteBranchFullName.of(branch).to_short_name()
```
PawelLipski, line 464: gr8

git_machete/git_operations.py
rifel123: I have added similar checks to `{Remote, Local}Branch{Full, Short}Name` and it revealed 4 mistyped  places. Maybe we should keep those checks then?

PawelLipski (approved):
That was a long but a very needed one if we're to maintain this project in the next years to come 😅 approving 🎉

git_machete/git_operations.py
PawelLipski: As discussed on Slack... :yolo:, let's keep

PR #296, opened 2021-10-18
Move `validate_python_version` to `git-machete` file

### LINKED ISSUE #285: What actually happens when git-machete is executed against Python 2 or 3.5?
There is a check for that versions in the code... but my suspicion is that Python will fail on syntax error (e.g. due to `f"..."`-style string interpolations) first. Let's verify and circumvent this issue if needed

rifel123:
closes #285 
Solution to check python version before syntax error is raised.

### DIFF
--- a/ci/checks/enforce-indent-two-spaces-outside-python.sh
+++ b/ci/checks/enforce-indent-two-spaces-outside-python.sh
@@ -22,5 +22,5 @@ awk_script="$(cat <<'EOF'
 EOF
 )"
 
-git ls-files ':!/.circleci/config.yml' ':!*/Dockerfile' ':!graphics/setup-sandbox' ':!docs/*' ':!*.md' ':!*.py' ':!*.svg' ':!*.png' ':!git-machete' \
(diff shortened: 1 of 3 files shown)


ci/checks/enforce-indent-two-spaces-outside-python.sh
```diff
@@ -22,5 +22,5 @@ awk_script="$(cat <<'EOF'
 EOF
 )"
 
-git ls-files ':!/.circleci/config.yml' ':!*/Dockerfile' ':!graphics/setup-sandbox' ':!docs/*' ':!*.md' ':!*.py' ':!*.svg' ':!*.png' \
+git ls-files ':!/.circleci/config.yml' ':!*/Dockerfile' ':!graphics/setup-sandbox' ':!docs/*' ':!*.md' ':!*.py' ':!*.svg' ':!*.png' ':!git-machete' \
```
PawelLipski: I wouldn't add an exception, just use two spaces in that file as well :)

Also, while you're here, pls restore abc order in the existing exceptions

PawelLipski (approved):

PawelLipski (approved):

PR #298, opened 2021-10-20
Issue #284: Fix the bash completion

### LINKED ISSUE #284: Bash completion scripts completes local paths for commands if no matching prefix found
image

Probably `COMPREPLY` is **not** set to empty in some path of execution

BartoszSiwinski:
Fix the issue that bash completion scripts completes local paths for commands if no matching prefix has been found
Screenshot 2021-10-20 at 14 35 23
.

### DIFF
--- a/completion/git-machete.completion.bash
+++ b/completion/git-machete.completion.bash
@@ -61,7 +61,11 @@ _git_machete() {
       esac ;;
     *)
       if [[ $COMP_CWORD -eq 2 ]]; then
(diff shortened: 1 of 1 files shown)


rifel123:
Ok, works as expected :) The only thing i found weird is that `git machete github create-pr` completes `-` and after another two hits of TAB we got list of available options for `create-pr`, but in general scope of this task is filled IMO.

rifel123 (approved):

PR #302, opened 2021-10-21
Issue #289: Add github and is-managed commands to top-level help.

### LINKED ISSUE #289: `github` command missing from the top-level `help`

BartoszSiwinski:
I've created new section `Utilities` and added both `github` and `is-managed` commands there.

Screenshot 2021-10-21 at 13 36 36

### DIFF
--- a/git_machete/cli.py
+++ b/git_machete/cli.py
@@ -31,7 +31,7 @@ alias_by_command: Dict[str, str] = {
 
 command_groups: List[Tuple[str, List[str]]] = [
     ("General topics",
-     ["file", "help", "hooks", "version"]),
+     ["file", "format", "help", "hooks", "version"]),
     ("Build, display and modify the tree of branch dependencies",
      ["add", "anno", "discover", "edit", "status"]),
     ("List, check out and delete branches",
--- a/git_machete/cli.py
+++ b/git_machete/cli.py
@@ -30,7 +30,6 @@ alias_by_command: Dict[str, str] = {
 }
 
 command_groups: List[Tuple[str, List[str]]] = [
-    # 'is-managed' is mostly for scripting use and therefore skipped
     ("General topics",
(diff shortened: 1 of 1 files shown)


git_machete/cli.py
```diff
@@ -40,7 +39,9 @@
     ("Determine changes specific to the given branch",
      ["diff", "fork-point", "log"]),
     ("Update git history in accordance with the tree of branch dependencies",
-     ["advance", "reapply", "slide-out", "squash", "traverse", "update"])
+     ["advance", "reapply", "slide-out", "squash", "traverse", "update"]),
+    ("Utilities",
```
PawelLipski: Hmmm I'd rather NOT merge `github` and `is-managed` into a single category... esp. that one is a major swiss-army-knife command and the other is a low-level tool for scripting. I'd say `is-managed` can be grouped with sth like `file` (and maybe `list`?), which is also a command for the use in scripts (rather than directly from the terminal).

git_machete/cli.py
```diff
@@ -30,7 +30,6 @@
 }
 
 command_groups: List[Tuple[str, List[str]]] = [
-    # 'is-managed' is mostly for scripting use and therefore skipped
     ("General topics",
      ["file", "help", "hooks", "version"]),
```
PawelLipski: Also, `format` seems to be missing 🤔

git_machete/cli.py
BartoszSiwinski: I will add `is-managed` to `List, check out and delete branches`. It seems to be the most relevant and existing group oriented on managing/analyzing branches.

PawelLipski: I'd expand the section title to sth like `Integrate with third party tools` for consistency with how other sections are named

PawelLipski (approved):

PR #303, opened 2021-10-21
Issue #299: Make help command and --help option work outside of a git repository.

### LINKED ISSUE #299: `git machete <command> --help` and `git machete help [<command>]` don't work outside of a git repository
`Not a git repository` error in each case. `git machete` isn't affected, curiously.

BartoszSiwinski:


git_machete/cli.py
```diff
@@ -491,6 +489,12 @@ def launch(orig_args: List[str]) -> None:
 
         cmd = parsed_cli.command
 
+        if cmd == "help":
+            print(get_help_description(parsed_cli.topic_or_cmd))
+            exit_script()
+
+        machete_client = MacheteClient(git)
+
```
BartoszSiwinski, line 497: MacheteClient have a line in `__init__`
`self._definition_file_path: str = self.__git.get_git_subpath("machete")`
Method called in here requires current location to be a git repository, so I had to move machete client lower. I had to also extract processing help from normal flow, because it has to be executed before MacheteClient instance is created.

PawelLipski (approved):

PR #317, opened 2021-10-26
Issue #305: Fix message for error raised when invalid github token is used.

### LINKED ISSUE #305: Using invalid github token ends up with wrong message while `POST` is used.
Since my previous access token has expired, i face strange error message from git-machete:

Screenshot 2021-10-22 at 11 44 18

Message is a little bit misleading, cause given endpoint exists,  and in public repositories there is always possibility to perform `GET` on it. For some reason while firing `POST` with invalid token, github API returns 404(not found) but i'd rather expect 401(not authorized) in such a case.

BartoszSiwinski:


PawelLipski:
Some transient error with dockerhub, restarted the failed job...

PawelLipski (approved):

PR #330, opened 2021-11-03
Issue #328: Bugfix - mistyped CommandLineOptions.opt_branch

### LINKED ISSUE #328: `AttributeError: 'str' object has no attribute 'full_name'` due to (?) mistyped `CommandLineOptions.opt_branch`
For an already existing, but not yet managed branch `foo/bar`:

```
$ git machete add foo/bar
  File "/usr/local/bin/git-machete", line 4, in <module>
    __import__('pkg_resources').run_script('git-machete==3.5.0', 'git-machete')
  File "/usr/lib/python3.7/site-packages/pkg_resources/__init__.py", line 666, in run_script
    self.require(requires)[0].run_script(script_name, ns)
  File "/usr/lib/python3.7/site-packages/pkg_resources/__init__.py", line 1469, in run_script
    exec(script_code, namespace, namespace)
  File "/usr/local/lib/python3.7/site-packages/git_machete-3.5.0-py3.7.egg/EGG-INFO/scripts/git-machete", line 19, in <module>
  File "/usr/local/lib/python3.7/site-packages/git_machete-3.5.0-py3.7.egg/git_machete/cli.py", line 803, in main
  File "/usr/local/lib/python3.7/site-packages/git_machete-3.5.0-py3.7.egg/git_machete/cli.py", line 522, in launch
  File "/usr/local/lib/python3.7/site-packages/git_machete-3.5.0-py3.7.egg/git_machete/client.py", line 263, in add
  File "/usr/local/lib/python3.7/site-packages/git_machete-3.5.0-py3.7.egg/git_machete/client.py", line 1502, in __infer_upstream
  File "/usr/local/lib/python3.7/site-packages/git_machete-3.5.0-py3.7.egg/git_machete/client.py", line 1479, in __match_log_to_filtered_reflogs
  File "/usr/local/lib/python3.7/site-packages/git_machete-3.5.0-py3.7.egg/git_machete/git_operations.py", line 495, in spoonfeed_log_shas
  File "/usr/local/lib/python3.7/site-packages/git_machete-3.5.0-py3.7.egg/git_machete/git_operations.py", line 487, in __get_log_shas
AttributeError: 'str' object has no attribute 'full_name'
```

This is most likely due to assignment `cli_opts.opt_branch = arg`, with `arg` being `str`-typed (?)... not sure why it passed mypy in the first place, pls investigate.

Also, regression test + preferable a systematic solution against this kind of errors would be very useful.

rifel123:
closes #328 

Mypy didn't catch obvious bug when function typed to return `LocalBranchShortName` actually returns a string instead.
I have manually checked functions which return something for type their type correctness, but here seems to be the only occurence.

Regression test for  `git machete add` is also included.


git_machete/cli.py
```diff
@@ -457,7 +457,7 @@ def set_utils_global_variables(
 
 def get_branch_arg_or_current_branch(
         cli_opts: git_machete.options.CommandLineOptions, git_context: GitContext) -> LocalBranchShortName:
-    return cli_opts.opt_branch or git_context.get_current_branch()
+    return LocalBranchShortName.of(cli_opts.opt_branch) or git_context.get_current_branch()
```
PawelLipski: Hmm does it address other uses of `cli_opts.opt_branch` across the code (if any)? shouldn't the conversion to `LocalBranchShortName` happen rather in the place where `opt_branch` is assigned?

rifel123: Yup, switched to typing in `update_cli_opts_using_parsed_args`, which I believe is more precise action.

git_machete/cli.py
```diff
@@ -360,21 +360,21 @@ def update_cli_opts_using_parsed_args(
         parsed_args: argparse.Namespace) -> None:
     for opt, arg in vars(parsed_args).items():
         if opt == "branch":
-            cli_opts.opt_branch = arg
```
PawelLipski, line 363: Okay... and if it comes to the original false positive in mypy (i.e. no error shown even though it should)... is there any flag (in mypy.ini) that could be switched on to prevent further errors like this one? or is it just a bug in mypy itself? maybe it should be reported in https://github.com/python/mypy/issues if so 🤔

rifel123: Huh, it took me quite a long to time to dig into mypy so deep. Unfortunately, this is normal mypy behaviour. Arguments that come from untyped functions/variables are silently treated by mypy as `Any`. Since `argparse` is not typed, everything that comes from `argparse.Namespace` will be taken as `Any` :(  
Even if we add `type=LocalBranchShortName` into argument parser for `branch`, python debugger will see `branch` as `LocalBranchShortName` but mypy always will see it as `Any` until you specifically tell mypy what is the exact type by casting (right now it's done this way, but casting does not solve the problem). There is no such a flag to change it. 

I also tried to build a class after `argparse.Namespace` with desired options typed, but still mypy sees them as `Any`.

The only reasonable solution here would be to use [TAP](https://pypi.org/project/typed-argument-parser/) which is a wrapper over `argparse`  with modernised solution for typing. 
But it would add external dependency to git-machete.

Personally, i would stick to precise casting rather than adding TAP. TBD

PawelLipski: Oh! okay, I see now ☹️ thanks for the breakdown!

Still, given the amount of instability that adding external deps will probably add... let's stick to the current approach!

PawelLipski: Actually! the only thing that could be done on this PR before merge is to add some comment in code that would specifically warn the future generations of this caveat (roughly your GH comment from above, maybe a bit redacted). Otherwise approving, thx for the research again

PawelLipski (approved):

PR #331, opened 2021-11-03
Issue #318: Enhance traverse message when no action is done

### LINKED ISSUE #318: git machete traverse didn't work
# Overview

- Working on this repo.
- Made more commits on feature branch.
- `git push` new commits to feature branch
- `git fetch`- develop behind upstream
- I'm on feature branch, execute `git machete traverse` - "nothing left to update"
- `git checkout master`, then `git machete traverse` - performs necessary updates

I'm not sure if this is as designed, but I've expected for `git machete traverse` to work the same regardless what is checked out. `git machete traverse --help`:

> Traverses the branch dependency tree in pre-order (i.e. simply in the order as they occur in the definition file) and for each branch

image

```
git version 2.33.1
git machete version 3.4.1
```

rifel123:
closes #318

### DIFF
--- a/docs/source/cli_help/traverse.rst
+++ b/docs/source/cli_help/traverse.rst
@@ -18,7 +18,9 @@ traverse
                          [--[no-]push] [--[no-]push-untracked]
                          [--return-to=WHERE] [--start-from=WHERE] [-w|--whole] [-W] [-y|--yes]
 
-Traverses the branch dependency tree in pre-order (i.e. simply in the order as they occur in the definition file) and for each branch:
+Traverses the branch dependency tree in pre-order (i.e. simply in the order as they occur in the definition file).
+By default traverse starts from current branch, however this behaviour can be customized using options: --start-from=, --whole or -w, -W.
+Command for each branch:
 
     * detects if the branch is merged to its parent/upstream:
 
--- a/docs/source/cli_help/traverse.rst
+++ b/docs/source/cli_help/traverse.rst
@@ -19,8 +19,8 @@ traverse
                          [--return-to=WHERE] [--start-from=WHERE] [-w|--whole] [-W] [-y|--yes]
 
 Traverses the branch dependency tree in pre-order (i.e. simply in the order as they occur in the definition file).
-By default traverse starts from current branch, however this behaviour can be customized using options: --start-from=, --whole or -w, -W.
-Command for each branch:
+By default traverse starts from current branch, however this behaviour can be customized using options: ``--start-from=``, ``--whole`` or ``-w``, ``-W``.
+For each branch, the command:
 
     * detects if the branch is merged to its parent/upstream:
 
--- a/git_machete/client.py
+++ b/git_machete/client.py
@@ -636,7 +636,7 @@ class MacheteClient:
         self.expect_at_least_one_managed_branch()
 
         self.__empty_line_status = True
-        handled_branches: List[LocalBranchShortName] = []
+        is_action_done: bool = False
 
         if opt_fetch:
             for remote in self.__git.get_remotes():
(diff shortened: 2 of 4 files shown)


git_machete/client.py
```diff
@@ -843,7 +847,8 @@ def traverse(
         else:
             msg = f"No successor of {bold(current_branch)} needs to be slid out or synced with upstream branch or remote"
         sys.stdout.write(f"{msg}; nothing left to update\n")
-
+        if not is_action_done and initial_branch != self.managed_branches[0]:
+            sys.stdout.write(fmt("Tip: `traverse` by default starts from the current branch, use flags (--starts-from=, --whole or -w, -W) to change this behavior.\nFurther info under `git machete traverse --help`.\n"))
```
PawelLipski: Pls split the line + surround options with backticks

git_machete/client.py
```diff
@@ -843,7 +847,8 @@ def traverse(
         else:
             msg = f"No successor of {bold(current_branch)} needs to be slid out or synced with upstream branch or remote"
         sys.stdout.write(f"{msg}; nothing left to update\n")
-
+        if not is_action_done and initial_branch != self.managed_branches[0]:
```
PawelLipski: Actually, as I'm now looking into the original issue #318... AFAICS this message would still NOT be printed in Konrad's case, right? maybe the just the `not is_action_done` part should be left (?)

git_machete/client.py
```diff
@@ -636,6 +636,7 @@ def traverse(
         self.expect_at_least_one_managed_branch()
 
         self.__empty_line_status = True
+        is_action_done: bool = False
```
PawelLipski: `any_action_done`? also, this name is not adequate... rather `any_action_suggested` (?)

git_machete/client.py
rifel123: It depends on case, there are several in Konrad's example:
- start from `issue/290-fish-completion` where there is nothing to be done: Tip will be printed.
- start from `develop` where some action actions needs to be done: Tip will not be printed.

And other possible cases:
- start from `develop` but everything is synced: Tip will be printed.
- start from `master` : Tip will not ever be printed, cause we checked all branches from the definition file.

rifel123: hmm, maybe it should not print tip when starts from any of root branch?

docs/source/cli_help/traverse.rst
```diff
@@ -18,7 +18,9 @@ traverse
                          [--[no-]push] [--[no-]push-untracked]
                          [--return-to=WHERE] [--start-from=WHERE] [-w|--whole] [-W] [-y|--yes]
 
-Traverses the branch dependency tree in pre-order (i.e. simply in the order as they occur in the definition file) and for each branch:
+Traverses the branch dependency tree in pre-order (i.e. simply in the order as they occur in the definition file).
+By default traverse starts from current branch, however this behaviour can be customized using options: --start-from=, --whole or -w, -W.
```
PawelLipski: Backticks pls (double backticks, in fact)

git_machete/docs.py
```diff
@@ -611,7 +611,9 @@
                                        [--[no-]push] [--[no-]push-untracked]
                                        [--return-to=WHERE] [--start-from=WHERE] [-w|--whole] [-W] [-y|--yes]</b>
 
-        Traverses the branch dependency tree in pre-order (i.e. simply in the order as they occur in the definition file) and for each branch:
+        Traverses the branch dependency tree in pre-order (i.e. simply in the order as they occur in the definition file).
+        By default traverse starts from current branch, however this behaviour can be customized using options: --starts-from=, --whole or -w, -W.
```
PawelLipski: Backticks pls

git_machete/docs.py
```diff
@@ -611,7 +611,9 @@
...
+        Command for each branch:
```
```suggestion
        For each branch, the command:
```

PawelLipski: Here and in rst

git_machete/client.py
PawelLipski: Okay! let's try _not print tip when starts from any root branch (not just the **first** root branch)_ approach then

PawelLipski (approved):

PR #338, opened 2021-11-08
Do not flush `__tree_sha_by_commit_sha_cached`, `__merge_base_cached`, `__contains_equivalent_tree`

### LINKED ISSUE #311: `__tree_sha_by_commit_sha_cached` doesn't ever need to be flushed
Thanks to git immutability guarantees, we don't need to worry about a tree hash changing for a commit of the given hash ;)
The same point applies to e.g. `__merge_base_cached` and `__contains_equivalent_tree_cached`. TBD what else.

These caches never need to be flushed. Once an entry is saved, it doesn't make sense to remove it (unlike with e.g. `__commit_sha_by_revision_cached` which is prone to getting outdated over time, as branches can be repointed to different commits).

rifel123:
closes #311

Other fields which are being flushed right now, TBD:
`__config_cached` - can change, egz when new branch is added
`__remotes_cached` - can change when new remote is added
`__counterparts_for_fetching_cached` - can change after change of branch upstream
`__short_commit_sha_by_revision_cached` - from #311 description:  can be repointed to different commit
`__commit_sha_by_revision_cached` - same as above
`__committer_unix_timestamp_by_revision_cached` - after repointing to different commit timestamp also changes
`__local_branches_cached` - can change, egz when new branch is added
`__remote_branches_cached` - can change, egz after `git fetch`
`__reflogs_cached` - can change, egz after new branch is created

Here, i am not sure:
`__initial_log_shas_cached` 
`__remaining_log_shas_cached`
But those fields are being filled with `git log <revision>`  command, so this can also change, IMO it should stay as it is.


git_machete/git_operations.py
```diff
@@ -552,16 +552,13 @@ def flush_caches(self) -> None:
         self.__remotes_cached = None
         self.__counterparts_for_fetching_cached = None
         self.__short_commit_sha_by_revision_cached = {}
-        self.__tree_sha_by_commit_sha_cached = None
         self.__commit_sha_by_revision_cached = None
         self.__committer_unix_timestamp_by_revision_cached = None
         self.__local_branches_cached = None
         self.__remote_branches_cached = None
         self.__initial_log_shas_cached = {}
```
PawelLipski: Okay! if it comes to `__{initial,remaining}_log_shas_cached`... in the current shape it indeed **must be flushed** (since the key in both dicts is a branch name... and a branch can always be repointed to another commit).

You can consider, however, changing the `branch name -> log hashes` mapping to `full commit hash -> log hashes`. This way this new mapping will be immutable no matter what. You'll need to look up `get_commit_sha_by_revision()` first, however.

git_machete/client.py
```diff
@@ -1483,7 +1483,9 @@ def branch_def_to_str(lb: str, lb_or_rb: str) -> str:
             debug(f"match_log_to_filtered_reflogs({branch})",
                   "branches containing the given SHA in their filtered reflog: \n%s\n" % "\n".join(log_result()))
 
-        for sha in self.__git.spoonfeed_log_shas(branch):
+        branch_full_hash: FullCommitHash = self.__git.get_full_sha(branch)
```
PawelLipski: Hmmm while you're here... I can see there's both `get_commit_sha_by_revision` and `get_full_sha`... these should be unified as they basically do the same thing 🤔

PawelLipski (approved):

PR #341, opened 2021-11-10
Issue #334: install git-machete locally from freshly generated brew formula 

### LINKED ISSUE #334: Dry-run installation of the brew formula within the build of `develop`/`hotfix/*`
As for now, we're just test-committing the changes to VirtusLab/homebrew-git-machete... but not trying to install that formula locally. Let's give it a shot, also let's try running our functional tests against this fresh git-machete installation.

rifel123:
closes #334

Here is testing job to look up:
https://app.circleci.com/pipelines/github/VirtusLab/git-machete/1118/workflows/468c6f9d-dcd1-4b1e-8681-29342fcb61fe/jobs/3805

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -98,6 +98,30 @@ jobs:
           name: Run tests against Python 3.10 and git 2.33.1
           command: PYTHON_VERSION=3.10 GIT_VERSION=2.33.1 bash ci/tox/ci-run.sh
 
+  "dry run deploy to homebrew test":
+    executor: ubuntu_executor
+    steps:
+      - checkout
+      - deploy:
+          name: '[Dry run] Deploy to PyPI'
+          # Side effects: does an actual release, but to the test PyPI
+          command: bash ci/pypi/ci-deploy.sh --dry-run
+
+      - deploy:
+          name: '[Dry run] Update Homebrew tap with local installation test'
+          # Prerequisites: artifacts must be released to test PyPI first
+          command: bash ci/homebrew-tap/ci-deploy.sh --dry-run
+
+      - deploy:
+          name: '[Dry run] Build .snap for linux distros'
+          # Build snap for current branch of git-machete repository without publishing results to snapstore.
+          # --use-lxd applied to use a LXD container instead of a VM, to work around lack of support for KVM on CircleCI VMs.
+          command: |
+            sudo apt-get update
+            sudo apt-get install -y snapd
+            sudo snap install snapcraft --classic
+            snapcraft --use-lxd
+
   "dry run deploy":
     executor: ubuntu_executor
     steps:
@@ -181,6 +205,13 @@ workflows:
       - test against python 3_8 and git 2_25_0 with PEP8 checks
       - test against python 3_9 and git 2_30_1 with coverage report and docs creation
       - test against python 3_10 and git 2_33_1
+      - dry run deploy to homebrew test:
+          filters:
+            branches:
+              only:
+                - "/ci.*/"
+          requires:
+            *mandatory_jobs
       - dry run deploy:
           filters:
             branches:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -98,6 +98,20 @@ jobs:
           name: Run tests against Python 3.10 and git 2.33.1
           command: PYTHON_VERSION=3.10 GIT_VERSION=2.33.1 bash ci/tox/ci-run.sh
 
+  "dry run deploy ci homebrew":
+    executor: ubuntu_executor
+    steps:
+      - checkout
+      - deploy:
+          name: '[Dry run] Deploy to PyPI'
+          # Side effects: does an actual release, but to the test PyPI
+          command: bash ci/pypi/ci-deploy.sh --dry-run
+
+      - deploy:
+          name: '[Dry run] Update Homebrew tap'
+          # Prerequisites: artifacts must be released to test PyPI first
+          command: bash ci/homebrew-tap/ci-deploy.sh --dry-run
+
   "dry run deploy":
     executor: ubuntu_executor
     steps:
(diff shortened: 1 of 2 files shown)


ci/homebrew-tap/ci-deploy.sh
```diff
@@ -21,6 +21,7 @@ sha256=$(
   curl -s https://$pypi_host/pypi/git-machete/$version/json \
   | jq --raw-output '.urls | map(select(.packagetype == "sdist")) | .[0].digests.sha256')
 sed -i "s/git-machete-.*\.tar\.gz/git-machete-$version.tar.gz/" git-machete.rb
+sed -i "s/pypi.io/$pypi_host/" git-machete.rb
```
PawelLipski: Nit: I'd escape `.` since it's not treated as literal period here, but rather as a regex single-char placeholder

PawelLipski: Hmmm why double nesting? one level being `bash -c`, the other being `$(...)`

ci/homebrew-tap/ci-deploy.sh
```diff
@@ -29,4 +30,10 @@ if [[ $do_push == true ]]; then
   git push origin master
 else
   echo "Refraining from push since it's a dry run"
+  sed -i "s/pypi.io/$pypi_host/" git-machete.rb
+  # install git-machete from local formula with homebrew
+  /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh)" < /dev/null
+  echo 'eval "$(/home/linuxbrew/.linuxbrew/bin/brew shellenv)"' >> /home/circleci/.bash_profile
+  eval "$(/home/linuxbrew/.linuxbrew/bin/brew shellenv)"
```
PawelLipski: Hmmm is it needed to both add the `eval` to bash_profile AND to run it in the current shell?

PawelLipski: Also, I'd avoid hardcoding HOME path, just use `~`

ci/homebrew-tap/ci-deploy.sh
```diff
@@ -29,4 +30,10 @@ if [[ $do_push == true ]]; then
...
+  brew install ./git-machete.rb
```
PawelLipski, line 36: Hmm some run of `git machete`, like checking the version would be useful... unless such test in already executed within the Brew formula itself?

ci/homebrew-tap/ci-deploy.sh
```diff
@@ -29,4 +30,10 @@ if [[ $do_push == true ]]; then
...
 fi
```
PawelLipski, line 43: Also! pls note that this pollutes the current CI VM... would be useful to check if that interferes e.g. with `git-machete` installed via snap

ci/homebrew-tap/ci-deploy.sh
```diff
@@ -29,4 +30,10 @@ if [[ $do_push == true ]]; then
   git push origin master
 else
   echo "Refraining from push since it's a dry run"
+  sed -i "s/pypi.io/$pypi_host/" git-machete.rb
+  # install git-machete from local formula with homebrew
+  /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh)" < /dev/null
```
PawelLipski, line 34: Hmmm why double nesting? one level being `bash -c`, the other being `$(...)`

PawelLipski (approved):

PawelLipski (changes requested):
⚙️

ci/homebrew-tap/ci-deploy.sh
rifel123: yes, `eval` to bash profile is no necessary  in this case indeed. 

HOME path is hardcoded already brew's install script, here is explanation from their installation site:
The installation script installs Homebrew to /home/linuxbrew/.linuxbrew using sudo if possible and in your home directory at ~/.linuxbrew otherwise. Homebrew does not use sudo after installation. Using /home/linuxbrew/.linuxbrew allows the use of more binary packages (bottles) than installing in your personal home directory.

The prefix /home/linuxbrew/.linuxbrew was chosen so that users without admin access can ask an admin to create a linuxbrew role account and still benefit from precompiled binaries.

ci/homebrew-tap/ci-deploy.sh
rifel123: i'll remove it with `brew remove git-machete` after version check.

ci/homebrew-tap/ci-deploy.sh
rifel123: hmm, this line is provided by brew, but `$(...)` is to grab the body of `install.sh` and `bash -c` to execute it.

I could split it into two lines, but i do not see much improvement here.

PawelLipski: Oh! okay, I see the point in that, let's leave as is

.circleci/config.yml
```diff
@@ -181,6 +205,13 @@ workflows:
       - test against python 3_8 and git 2_25_0 with PEP8 checks
       - test against python 3_9 and git 2_30_1 with coverage report and docs creation
       - test against python 3_10 and git 2_33_1
+      - dry run deploy to homebrew test:
+          filters:
+            branches:
+              only:
+                - "/ci.*/"
```
PawelLipski: Hmmm is this job meant to be retained? 🤔

ci/homebrew-tap/ci-deploy.sh
PawelLipski: Resolved? 🤔

PawelLipski (approved):

PR #346, opened 2021-11-15
Issue #345: Better handling of error messages from `UnprocessableEntityError`

### LINKED ISSUE #177: Create-pr: Check if reviewers belong to current organisation on Github
Currently there is no check performed before adding reviewers to pull request, which causes HTTP 422 error `UnprocessableEntity`. There has to be check before if reviewer belongs to organisation. Informations about organization can be found under [orgs endpoint](https://docs.github.com/en/rest/reference/orgs) or directly  under [members endpoint](https://docs.github.com/en/rest/reference/orgs#list-organization-members)

### LINKED ISSUE #345: Unify error message handling from `UnprocessableEntityError`
When Github API returns status code 422, the data can be served on 3 ways (3 shapes of response was observed):
- in case of adding reviewers from outside organisation:
`{'message': 'Reviews may only be requested from collaborators. One or more of the users or teams you specified is not a collaborator of the rifel123/sandbox repository.', 'documentation_url': 'https://docs.github.com/rest/reference/pulls#request-reviewers-for-a-pull-request'}`
- in case of creating pull request which already exists:
`{'message': 'Validation Failed', 'errors': [{'resource': 'PullRequest', 'code': 'custom', 'message': 'A pull request already exists for rifel123:new_branch.'}], 'documentation_url': 'https://docs.github.com/rest/reference/pulls#create-a-pull-request'}`
- in case when base branch for PR does not exists on remote:
`{'message': 'Validation Failed', 'errors': [{'resource': 'PullRequest', 'field': 'base', 'code': 'invalid'}], 'documentation_url': 'https://docs.github.com/rest/reference/pulls#create-a-pull-request'}`
- in case of creating a PR with base branch equal to head branch:
`{'message': 'Validation Failed', 'errors': [{'resource': 'PullRequest', 'code': 'custom', 'message': 'No commits between develop and same'}], 'documentation_url': 'https://docs.github.com/rest/reference/pulls#create-a-pull-request'}`

Currently only second and fourth option is covered. Question is rather what to do when no relevant info is available, like in third response? Maybe just print out all response? TBD

rifel123:
closes #345 

does not fully close #177, as there should be info served how to force git-machete not to add reviewers. 
Draft, because there is still open point what to do when no relevant message is received, currently print all the response.

### DIFF
--- a/git_machete/github.py
+++ b/git_machete/github.py
@@ -125,6 +125,19 @@ def __get_github_token() -> Optional[str]:
             get_token_from_hub())
 
 
+def __exctract_failure_info_from_422(response: Any) -> str:
+    if response['message'] != 'Validation Failed':
+        return str(response['message'])
+    ret: List[str] = []
+    if response.get('errors'):
+        for error in response['errors']:
+            if error.get('message'):
+                ret.append(error['message'])
+    if ret:
+        return '\n'.join(ret)
+    return str(response)
+
+
 def __fire_github_api_request(method: str, path: str, token: Optional[str], request_body: Optional[Dict[str, Any]] = None) -> Any:
     headers: Dict[str, str] = {
         'Content-type': 'application/json',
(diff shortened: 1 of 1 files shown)


git_machete/github.py
```diff
@@ -125,6 +125,19 @@ def get_token_from_file_in_home_directory() -> Optional[str]:
             get_token_from_hub())
 
 
+def __exctract_failure_info_from_422(response: Any) -> str:
```
PawelLipski: `extract` ;)

git_machete/github.py
```diff
@@ -125,6 +125,19 @@ def get_token_from_file_in_home_directory() -> Optional[str]:
...
+    if response['message'] != 'Validation Failed':
+        return str(response['message'])
```
PawelLipski, line 130: Would it make sense to append `errors` (if any) even if `message` is present?

rifel123: Seems like `errors` are available only when main message is other than `Validation Failed`.

PawelLipski (approved):

PR #347, opened 2021-11-16
Issue #177: Enchance error message when assigning reviewers from outside organisation

### LINKED ISSUE #177: Create-pr: Check if reviewers belong to current organisation on Github

rifel123:
closes #177 after #345 is also merged.

### DIFF
--- a/git_machete/client.py
+++ b/git_machete/client.py
@@ -2055,7 +2055,7 @@ class MacheteClient:
             except UnprocessableEntityHTTPError as e:
                 if 'Reviews may only be requested from collaborators.' in e.msg:
                     warn(f"There are some invalid reviewers in .{self.__git.get_git_subpath('info', 'reviewers')} file.\n"
-                         "Pull request will be created without reviewers.")
+                         "Skipping adding reviewers to pull request.")
                 else:
                     raise e
             print(fmt(ok_str))
--- a/git_machete/client.py
+++ b/git_machete/client.py
@@ -2054,7 +2054,7 @@ class MacheteClient:
                 add_reviewers_to_pull_request(org, repo, pr.number, reviewers)
             except UnprocessableEntityHTTPError as e:
                 if 'Reviews may only be requested from collaborators.' in e.msg:
-                    warn(f"There are some invalid reviewers in .{self.__git.get_git_subpath('info', 'reviewers')} file.\n"
+                    warn(f"There are some invalid reviewers in {self.__git.get_git_subpath('info', 'reviewers')} file.\n"
                          "Skipping adding reviewers to pull request.")
                 else:
                     raise e
--- a/git_machete/client.py
+++ b/git_machete/client.py
@@ -12,7 +12,7 @@ from git_machete import utils
 from git_machete.constants import (
     DISCOVER_DEFAULT_FRESH_BRANCH_COUNT, PICK_FIRST_ROOT, PICK_LAST_ROOT,
     EscapeCodes, SyncToRemoteStatuses)
-from git_machete.exceptions import MacheteException, StopInteraction
+from git_machete.exceptions import MacheteException, StopInteraction, UnprocessableEntityHTTPError
 from git_machete.git_operations import AnyBranchName, AnyRevision, FullCommitHash, GitContext, HEAD, LocalBranchShortName, RemoteBranchShortName
 from git_machete.github import (
     add_assignees_to_pull_request, add_reviewers_to_pull_request,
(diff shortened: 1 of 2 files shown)


PawelLipski:
It's _enhance_ ;)

git_machete/github.py
```diff
@@ -126,8 +126,12 @@ def get_token_from_file_in_home_directory() -> Optional[str]:
 
 
 def __extract_failure_info_from_422(response: Any) -> str:
+    additional_msg: str = ''
     if response['message'] != 'Validation Failed':
```
PawelLipski: Hmmm it's pretty brittle... what if the failure is NOT due to reviewers being outside the organization? `message` being `Validation Failed` can mean many different things in principle (even if we haven't observed anything else so far)

git_machete/github.py
```diff
@@ -126,8 +126,12 @@ def get_token_from_file_in_home_directory() -> Optional[str]:
...
-        return str(response['message'])
+        if 'Reviews may only be requested from collaborators.' in response['message']:
+            additional_msg = "\nThere are some invalid reviewers in .git/info/reviewers file. " \
+                             "Please remove them or all file to create a pull request without reviewers."
```
PawelLipski: _all file_?...

PawelLipski: Oh, _all_ never applies to a single thing (same as _wszystkie_ in Polish), I'd rather say _or remove the entire file_

git_machete/github.py
```diff
@@ -125,6 +125,23 @@ def get_token_from_file_in_home_directory() -> Optional[str]:
             get_token_from_hub())
 
 
+def __extract_failure_info_from_422(response: Any) -> str:
+    additional_msg: str = ''
+    if response['message'] != 'Validation Failed':
+        if 'Reviews may only be requested from collaborators.' in response['message']:
+            additional_msg = "\nThere are some invalid reviewers in .git/info/reviewers file. " \
```
PawelLipski: Actually! use `get_git_subpath` rather than `.git/`, the git folder doesn't always need to live under `.git/`

PawelLipski: (see `git machete help file`)

git_machete/client.py
```diff
@@ -2048,9 +2048,16 @@ def create_github_pr(
...
-                fmt(f'Adding {", ".join(f"`{reviewer}`" for reviewer in reviewers)} as reviewer{"s" if len(reviewers) > 1 else ""} to PR #{pr.number}...'),
+                fmt(f'Adding {", ".join(f"`{reviewer}`" for reviewer in reviewers)} as reviewer{"s" if len(reviewers) > 1 else ""} to PR #{pr.number}...\n'),
                 end='', flush=True)
-            add_reviewers_to_pull_request(org, repo, pr.number, reviewers)
+            try:
+                add_reviewers_to_pull_request(org, repo, pr.number, reviewers)
+            except UnprocessableEntityHTTPError as e:
+                if 'Reviews may only be requested from collaborators.' in e.msg:
+                    warn(f"There are some invalid reviewers in .{self.__git.get_git_subpath('info', 'reviewers')} file.\n"
+                         "Pull request will be created without reviewers.")
```
rifel123: I decided t move it here because PR is already created at this point, so previous message would be misleading.

Since PR is created earlier, it should be reflected in git-machete annotations.

PawelLipski: Good spot, indeed!

git_machete/client.py
```diff
@@ -2048,9 +2048,16 @@ def create_github_pr(
...
             print(
-                fmt(f'Adding {", ".join(f"`{reviewer}`" for reviewer in reviewers)} as reviewer{"s" if len(reviewers) > 1 else ""} to PR #{pr.number}...'),
+                fmt(f'Adding {", ".join(f"`{reviewer}`" for reviewer in reviewers)} as reviewer{"s" if len(reviewers) > 1 else ""} to PR #{pr.number}...\n'),
                 end='', flush=True)
-            add_reviewers_to_pull_request(org, repo, pr.number, reviewers)
+            try:
+                add_reviewers_to_pull_request(org, repo, pr.number, reviewers)
+            except UnprocessableEntityHTTPError as e:
+                if 'Reviews may only be requested from collaborators.' in e.msg:
+                    warn(f"There are some invalid reviewers in .{self.__git.get_git_subpath('info', 'reviewers')} file.\n"
```
PawelLipski: Stray period, probably
```suggestion
                    warn(f"There are some invalid reviewers in {self.__git.get_git_subpath('info', 'reviewers')} file.\n"
```

PawelLipski (approved):

PR #352, opened 2021-11-19
issue #:312Remove `yq` installation from CI

### LINKED ISSUE #312: Is `yq` installation in CI config still needed?
We do install `yq`... but only in the `initial checks` job, where `yq` is **not** used AFAICS.

`yq` is in turn used in the other jobs (in `ci/docker-build-and-push.sh`)... but `yq` isn't getting explicitly installed within these jobs.

I suspect `yq` is already included in the Ubuntu VM we're using... let's research that remove the explicit install from CI config if possible.

rifel123:
closes #312

Indeed `yq` isntallation is no longer needed as without it everything runs smoothly.
A Job with dry runs and disabled `yq`:
https://app.circleci.com/pipelines/github/VirtusLab/git-machete/1144/workflows/439ab1b6-39a1-4a13-8d25-04adfe175157/jobs/3972
(or build 1144 in CircleCI)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -30,11 +30,6 @@ jobs:
 
             # For ensure-links-correct.sh
             npm install --global remark-cli remark-lint-no-dead-urls remark-validate-links
-
-            # For the scripts doing `docker-compose pull` to work around https://github.com/docker/compose/issues/7258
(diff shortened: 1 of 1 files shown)


PawelLipski (approved):

PawelLipski:
Yeah, indeed I can see a successful run of `yq eval .services.deb-ppa-upload.image -` in deb/PPA step... so `yq` seems to be baked into the VM!

PR #375, opened 2021-12-29
Dockerize git-machete

### LINKED ISSUE #368: Dockerize git-machete
Would you like a `Dockerfile` for this project? @PawelLipski 
If yes, I can write one.

mohitsaxenaknoldus:
Closes #368


PawelLipski (commented):
Great! I'll take a deeper look once I'm back from new year's break

Dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM ubuntu:latest
+
+RUN sudo apt-get install -y software-properties-common && sudo add-apt-repository ppa:virtuslab/git-machete && sudo apt-get update && sudo apt-get install -y python3-git-machete
```
PawelLipski: Okay! so as first thing... pls split this `RUN` into a few lines so that commenting on individual lines is easier!

Dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM ubuntu:latest
+
+RUN sudo apt-get install -y software-properties-common && sudo add-apt-repository ppa:virtuslab/git-machete && sudo apt-get update && sudo apt-get install -y python3-git-machete 
+
+
+
```
PawelLipski: Also... if you could add a step to CircleCI config, to be executed on `master` branch builds, that will automatically push the image to Dockerhub (I'll create the required image under the `gitmachete` organization)

Dockerfile
mohitsaxenaknoldus: Having it in a single line actually reduces the number of layers which makes the image altogether a lot smaller and efficient.

PawelLipski: Oh! you mean a single **Dockerfile instruction** (`RUN` in this case), yes... but for the sake of readability, the instruction can be split with `\`, as done e.g. in https://github.com/VirtusLab/git-machete/blob/master/ci/deb-ppa-upload/Dockerfile#L11

Dockerfile
```diff
@@ -1,6 +1,9 @@
 FROM ubuntu:latest
 
-RUN sudo apt-get install -y software-properties-common && sudo add-apt-repository ppa:virtuslab/git-machete && sudo apt-get update && sudo apt-get install -y python3-git-machete 
+RUN sudo apt-get install -y software-properties-common \
+    && sudo add-apt-repository ppa:virtuslab/git-machete \
+    && sudo apt-get update \
+    && sudo apt-get install -y python3-git-machete
```
PawelLipski: Okay! now let's proceed with CircleCI automation, then I'll probably have yet another comment to this Dockerfile itself

Dockerfile
mohitsaxenaknoldus: What's the use of that?

PawelLipski: Oh! is the idea for this Docker image to be built on end user's machines from Dockerfile rather than being readily downloadable from Dockerhub? 🤔

mohitsaxenaknoldus: I meant, we can push it to Dockerhub manually once, and then it will be handled by the package manager. Why push it on every change on the master if we are not building the Dockerfile from the code?

PawelLipski: Hmmm okay but pls note that the package you're installing here (`python3-git-machete`) is itself updated on every single `master` build: https://github.com/VirtusLab/git-machete/blob/master/.circleci/config.yml#L145

As the most recent version of `python3-git-machete` installed is 3.6.1, this effectively means that the docker image built from this Dockerfile will point to git-machete 3.6.1... at least until the new version (3.6.2? 3.7.0?) is published!

PawelLipski: So... I'd rather have a versioning scheme of Docker images consistent with the versions of git-machete itself 🤔

mohitsaxenaknoldus: oh ok, now I get what you mean. Let me do it.

.circleci/config.yml
```diff
@@ -166,6 +166,21 @@ jobs:
...
+    working_directory: /git-machete
+    docker:
+      - image: docker:17.05.0-ce-git
+    steps:
+      - checkout
+      - setup_remote_docker
+      - deploy:
+          name: Publish application to docker hub
+          command: |
+            docker login -e $DOCKER_HUB_EMAIL -u $DOCKER_HUB_USER_ID -p $DOCKER_HUB_PWD
```
PawelLipski: Pls take a look at https://github.com/VirtusLab/git-machete/blob/master/ci/docker-build-and-push.sh#L19 , the credentials are already provided in the CI... also, password should rather be passed on stdin rather than via command line

.circleci/config.yml
```diff
@@ -166,6 +166,21 @@ jobs:
...
+            docker tag git-machete $DOCKER_HUB_USER_ID/dockerapp:latest
```
PawelLipski: Hmmm where is this `git-machete` image being built? 🤔

.circleci/config.yml
mohitsaxenaknoldus: Please check again if it is good now.

.circleci/config.yml
```diff
@@ -166,6 +166,21 @@ jobs:
       #    is triggered automatically via GitHub webhook on https://readthedocs.org/projects/git-machete/builds/,
       # 3) every time a https://github.com/r-ryantm bot runs on NixOS/nixpkgs repository,
       #    Nix git-machete package is updated with the fresh version/package hash according to the data from PyPI.
+
+  "Publish Dockerfile to Docker Hub":
```
PawelLipski: `Publish image`, not Dockerfile ;)

Dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM ubuntu:latest
+
+RUN apt-get update \
+  && apt-get install -y software-properties-common \
+  && add-apt-repository ppa:virtuslab/git-machete \
```
PawelLipski: Please note that at the moment when this step is being executed, the package might still **not** be built in Ubuntu Launchpad... in fact, from my observation it takes around ~20 minutes for a package to be available in the PPA once the `ci/deb-ppa-upload/ci-deploy.sh` script is complete. The result will be that you're gonna install the **previous** version of `python3-git-machete` package, not the one you've just built in the CI for `master` branch :///

PawelLipski: As a remedy, I'd suggest switching to some `python3-alpine` (?) image as a base, and installing `git-machete` package from PyPI. As opposed to Ubuntu PPA (where package publishing is asynchronous), with PyPI once `ci/pypi/ci-deploy.sh` completes, the package is already available.

PawelLipski: In fact, to make sure that the correct version of the package is installed from PyPI, you can pass `git_machete_version` as an ARG to this dockerfile... see e.g. https://github.com/VirtusLab/git-machete/blob/master/ci/homebrew-tap/ci-deploy.sh#L13 for how to extract the version

.circleci/config.yml
```diff
@@ -166,6 +166,21 @@ jobs:
...
+    docker:
+      - image: docker:17.05.0-ce-git
+    steps:
+      - checkout
+      - setup_remote_docker
+      - deploy:
+          name: Publish application to docker hub
+          command: |
+            echo $DOCKER_PASSWORD | docker login -u $DOCKER_USERNAME --password-stdin
+            docker build -t $DOCKER_USERNAME/git-machete:latest .
```
PawelLipski: Pls use `gitmachete` rather than `$DOCKER_USERNAME` (which is some technical user IIRC) as the organization

.circleci/config.yml
```diff
@@ -166,6 +166,21 @@ jobs:
       #    is triggered automatically via GitHub webhook on https://readthedocs.org/projects/git-machete/builds/,
       # 3) every time a https://github.com/r-ryantm bot runs on NixOS/nixpkgs repository,
       #    Nix git-machete package is updated with the fresh version/package hash according to the data from PyPI.
```
PawelLipski, line 174: Also! to fix the failing CI, pls rebase onto the most recent develop, I've patched a few things

.circleci/config.yml
```diff
@@ -166,6 +166,21 @@ jobs:
...
+    working_directory: /git-machete
+    docker:
+      - image: docker:17.05.0-ce-git
+    steps:
+      - checkout
+      - setup_remote_docker
+      - deploy:
+          name: Publish application to docker hub
+          command: |
+            echo $DOCKER_PASSWORD | docker login -u gitmachete --password-stdin
```
PawelLipski: But here `$DOCKER_LOGIN` is desired!
Since for logging in, we use some **technical user** (whose name is stored in `DOCKER_LOGIN` variable), but we push to images under `gitmachete` organization (https://hub.docker.com/u/gitmachete)

PawelLipski (approved):
Ok! looks good to me, feel free to merge. I'll release 3.6.2 or 3.7.0 quite soon, and then the image will be published as well :)

mohitsaxenaknoldus:
> Ok! looks good to me, feel free to merge. I'll release 3.6.2 or 3.7.0 quite soon, and then the image will be published as well :)

I don't have write access for merging :(

PawelLipski:
> > Ok! looks good to me, feel free to merge. I'll release 3.6.2 or 3.7.0 quite soon, and then the image will be published as well :)
> 
> I don't have write access for merging :(

Ouch sorry, forgot... done myself ;)

PR #379, opened 2022-01-09
Add support for extra rebase options via `GIT_MACHETE_REBASE_OPTS` env var

### LINKED ISSUE #376: Add ability to pass extra arguments to git-rebase
# Overview

Feature request.
When doing `git machete traverse`, seldom I'd like to be able to pass extra flags to git-rebase, e.g. `-Xignore-all-space --rebase-merges`.

## Example use case

I'm batching multiple smaller PRs. By preserving merges in the history, I'm able to auto-close PRs on GitHub. In order to do that, I need `--rebase-merges` flag.

PawelLipski:


docs/source/cli_help/reapply.rst
```diff
@@ -20,3 +20,7 @@ but there is also dedicated ``squash`` command that achieves the same goal witho
 **Options:**
 
 -f, --fork-point=<fork-point-commit>    Specifies the alternative fork point commit after which the rebased part of history is meant to start.
+
+**Environment variables:**
+
+GIT_MACHETE_REBASE_OPTS                 Extra options to pass to the underlying ``git rebase`` invocation, space-separated. Example: ``GIT_MACHETE_REBASE_OPTS="--keep-empty --rebase-merges" git machete reapply``.
```
PawelLipski: Formatting to be fixed: https://4675-122743101-gh.circle-artifacts.com/0/docs/index.html#reapply

git_machete/git_operations.py
```diff
@@ -754,11 +754,15 @@ def merge_fast_forward_only(self, branch: LocalBranchShortName) -> None:  # refs
 
     def rebase(self, onto: AnyRevision, fork_revision: AnyRevision, branch: LocalBranchShortName, opt_no_interactive_rebase: bool) -> None:
         def do_rebase() -> None:
+            # Let's use `OPTS` suffix for consistency with git's built-in env var `GIT_DIFF_OPTS`
+            git_machete_rebase_opts_var = 'GIT_MACHETE_REBASE_OPTS'
+            rebase_opts_str = os.environ.get(git_machete_rebase_opts_var)
+            rebase_opts = filter(None, rebase_opts_str.split(' ')) if rebase_opts_str else []
```
kgadek: May I propose:

```suggestion
            rebase_opts_str = os.environ.get(git_machete_rebase_opts_var, "")
            rebase_opts = [option for option in rebase_opts_str.split() if option]
```

- `.get` can return default. Another benefit, if #118 is pursued, is that `rebase_opts_str: str` instead of `rebase_opts_str: Optional[str]`
  > Return the value for key if key is in the dictionary, else default. If default is not given, it defaults to None, so that this method never raises a KeyError.

- I'm not sure if it's a common knowledge what `filter(None, …)` does
  > Note that filter(function, iterable) is equivalent to the generator expression (item for item in iterable if function(item)) if function is not None and (item for item in iterable if item) if function is None.

- `split()` is probably better:
  > If sep is not specified or is None, a different splitting algorithm is applied: runs of consecutive whitespace are regarded as a single separator, and the result will contain no empty strings at the start or end if the string has leading or trailing whitespace. Consequently, splitting an empty string or a string consisting of just whitespace with a None separator returns [].

git_machete/git_operations.py
```diff
@@ -754,11 +754,15 @@ def merge_fast_forward_only(self, branch: LocalBranchShortName) -> None:  # refs
...
             try:
                 if opt_no_interactive_rebase:
-                    self._run_git("rebase", "--onto", onto, fork_revision, branch)
+                    self._run_git("rebase", *rebase_opts, "--onto", onto, fork_revision, branch)
                 else:
-                    self._run_git("rebase", "--interactive", "--onto", onto, fork_revision, branch)
+                    self._run_git("rebase", *rebase_opts, "--interactive", "--onto", onto, fork_revision, branch)
```
kgadek: Since we have `rebase_opts`, we could do:
```python
if opt_no_interactive_rebase:
    rebase_opts.append("--interactive")

self._run_git("rebase", *rebase_opts, "--onto", onto, fork_revision, branch)
```

kgadek (approved):

git_machete/git_operations.py
PawelLipski: It's `if NOT opt_no_interactive_rebase: rebase_opts.append("--interactive")` ;) changed

git_machete/git_operations.py
PawelLipski: Good spots, all three

PR #383, opened 2022-01-26
Issue #373: `git merge-base` works when branches have no common commit

### LINKED ISSUE #373: `git merge-base` fails with exit code 1 when branches have no common commit, crashing git-machete
Pretty easy to reproduce, just list 2 branches in .git/machete so that each have no commit in common (e.g. 2 different root commits).

Bug NOT present in git-machete-intellij-plugin.

amalota:
closes [#373](https://github.com/VirtusLab/git-machete/issues/373)


RELEASE_NOTES.md
```diff
@@ -1,5 +1,9 @@
 # Release notes
 
+## New in git-machete 3.7.3
+
+- fixed: merge-base works for branches that have no common commit
```
PawelLipski: We still remain in 3.7.2! take a look at https://github.com/VirtusLab/git-machete/releases, 3.7.2 hasn't been released yet

git_machete/git_operations.py
```diff
@@ -645,7 +646,7 @@ def __get_merge_base(self, sha1: FullCommitHash, sha2: FullCommitHash) -> FullCo
             #   then there is exactly one merge-base - the ancestor,
             # * if neither of sha1, sha2 is an ancestor of another,
             #   then none of the (possibly more than one) merge-bases is equal to either of sha1/sha2 anyway.
-            self.__merge_base_cached[sha1, sha2] = FullCommitHash.of(self._popen_git("merge-base", sha1, sha2).rstrip())
```
PawelLipski, line 648: Hmmm one thing: pls update the above comment to include the zero merge-base case 🤔

git_machete/git_operations.py
```diff
@@ -645,7 +646,7 @@ def __get_merge_base(self, sha1: FullCommitHash, sha2: FullCommitHash) -> FullCo
...
+            self.__merge_base_cached[sha1, sha2] = FullCommitHash.of(self._popen_git("merge-base", sha1, sha2, allow_non_zero=True).rstrip())
```
PawelLipski: Also, I'd extract the first line from the output of `git merge-base`... in case of criss-cross histories, there will be >1 line in the output, and we shouldn't wrap this multiline string into `FullCommitHash`

PawelLipski: Actually... in the zero merge-base case, it should probably be `None` saved into `self.__merge_base_cached[sha1, sha2]`... currently it'd be `FullCommitHash.of("")`

amalota: > Also, I'd extract the first line from the output of git merge-base... in case of criss-cross histories, there will be >1 line in the output, and we shouldn't wrap this multiline string into FullCommitHash

There is no need for that because `merge-base`, by default returns only one merge base, you would have to provide `--all` option in order to get all merge bases.

amalota: > Actually... in the zero merge-base case, it should probably be None saved into self.__merge_base_cached[sha1, sha2]... currently it'd be FullCommitHash.of("")

Yes, None is already saved into `self.__merge_base_cached[sha1, sha2]`, because `FullCommitHash.of("")` returns None.

git_machete/git_operations.py
```diff
@@ -645,7 +646,10 @@ def __get_merge_base(self, sha1: FullCommitHash, sha2: FullCommitHash) -> FullCo
             #   then there is exactly one merge-base - the ancestor,
             # * if neither of sha1, sha2 is an ancestor of another,
             #   then none of the (possibly more than one) merge-bases is equal to either of sha1/sha2 anyway.
-            self.__merge_base_cached[sha1, sha2] = FullCommitHash.of(self._popen_git("merge-base", sha1, sha2).rstrip())
```
amalota, line 648: There is no need for `rstrip()` since `self._popen_git()` does `strip()`.

git_machete/git_operations.py
PawelLipski: Oh! okay, sorry I didn't remember the former and wasn't aware of the latter 😅 let's leave as is then

RELEASE_NOTES.md
```diff
@@ -3,6 +3,7 @@
 ## New in git-machete 3.7.2
 
 - fixed: package version retrieval outside of git repository
+- fixed: `git merge-base` works for branches that have no common commit
```
PawelLipski: I'd put it like this:
```suggestion
- fixed: checking whether a branch is merged to parent works for branches that have no common commit
```

Since generally implementation details (like underlying git commands) are better off not included in release notes

git_machete/git_operations.py
```diff
@@ -645,7 +646,10 @@ def __get_merge_base(self, sha1: FullCommitHash, sha2: FullCommitHash) -> FullCo
...
+            # In the rare case when sha1, sha2 have no common commits, the flag: allow_non_zero=True
+            # (allows, non zero exit code to be returned by git merge-base command, without raising an exception)
+            # is used and the __get_merge_base function returns None.
```
PawelLipski, line 651: Generally `Optional`s are still treated very liberally coz we haven't applied mypy strict mode yet (#118)... but I'd change the return type of this method to `Optional[FullCommitHash]` while we're here, even though mypy doesn't complain yet

PawelLipski (approved):

git_machete/git_operations.py
amalota: This method has return type `Optional[FullCommitHash]` already, no need to change it :)

PawelLipski: Hmmm wait
```python
def __get_merge_base(self, sha1: FullCommitHash, sha2: FullCommitHash) -> FullCommitHash:
```

PR #384, opened 2022-01-27
[Bugfix] Fix `**kwargs` typing

amalota:


RELEASE_NOTES.md
```diff
@@ -1,5 +1,9 @@
 # Release notes
 
+## New in git-machete 3.7.3
+
+- fixed: *args and **kwargs typing
```
PawelLipski, line 5: Nit: since this is formatted as markdown, pls `\`-escape the asterisks

PawelLipski: Oh! actually, since 3.7.2 has NOT been released yet, and this change is another bugfix (not a feature that would change CLI), we can stay within **3.7.2**. No need to bump version; I'd actually even skip this entry from changelog b/c it's very low-level/not particularly interesting for the users

PawelLipski: See https://github.com/VirtusLab/git-machete/blob/master/CONTRIBUTING.md#versioning and the surrounding paragraphs btw

git_machete/__init__.py
```diff
@@ -1 +1 @@
-__version__ = '3.7.2'
+__version__ = '3.7.3'
```
PawelLipski: To be reverted

git_machete/git_operations.py
```diff
@@ -179,17 +179,17 @@ def __init__(self) -> None:
         self.__contains_equivalent_tree_cached: Dict[Tuple[FullCommitHash, FullCommitHash], bool] = {}
 
     @staticmethod
-    def _run_git(git_cmd: str, *args: str, **kwargs: Dict[str, str]) -> int:
+    def _run_git(git_cmd: str, *args: str, **kwargs: Any) -> int:
```
PawelLipski, line 182: Actually... why hasn't `mypy` capture that problem so far, unlike in the [blog post you've linked](https://adamj.eu/tech/2021/05/11/python-type-hints-args-and-kwargs/)? 🤔 is it possible to reconfigure our mypy to capture this type of errros?

amalota: Both `pycharm` and `mypy` captured the problem while trying to do https://github.com/VirtusLab/git-machete/pull/383, so there is no need to configure anything. 
image

PawelLipski: Oh! okay, as in, `mypy` captures the problem but at **call site** (which has only been added now right?) and not at the **declaration**

PawelLipski: That's fine, let's leave as is

PawelLipski (approved):

PR #385, opened 2022-01-27
Issue #365: Check if release notes are updated with version bump

### LINKED ISSUE #365: Make sure that RELEASE_NOTES are updated with version bump
I.e. add a script to `ci/checks/` that would verify that if version in `__init__.py` is `x.y.z`, then the latest entry in `RELEASE_NOTES.md` also corresponds to `x.y.z`

amalota:
closes https://github.com/VirtusLab/git-machete/issues/365

### DIFF
--- a/ci/checks/ensure-release-notes-up-to-date.sh
+++ b/ci/checks/ensure-release-notes-up-to-date.sh
@@ -3,7 +3,7 @@
 set -e -o pipefail -u
 
 current_version=$(grep '__version__ = ' git_machete/__init__.py | cut -d\' -f2)
-release_notes_version=$(egrep -o [0-9].[0-9].[0-9] RELEASE_NOTES.md -m 1)
+release_notes_version=$(egrep -o '(0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*)(?:-((?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\+([0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?' RELEASE_NOTES.md --max-count 1)
 
 if [[ "$(printf "$current_version\n$release_notes_version" | sort --version-sort | tail -n 1)" != "$release_notes_version" ]]; then
   echo "Please update RELEASE_NOTES! Current version: $current_version, latest version in release notes: $release_notes_version"
--- a/ci/checks/ensure-release-notes-up-to-date.sh
+++ b/ci/checks/ensure-release-notes-up-to-date.sh
@@ -1,0 +1,11 @@
+#!/usr/bin/env bash
+
+set -e -o pipefail -u
+
+current_version=$(grep '__version__ = ' git_machete/__init__.py | cut -d\' -f2)
+release_notes_version=$(awk -F"## New in git-machete " '/## New in git-machete /{print $2;exit;}' RELEASE_NOTES.md)
+
+if [[ "$(printf "$current_version\n$release_notes_version" | sort --version-sort | tail -n 1)" != "$release_notes_version" ]]; then
+  echo "Please update RELEASE_NOTES! Current version: $current_version, latest version in release notes: $release_notes_version"
+  exit 1
+fi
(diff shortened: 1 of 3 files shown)


ci/checks/ensure-release-notes-up-to-date.sh
```diff
@@ -3,7 +3,7 @@
 set -e -o pipefail -u
 
 current_version=$(grep '__version__ = ' git_machete/__init__.py | cut -d\' -f2)
-release_notes_version=$(awk -F"## New in git-machete " '/## New in git-machete /{print $2;exit;}' RELEASE_NOTES.md)
+release_notes_version=$(egrep -o [0-9].[0-9].[0-9] RELEASE_NOTES.md -m 1)
```
PawelLipski: Note that it won't match e.g. `3.10.1`
Also, I'd expand `-m` into the long form since it's a relatively obscure option

ci/checks/ensure-release-notes-up-to-date.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
+
+set -e -o pipefail -u
+
+current_version=$(grep '__version__ = ' git_machete/__init__.py | cut -d\' -f2)
+release_notes_version=$(egrep -o [0-9].[0-9].[0-9] RELEASE_NOTES.md -m 1)
+
+if [[ "$(printf "$current_version\n$release_notes_version" | sort --version-sort | tail -n 1)" != "$release_notes_version" ]]; then
```
PawelLipski: Hmmm isn't it enough to just compare `$current_version` and `$release_notes_version`? 🤔

PawelLipski: In `ensure-version-bumped.sh`, such logic makes sense, since there you're checking whether what's in `__init__.py` is **higher** than the latest git tag... but here only equality matters

ci/checks/ensure-release-notes-up-to-date.sh
amalota: Great catch, thanks!

ci/checks/ensure-release-notes-up-to-date.sh
amalota: Yes, i though about that, this comparison also checks the case where `$release_notes_version` > `$current_version` but now since you mentioned it i am not so sure that `$release_notes_version` being higher then `$current_version` is correct, maybe it shouldn't be allowed, wdyt?

ci/checks/ensure-release-notes-up-to-date.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
+
+set -e -o pipefail -u
+
+current_version=$(grep '__version__ = ' git_machete/__init__.py | cut -d\' -f2)
+release_notes_version=$(egrep -o '(0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*)(?:-((?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\+([0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?' RELEASE_NOTES.md --max-count 1)
```
amalota: I managed to find official regex for semantic version matching, at the very end of site: https://semver.org/, should we use this or the previous, simpler one ( `\d{1,}.\d{1,}.\d{1,}`)?

PawelLipski: image

PawelLipski: Let's stick to the short one! I don't plan on ever using build metadata/pre-release versions

ci/checks/ensure-release-notes-up-to-date.sh
PawelLipski: Exactly, `$release_notes_version` being higher than `$current_version` should **NOT** be allowed.
Actually, some tooling (Debian changelog generation) even depends on the fact that the two are equal.

ci/checks/ensure-release-notes-up-to-date.sh
amalota: image

PawelLipski (approved):

PR #388, opened 2022-01-31
Issue #387: `github create-pr` takes the PR title from first unique commit

### LINKED ISSUE #387: `github create-pr` should take the PR title from first unique commit, not the latest commit
Just as the message of the commit created by `git machete squash` is taken from the first unique commit (the commit right past the branch's fork point) rather than from the latest commit &mdash; given my own user's experience, that's what I expect in >90% of cases for both `squash` and `github create-pr`.

Btw, `git_machete.git_operations.GitContext.get_commit_information` should probably NOT have a default value in `commit` &mdash; let's make it always clear from which commit the information is extracted.

amalota:
closes #387

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -5,6 +5,7 @@
 - fixed: package version retrieval outside of git repository
 - fixed: checking whether a branch is merged to parent works for branches that have no common commit
 - added: CI/CD check ensuring that RELEASE_NOTES are up to date
+- fixed: `github create-pr` takes the PR title from the first unique commit
 
 ## New in git-machete 3.7.1
 
--- a/git_machete/client.py
+++ b/git_machete/client.py
@@ -1290,11 +1290,11 @@ class MacheteClient:
             return
 
         earliest_sha, earliest_short_sha, earliest_subject = commits[0]
-        earliest_full_body = self.__git.get_commit_information("raw body", FullCommitHash.of(earliest_sha)).strip()
+        earliest_full_body = self.__git.get_commit_information(FullCommitHash.of(earliest_sha), "raw body").strip()
         # %ai for ISO-8601 format; %aE/%aN for respecting .mailmap; see `git rev-list --help`
-        earliest_author_date = self.__git.get_commit_information("author date", FullCommitHash.of(earliest_sha)).strip()
-        earliest_author_email = self.__git.get_commit_information("author email", FullCommitHash.of(earliest_sha)).strip()
-        earliest_author_name = self.__git.get_commit_information("author name", FullCommitHash.of(earliest_sha)).strip()
+        earliest_author_date = self.__git.get_commit_information(FullCommitHash.of(earliest_sha), "author date").strip()
+        earliest_author_email = self.__git.get_commit_information(FullCommitHash.of(earliest_sha), "author email").strip()
+        earliest_author_name = self.__git.get_commit_information(FullCommitHash.of(earliest_sha), "author name").strip()
 
         # Following the convention of `git cherry-pick`, `git commit --amend`, `git rebase` etc.,
         # let's retain the original author (only committer will be overwritten).
@@ -2021,7 +2021,9 @@ class MacheteClient:
         debug(f'create_github_pr({head})', f'organization is {org}, repository is {repo}')
         debug(f'create_github_pr({head})', 'current GitHub user is ' + (current_user or '<none>'))
 
-        title: str = self.__git.get_commit_information("subject").strip()
+        fork_point = self.__git.get_merge_base(self.__git.get_commit_sha_by_revision(head), self.__git.get_commit_sha_by_revision(base))
+        commits: List[Hash_ShortHash_Message] = self.__git.get_commits_between(fork_point, head)
+        _, _, title = commits[0]
         description_path = self.__git.get_git_subpath('info', 'description')
         description: str = utils.slurp_file_or_empty(description_path)
 
--- a/git_machete/client.py
+++ b/git_machete/client.py
@@ -10,7 +10,7 @@ import git_machete.github
 import git_machete.options
 from git_machete import utils
 from git_machete.constants import (
-    DISCOVER_DEFAULT_FRESH_BRANCH_COUNT, PICK_FIRST_ROOT, PICK_LAST_ROOT,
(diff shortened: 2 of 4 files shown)


git_machete/git_operations.py
```diff
@@ -865,7 +865,7 @@ def get_latest_checkout_timestamps(self) -> Dict[str, int]:  # TODO (#110): defa
                     result[to_branch] = int(match.group(1))
         return result
 
-    def get_commit_information(self, information: str, commit: Optional[AnyRevision] = None) -> str:
+    def get_commit_information(self, commit: AnyRevision, information: str) -> str:
         if information not in GIT_FORMAT_PATTERNS:
```
PawelLipski: Actually! while you're here, maybe instead of using raw strings like `"author date"` to distinguish b/w different pieces of information... let's use constants (enumeration?) instead, which will fail during mypy check in case of a typo etc., rather than in runtime

RELEASE_NOTES.md
```diff
@@ -4,6 +4,7 @@
 
 - fixed: package version retrieval outside of git repository
 - fixed: checking whether a branch is merged to parent works for branches that have no common commit
+- added: `github create-pr` takes the PR title from the first unique commit
```
PawelLipski: Nitpick: I'd say `fixed:`

git_machete/client.py
```diff
@@ -2021,7 +2021,9 @@ def create_github_pr(
         debug(f'create_github_pr({head})', f'organization is {org}, repository is {repo}')
         debug(f'create_github_pr({head})', 'current GitHub user is ' + (current_user or '<none>'))
 
-        title: str = self.__git.get_commit_information("subject").strip()
+        fork_point = self.__git.get_merge_base(self.__git.get_commit_sha_by_revision(head), self.__git.get_commit_sha_by_revision(base))
```
PawelLipski: Fork point should rather be computed using `git_machete.client.MacheteClient.fork_point`!

See https://medium.com/virtuslab/make-your-way-through-the-git-rebase-jungle-with-git-machete-e2ed4dbacd02 (esp. the appendix)

The concept is an improvement over what `git rebase --fork-point` does &mdash; basically, to determine **after which commit the unique history** of the given branch started. Or in other words, what's the latest ancestor commit of the given branch that has been committed while a different branch has been checked out.

This will indeed be sometimes equal to `merge_base(branch, parent(branch))` (e.g. in all green-edge cases)... but in many cases (esp. red edges) it isn't! Feel free to ask further questions &mdash; this concept is pretty non-trivial since git does NOT assign commits to branches - branches do NOT contain commits, branches only **point to** commits. So finding the set of commits "uniquely belonging" to a branch is working around git data model... although still quite useful ;)

git_machete/client.py
```diff
@@ -2021,7 +2021,9 @@ def create_github_pr(
...
+        commits: List[Hash_ShortHash_Message] = self.__git.get_commits_between(fork_point, head)
```
PawelLipski, line 2027: Also, see how `squash` is handled in cli.py

git_machete/git_operations.py
```diff
@@ -865,14 +864,14 @@ def get_latest_checkout_timestamps(self) -> Dict[str, int]:  # TODO (#110): defa
                     result[to_branch] = int(match.group(1))
         return result
 
-    def get_commit_information(self, information: str, commit: Optional[AnyRevision] = None) -> str:
-        if information not in GIT_FORMAT_PATTERNS:
+    def get_commit_information(self, commit: AnyRevision, information: GitFormatPatterns) -> str:
+        if information not in GitFormatPatterns:
```
PawelLipski, line 868: This check is probably redundant now right? I mean, it doesn't seem possible to pass a value of `information` that would NOT be `in GitFormatPatterns` without some heavy hacking (assuming mypy still passes), right?

git_machete/client.py
```diff
@@ -2021,7 +2021,11 @@ def create_github_pr(
         debug(f'create_github_pr({head})', f'organization is {org}, repository is {repo}')
         debug(f'create_github_pr({head})', 'current GitHub user is ' + (current_user or '<none>'))
 
-        title: str = self.__git.get_commit_information("subject").strip()
+        fork_point = self.fork_point(head, True, opt_no_detect_squash_merges=False)
+        if not fork_point:
+            raise MacheteException(f"Could not find a fork-point for branch {head}.")
+        commits: List[Hash_ShortHash_Message] = self.__git.get_commits_between(fork_point, head)
+        _, _, title = commits[0]
```
PawelLipski, line 2028: Hmmm in case `commits` is empty (which might happen! when no unique commit committed on the current branch has been detected &mdash; e.g. that's a brand new branch)... this will fail! in such case, I'd fall back to the current commit (as it used to be before your fix)

git_machete/client.py
```diff
@@ -2021,7 +2021,11 @@ def create_github_pr(
         debug(f'create_github_pr({head})', f'organization is {org}, repository is {repo}')
         debug(f'create_github_pr({head})', 'current GitHub user is ' + (current_user or '<none>'))
 
-        title: str = self.__git.get_commit_information("subject").strip()
+        fork_point = self.fork_point(head, True, opt_no_detect_squash_merges=False)
```
PawelLipski: Pls name the `True` param

git_machete/git_operations.py
amalota: Well, you can still make a TYPO or type enum field that is not in the class.

e.g. 
`class A(Enum):`. 
 `    FIELD_1 = 1`

`A.FEILD_1`

git_machete/client.py
amalota: This corner case is already taken care of by the `__sync_before_creating_pr` function, which wont allow to create the PR.

```
if self.__git.is_ancestor_or_equal(current_branch.full_name(), up_branch.full_name()):
    raise MacheteException(
        f'All commits in `{current_branch}` branch are already included in `{up_branch}` branch.\nCannot create pull request.')
```

PawelLipski: Ouch! good spot ;)

PawelLipski (approved):
Approving... to make the CI pass, pls address https://github.com/VirtusLab/git-machete/issues/389

PR #392, opened 2022-02-02
Issue #390: Use namedtuple instead of tuple for Hash_ShortHash_Message, BranchDef, ReflogEntry

### LINKED ISSUE #390: Use namedtuple instead of tuple in `Hash_ShortHash_Message = Tuple[str, str, str]`
I know that type name: `Hash_ShortHash_Message` has field names in it but it gets lost along the way and it can be easier and cleaner to use namedtuple.
e.g.

change this:
`        

        commits: List[Hash_ShortHash_Message] = self.__git.get_commits_between(
            opt_fork_point, current_branch)

        if not commits:
            raise MacheteException(
                "No commits to squash. Use `-f` or `--fork-point` to specify the "
                "start of range of commits to squash.")

        if len(commits) == 1:
            sha, short_sha, subject = commits[0]
            print(f"Exactly one commit ({short_sha}) to squash, ignoring.\n")
            print("Tip: use `-f` or `--fork-point` to specify where the range of "
                  "commits to squash starts.")
            return`

to this:
`        

        if len(commits) == 1:
            print(f"Exactly one commit ({commits[0].short_sha}) to squash, ignoring.\n")
            print("Tip: use `-f` or `--fork-point` to specify where the range of "
                  "commits to squash starts.")
            return`

amalota:
closes https://github.com/VirtusLab/git-machete/issues/390

### DIFF
--- a/git_machete/client.py
+++ b/git_machete/client.py
@@ -11,7 +11,7 @@ import git_machete.options
 from git_machete import utils
 from git_machete.constants import (
     DISCOVER_DEFAULT_FRESH_BRANCH_COUNT, GitFormatPatterns, PICK_FIRST_ROOT, PICK_LAST_ROOT,
-    EscapeCodes, SyncToRemoteStatuses)
+    EscapeCodes, SyncToRemoteStatuses, GitLogResponse)
 from git_machete.exceptions import MacheteException, StopInteraction, UnprocessableEntityHTTPError
 from git_machete.git_operations import (
     AnyBranchName, AnyRevision, ForkPointOverrideData,
@@ -27,7 +27,6 @@ from git_machete.utils import (
 
 
 BranchDef = Tuple[LocalBranchShortName, str]
-Hash_ShortHash_Message = Tuple[str, str, str]
 
 
 # Allowed parameter values for show/go command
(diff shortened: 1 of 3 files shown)


PawelLipski:
Hmmm so this `namedtuple` feature is supposed to work only since 3.8... how come do the CI tests for 3.6 and 3.7 pass? 😯

amalota:
@PawelLipski It looks like it is available in python 3.6 https://docs.python.org/3.6/search.html?q=namedtuple&check_keywords=yes&area=default

PawelLipski:
Hmmm okay... so where did you take the information about it only being available in Python 3.8. then from? 🤔

amalota:
Here https://docs.python.org/3/library/typing.html#typing.NamedTuple, above `Other special directives` there is a text: `New in version 3.8.` but as it turns out its for feature above it, not below as i initially thought.

git_machete/client.py
```diff
@@ -1472,7 +1465,7 @@ def generate_entries() -> Generator[Tuple[FullCommitHash, BranchDef], None, None
                     self.__branch_defs_by_sha_in_reflog[sha] = [branch_def]
 
             def log_result() -> Generator[str, None, None]:
-                branch_defs_: List[BranchDef]
+                branch_defs_: List[BranchPair]
```
PawelLipski: Pls rename all vars/def like this one adequately

git_machete/git_operations.py
```diff
@@ -149,7 +149,20 @@ def __init__(self, to_hash: FullCommitHash, while_descendant_of_hash: FullCommit
...
+
+
+class GitReflogEntry(NamedTuple):
+    hash: FullCommitHash
+    reflog_subject: str
+
+
+class BranchPair(NamedTuple):
+    local_branch: LocalBranchShortName
+    remote_branch: AnyBranchName
```
PawelLipski: `local_or_remote_branch` (see e.g. `def branch_def_to_str`)

PawelLipski (approved):

Issue #396, opened 2022-02-03
Non-fixed UID and GID of VM users in CircleCI builds leads to forbidden access to volume shared b/w VM and container


PawelLipski:
## Problem

https://app.circleci.com/pipelines/github/VirtusLab/git-machete/1353/workflows/bec9fc8d-fb67-4771-9264-4a9db722c9f5 (and a few other builds previously) fail on `py.error.EACCES: [Permission denied]: mkdir('/home/ci-user/git-machete/.tox',)`

The problem arises because `ci-user` within the container has a different GID/UID than the user in the underlying VM provided by CI.

The following sentence in the [blog post on this project's CI setup](https://medium.com/virtuslab/nifty-docker-tricks-for-your-ci-vol-2-c5191a67f1a4) is NOT strictly correct:

_If you use any modern Linux distribution and have just one non-root user on your machine, they're both likely equal to 1000 [...]; on CircleCI VMs they both turn out to be 2000._

Yeah... they sometimes _turn out_ to be 2000, but at other times turn out to be something else! This means that some images stored in https://hub.docker.com/r/gitmachete/ci-tox have UID/GID of `ci-user` = 2000, and some... something else :/

If a pipeline where the VM user has UID=1000 tries to run `gitmachete/ci-tox` image with `ci-user` UID = 2000... the failure comes as soon as the container tries to write anything into (or maybe even read from, TBD) the volume (files in `.tox` happen to be the first offender), since these are owned by user of UID=1000 and not 2000 in both container and the enclosing VM.

Note that not only `ci-tox` image is affected, but also the other images (`ci-deb-ppa-upload`, `ci-rpm`) where `ci-user` is explicitly created.

## Solutions

1. Extend the Docker image tag with user and group id. E.g. for `ci-tox`, instead of `git<GIT-VERSION>-python<PYTHON-VERSION>-<ci/tox-DIRECTORY-SHA1>`, the tags would be `git<GIT-VERSION>-python<PYTHON-VERSION>-uid<UID>-gid<GID>--<ci/tox-DIRECTORY-SHA1>`.
2. `machine` executor in CircleCI (**probably**, to be verified) does NOT allow for selecting UID/GID. Maybe instead switch to Docker executor? Like, run the entire `ci-run.sh` script from within a `dind` (Docker-in-Docker) container, where UID/GID can be much more easily controlled.
[...]

PR #401, opened 2022-02-05
Mock `def exit_script` in tests to test CLI failure errors

PawelLipski:


ci/tox/Dockerfile
```diff
@@ -13,6 +13,7 @@ RUN set -x \
     && cd git-$git_version/ \
     && make configure \
     && ./configure \
+    && sed -i 's/#warning .*//' /usr/include/sys/poll.h  `# to reduce amount of spam warnings in logs` \
```
PawelLipski: This change, while useful on its own, will have a side effect of a new `gitmachete/ci-tox` Docker image being generated (since the SHA-1 git hash of `ci/tox/` directory changes)... which serves as kinda-workaround for #396 (the new image will use the UID/GID of the VM user), but only until CircleCI provides a VM with a new UID/GID

PawelLipski:
Crap, closed due to #402 :trollface:

PR #403, opened 2022-02-05
Mock `def exit_script` in tests to test CLI failure errors

PawelLipski:


ci/tox/Dockerfile
```diff
@@ -13,6 +13,7 @@ RUN set -x \
     && cd git-$git_version/ \
     && make configure \
     && ./configure \
+    && sed -i 's/#warning .*//' /usr/include/sys/poll.h  `# to reduce amount of spam warnings in logs` \
```
PawelLipski, line 16: This change, while useful on its own, will have a side effect of a new `gitmachete/ci-tox` Docker image being generated (since the SHA-1 git hash of `ci/tox/` directory changes)... which serves as kinda-workaround for #396 (the new image will use the UID/GID of the VM user), but only until CircleCI provides a VM with a new UID/GID

git_machete/tests/functional/test_machete.py
```diff
@@ -465,18 +472,18 @@ def test_branch_reappears_in_definition(self) -> None:
...
             """
-        expected_error_msg: str = fmt('.git/machete, line 5: branch `develop` re-appears in the tree definition. '
-                                      'Edit the definition file manually with `git machete edit`')
 
         self.repo_sandbox.new_branch("root")
         self.rewrite_definition_file(body)
 
-        machete_client = MacheteClient(git)  # Only to workaround sys.exit while calling launch(['status'])
-        try:
-            machete_client.read_definition_file()
```
PawelLipski, line 476: Removed the inconsistency here: some test methods (like this one) tested on the level of `MacheteClient` (which is CLI-agnostic), while most tested on the highest possible level - CLI itself (wrapped into `self.launch_command` call).

This PR unifies all test cases to cover all the levels from CLI downwards.

git_machete/cli.py
```diff
@@ -790,11 +792,9 @@ def strip_remote_name(remote_branch: RemoteBranchShortName) -> LocalBranchShortN
...
-        print(f"\n{e}", file=sys.stderr)
-        exit_script(2)
+        exit_script(2, e)
     except MacheteException as e:
-        print(f"\n{e}", file=sys.stderr)
-        exit_script(1)
+        exit_script(1, e)
     except KeyboardInterrupt:
         print("\nInterrupted by the user", file=sys.stderr)
         exit_script(1)
```
amalota:
```suggestion
    except (MacheteException, KeyboardInterrupt) as e:
        exit_script(1, e)
```

amalota: I think that `KeyboardInterrupt` provides enough information.

PR #407, opened 2022-02-07
Issue #396: (Option #1) Extend the Docker image tag with user and group id.

### LINKED ISSUE #396: Non-fixed UID and GID of VM users in CircleCI builds leads to forbidden access to volume shared b/w VM and container

amalota:
closes https://github.com/VirtusLab/git-machete/issues/396

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -6,6 +6,7 @@
 - fixed: checking whether a branch is merged to parent works for branches that have no common commit
 - added: CI/CD check ensuring that RELEASE_NOTES are up to date
 - fixed: `github create-pr` takes the PR title from the first unique commit
+- fixed: problems with access to volume shared b/w VM and container
 
 ## New in git-machete 3.7.1
 
--- a/ci/docker-pull-or-build-and-push.sh
+++ b/ci/docker-pull-or-build-and-push.sh
@@ -7,6 +7,9 @@ DIRECTORY_HASH=$(git rev-parse HEAD:ci/$image_name)
 export DIRECTORY_HASH
 cd ci/$image_name/
 
+export USER_ID=$(id -u)
+export GROUP_ID=$(id -g)
+
 function retry() {
   attempts=$1
   interval=10
--- a/ci/docker-pull-or-build-and-push.sh
+++ b/ci/docker-pull-or-build-and-push.sh
@@ -37,7 +37,7 @@ retry 3 docker-compose --ansi never pull $image_name
 # (since v1.25.1, `docker-compose pull` is NOT failing when it can't fetch the image).
 image_tag=$(docker-compose --ansi never config | yq eval ".services.$image_name.image" -)
 docker image inspect "$image_tag" &>/dev/null || {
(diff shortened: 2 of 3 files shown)


PawelLipski:
Okay... so is there a problem with this solution as well? Or is it viable (even if cumbersome due to even super-long image tags)?

amalota:
No, this one works just fine

PawelLipski:
Let's keep that as a backup in case #410 causes some unexpected troubles

PR #408, opened 2022-02-07
Issue #395: Update CONTRIBUTING.md

### LINKED ISSUE #395: Document releases in more detail
That a develop->master PR needs to opened, and that it must NOT be merged via github merge button, but via FF merge from console

amalota:
closes https://github.com/VirtusLab/git-machete/issues/395


CONTRIBUTING.md
```diff
@@ -92,7 +92,15 @@ Review fixes should be pushed on separate commits for easier viewing on GitHub (
...
 ## Release TODO list
 
-1. Merge the changes from `develop` to `master` and push `master`.
+1. Create release PR from `develop` into `master`.
+
+1. Verify that all checks have passed.
+
+1. Merge develop into master and push to remote repository using console:
+
+         git checkout master
```
PawelLipski, line 103: From my experience, it happens pretty often that ppl have their `develop`/`master` branches NOT up to date with remote when doing this FF merge... so I'd also add `git pull`s on both branches

CONTRIBUTING.md
```diff
@@ -98,6 +98,7 @@ Review fixes should be pushed on separate commits for easier viewing on GitHub (
 
 1. Merge develop into master and push to remote repository using console:
 
+         git pull origin develop master
```
PawelLipski: This does not have the effect of updating the local branches as expected :/

See the following scenario:

image

and then:

image

but still:

image

So I'd rather do sth like:

```suggestion
         git checkout develop
         git pull origin develop
         git checkout master
         git pull origin master
```

PawelLipski: In general, git (and git-machete CLI alike) gives very limited (if any) capabilities on doing **modifying operations on non-currently-checked-out branch** :/

git-machete-intellij-plugin has this capability (not without nasty hacking, mostly using `git fetch` from the current repository to itself...), which even got a mention in someone's blogpost on DZone: https://dzone.com/articles/5-jetbrain-plugins-upgrade-git-support#git-machete -> _After right-clicking on a chosen branch in the graph, you can perform git actions like rebase/push/pull on that branch without a need to switch from your current branch, which is very handy!_

amalota: Oh, my bad, i though it works correctly when i saw: 
```
 * branch            develop    -> FETCH_HEAD
 * branch            master     -> FETCH_HEAD
Already up to date.
```
great catch.

CONTRIBUTING.md
```diff
@@ -98,7 +98,10 @@ Review fixes should be pushed on separate commits for easier viewing on GitHub (
 
 1. Merge develop into master and push to remote repository using console:
 
-         git pull origin develop master
+         git checkout develop
+         git pull origin develop
+         git checkout master
+         git pull origin master
          git checkout master
```
PawelLipski: `master`'s already checked out at this point, no need to re-checkout ;)

PawelLipski (approved):

PR #410, opened 2022-02-07
Issue #396: (Option #5) Add new user with set UID, GUI

amalota:
closes https://github.com/VirtusLab/git-machete/issues/396

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -6,6 +6,7 @@
 - fixed: checking whether a branch is merged to parent works for branches that have no common commit
 - added: CI/CD check ensuring that RELEASE_NOTES are up to date
 - fixed: `github create-pr` takes the PR title from the first unique commit
+- fixed: problems with access to volume shared b/w VM and container
 
 ## New in git-machete 3.7.1
 
--- a/ci/docker-pull-or-build-and-push.sh
+++ b/ci/docker-pull-or-build-and-push.sh
@@ -11,8 +11,11 @@ export USER_ID=3000
 export GROUP_ID=3000
 sudo groupadd -g $GROUP_ID docker_user
 sudo useradd docker_user -u $USER_ID -g $GROUP_ID -m -s /bin/bash
-sudo su - docker_user
 whoami
+echo `id -un`
+echo `id -gn`
+echo `id -u`
+echo `id -g`
 
 function retry() {
   attempts=$1
--- a/ci/docker-pull-or-build-and-push.sh
+++ b/ci/docker-pull-or-build-and-push.sh
@@ -10,7 +10,7 @@ cd ci/$image_name/
 export USER_ID=3000
 export GROUP_ID=3000
 sudo groupadd -g $USER_ID docker_user
-sudo useradd docker_user -u $USER_ID -g $GROUP_ID -s /bin/bash
+sudo useradd docker_user -u $USER_ID -g $GROUP_ID -m -s /bin/bash
 
 function retry() {
   attempts=$1
--- a/ci/docker-pull-or-build-and-push.sh
+++ b/ci/docker-pull-or-build-and-push.sh
@@ -7,8 +7,8 @@ DIRECTORY_HASH=$(git rev-parse HEAD:ci/$image_name)
 export DIRECTORY_HASH
 cd ci/$image_name/
 
-export USER_ID=3000
-export GROUP_ID=3000
+export USER_ID=3001
+export GROUP_ID=3001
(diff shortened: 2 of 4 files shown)


ci/docker-pull-or-build-and-push.sh
```diff
@@ -31,7 +38,7 @@ retry 3 docker-compose --ansi never pull $image_name
 # (since v1.25.1, `docker-compose pull` is NOT failing when it can't fetch the image).
 image_tag=$(docker-compose --ansi never config | yq eval ".services.$image_name.image" -)
 docker image inspect "$image_tag" &>/dev/null || {
-  docker-compose --ansi never build --build-arg user_id="$(id -u)" --build-arg group_id="$(id -g)" $image_name
+  docker-compose --ansi never build --build-arg user_id=$USER_ID --build-arg group_id=$GROUP_ID $image_name
```
PawelLipski, line 44: For clarity: is it possible to just select the user/group with `user:` in docker-compose.yml **without** having the user previously created in the image? I'd intuitively say **no**, but I've seen a solution like that being PoCed on one of your PRs, following certain blog post 🤔

ci/docker-pull-or-build-and-push.sh
```diff
@@ -7,6 +7,13 @@ DIRECTORY_HASH=$(git rev-parse HEAD:ci/$image_name)
 export DIRECTORY_HASH
 cd ci/$image_name/
 
+USER_ID=3000
+export USER_ID
```
PawelLipski: Nit: no need to split assignment and `export` in case of trivial (non-failable) right-hand sides of the assignment like `3000`. The entire reason for splitting assignment and export is https://github.com/codacy/codacy-hadolint/blob/master/codacy-hadolint/docs/description/SC2155.md &mdash; but it only applies when right-hand side contains a failable stuff like command substition `$(...)`

ci/docker-pull-or-build-and-push.sh
amalota: Well, i tried it like it was presented in the blog and also tried different approaches but it doesn't work, i always got default root user inside the container.

ci/docker-pull-or-build-and-push.sh
```diff
@@ -7,6 +7,11 @@ DIRECTORY_HASH=$(git rev-parse HEAD:ci/$image_name)
 export DIRECTORY_HASH
 cd ci/$image_name/
 
+export USER_ID=3001
+export GROUP_ID=3001
```
PawelLipski: Any specific reason why the change from 3000 to 3001 was needed now? 🤔

amalota: No reason, just wanted to push new changes to rerun CI tests.

PawelLipski:
TBH this one feels better than #407 (less "intrusive" so to speak, no need to bake the UID/GID into Docker image tag)... I vote for this one!

Pls update `blogs/docker-ci-tricks-*` as well... I aim at keeping them a (roughly) up-to-date docs for the CI setup

ci/docker-pull-or-build-and-push.sh
```diff
@@ -7,6 +7,11 @@ DIRECTORY_HASH=$(git rev-parse HEAD:ci/$image_name)
 export DIRECTORY_HASH
 cd ci/$image_name/
 
+export USER_ID=3000
```
PawelLipski, line 10: Actually, do these two need to be exported at all?

ci/docker-pull-or-build-and-push.sh
```diff
@@ -7,6 +7,11 @@ DIRECTORY_HASH=$(git rev-parse HEAD:ci/$image_name)
...
+export GROUP_ID=3000
+sudo groupadd -g $USER_ID docker_user
```
PawelLipski:
```suggestion
sudo groupadd -g $GROUP_ID docker_user
```
Both are equal, but in general I think that was your intention!

RELEASE_NOTES.md
```diff
@@ -6,6 +6,7 @@
 - fixed: checking whether a branch is merged to parent works for branches that have no common commit
 - added: CI/CD check ensuring that RELEASE_NOTES are up to date
 - fixed: `github create-pr` takes the PR title from the first unique commit
+- fixed: problems with access to volume shared b/w VM and container
```
PawelLipski, line 9: Skip that from release notes, that doesn't concern ~mere mortals~ the users

PawelLipski: I mean, I only mention things like that when that's the sole change in the release (like 3.7.1 below)

PR #420, opened 2022-02-09
 Issue #396: (Option #3) Use root user in the container

amalota:


ci/tox/Dockerfile
```diff
@@ -33,25 +33,9 @@ ENV PYTHON_VERSION=${python_version}
...
-    && adduser --uid=${user_id} --ingroup=$(getent group $group_id | cut -d: -f1) --disabled-password ci-user
-
-USER ci-user
-
-RUN echo `id -un`
-RUN echo `id -gn`
-RUN echo `id -u`
-RUN echo `id -g`
-
 RUN $PYTHON -m pip install --user tox
```
PawelLipski, line 36: Note that now you probably don't need `--user`, nor a change to `PATH` (since you're root) &mdash; to be checked

ci/tox/Dockerfile
```diff
@@ -33,25 +33,9 @@ ENV PYTHON_VERSION=${python_version}
 ENV PYTHON=python${python_version}
 RUN apk add --no-cache gcc musl-dev # both packages are required to install mypy
 
-ARG user_id
```
PawelLipski, line 36: Crap, on my machine I still have the files visible as belonging to root after local-run.sh :///

How about you do these steps optional indeed? So that for local runs, the user of the same UID/GID as the host user is still created (so that `userns-remap` magic is not needed)

PR #441, opened 2022-02-18
Issue #437: Do not switch current branch when >1 PR is checked out

### LINKED ISSUE #437: Should `git machete github checkout-prs --by=...` switch current branch when >1 PR is checked out?
Repro (for the state of this repo as of Tue 15 Feb 16:36:18 CET 2022):

1. remove local branches `cli/checkout_prs_zsh_bash_completion` and `feature/command_clean`
2. remove the references to them from .git/machete if they were present there
3. `git machete github checkout-prs --by=amalota`

What happens is that the current branch is switched to one of the PRs:

```
(current branch: develop)

$ git machete github checkout-prs --by=amalota
branch 'cli/checkout_prs_zsh_bash_completion' set up to track 'origin/cli/checkout_prs_zsh_bash_completion'.
Switched to a new branch 'cli/checkout_prs_zsh_bash_completion'
Pull request #416 checked out at local branch cli/checkout_prs_zsh_bash_completion
branch 'feature/command_clean' set up to track 'origin/feature/command_clean'.
Switched to a new branch 'feature/command_clean'
Pull request #435 checked out at local branch feature/command_clean

(current branch: feature/command_clean)
```

Is that expected?

amalota:
closes https://github.com/VirtusLab/git-machete/issues/437

### DIFF
--- a/git_machete/client.py
+++ b/git_machete/client.py
@@ -1967,8 +1967,7 @@ class MacheteClient:
         self.__sync_annotations_to_definition_file(all_open_prs, current_user, verbose=verbose)
         if pr and len(checked_out_prs) == 1:
             self.__git.checkout(LocalBranchShortName.of(checked_out_prs[0].head))
-            if verbose:
-                print(fmt(f"Switched to local branch `{checked_out_prs[0].head}`"))
+            print(fmt(f"Switched to local branch `{checked_out_prs[0].head}`"))
 
     @staticmethod
     def __get_path_from_pr_chain(current_pr: GitHubPullRequest, all_open_prs: List[GitHubPullRequest]) -> List[LocalBranchShortName]:
--- a/git_machete/client.py
+++ b/git_machete/client.py
@@ -1966,9 +1966,9 @@ class MacheteClient:
               'Current GitHub user is ' + (current_user or '<none>'))
         self.__sync_annotations_to_definition_file(all_open_prs, current_user, verbose=verbose)
         if pr and len(checked_out_prs) == 1:
-            self.__git.checkout(LocalBranchShortName.of(pr.head))
+            self.__git.checkout(LocalBranchShortName.of(checked_out_prs[0].head))
             if verbose:
-                print(fmt(f"Switched to local branch `{pr.head}`"))
+                print(fmt(f"Switched to local branch `{checked_out_prs[0].head}`"))
 
     @staticmethod
     def __get_path_from_pr_chain(current_pr: GitHubPullRequest, all_open_prs: List[GitHubPullRequest]) -> List[LocalBranchShortName]:
--- a/git_machete/client.py
+++ b/git_machete/client.py
@@ -1,4 +1,5 @@
 import datetime
+import inspect
 import io
 import itertools
 import os
@@ -215,6 +216,8 @@ class MacheteClient:
             opt_onto: Optional[LocalBranchShortName],
             opt_as_root: bool,
             opt_yes: bool,
+            verbose: bool = True,
+            switch_head_to_branch: bool = True
             ) -> None:
         if branch in self.managed_branches:
             raise MacheteException(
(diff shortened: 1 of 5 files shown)


git_machete/tests/functional/test_machete.py
```diff
@@ -2836,13 +2840,3 @@ def test_clean(self) -> None:
...
         )
         self.assert_command(['status'], expected_status_output)
-
-def suite():
-    suite = unittest.TestSuite()
-    suite.addTest(MacheteTester('test_clean'))
-    return suite
-
-
-if __name__ == '__main__':
```
PawelLipski, line 2846: 😯 I've never scrolled that far, didn't realize it was here 😅

PawelLipski: Doesn't look like a proper way to execute tests indeed 😛

git_machete/client.py
```diff
@@ -1968,10 +1965,9 @@ def checkout_github_prs(self,
         debug('checkout_github_pr()',
```
PawelLipski: While you're here... pls update the method name in `debug` (unless there some Python magic to like `__current_method_name__`?)

git_machete/client.py
```diff
@@ -2387,6 +2383,91 @@ def __sync_before_creating_pr(self, *, opt_onto: Optional[LocalBranchShortName],
                 return
             raise MacheteException('Pull request creation interrupted.')
 
+    def __add_branch(self,
+                     *,
+                     branch: LocalBranchShortName,
+                     opt_onto: Optional[LocalBranchShortName],
+                     opt_as_root: bool,
+                     opt_yes: bool,
```
PawelLipski: Hmmm AFAICS this is always `True` right? better to refactor it away then

PawelLipski: Or rather, generally:
1. if this method is likely to be reused across `git machete github checkout-prs` and `git machete add`, let's keep it generic BUT also use it from `def add`
2. alternatively, if this method will only serve the purposes of `git machete github checkout-prs`, let's keep it as simple and UNconfigurable as possible

git_machete/tests/functional/test_machete.py
```diff
@@ -50,7 +50,7 @@ def mock_exit_script(status_code: Optional[int] = None, error: Optional[BaseExce
 
 def mock_fetch_ref(cls: Any, remote: str, ref: str) -> None:
     branch: LocalBranchShortName = LocalBranchShortName.of(ref[ref.index(':') + 1:])
-    git.create_branch(branch, get_current_commit_hash())
+    git.create_and_switch_head_to_branch(branch, get_current_commit_hash())
```
PawelLipski, line 53: Actually! while you're here, I'd add a regression test that would've captured the original issue in the first place

git_machete/tests/functional/test_machete.py
amalota: yeah, i just use it to debug the tests

git_machete/client.py
amalota: yeah, its possible

git_machete/client.py
amalota: Alright, i have chosen to go for option 1.

git_machete/client.py
```diff
@@ -1899,6 +1907,7 @@ def checkout_github_prs(self,
         if not current_user and my_opened_prs:
             raise MacheteException(
                 "Could not determine current user name, please check your token.")
+        current_function_name: str = f'{inspect.stack()[0].function}()'
```
PawelLipski, line 1910: Wow okay... pls leave this change as is, but also open an issue to apply a similar change to all invocations of `debug()` (maybe even better, let `debug` itself extract the name of the calling method from `stack()[1]` or whatnot?)

PawelLipski (approved):

PR #451, opened 2022-02-25
Issue #443: git-machete should NOT always interactively ask about sliding out non-existent branches

### LINKED ISSUE #443: In certain contexts, git-machete should NOT interactively ask about sliding out non-existent branches
For example, `git machete list managed` does ask about sliding out non-existent branch (as does probably **each** command that parses .git/machete under the hood)... but the effect is that Bash completion (try: `git machete add -o <TAB>) doesn't work as expected if there is a non-existent branch in .git/machete (seem to hang) :/

amalota:
closes #443 

### Commands that read definition file:
- add
- advance
- anno
- delete-unmanaged
- diff
- fork-point
- go
- github
- is-managed
- list
- log
- reaply
- show
- slide-out
- squash
- status (high-lvl?)
- traverse (high-lvl?)
- update

### DIFF
--- a/git_machete/cli.py
+++ b/git_machete/cli.py
@@ -531,8 +531,9 @@ def launch(orig_args: List[str]) -> None:
                     f"{machete_client.definition_file_path} is a directory "
                     "rather than a regular file, aborting")
 
+        perform_interactive_slide_out_flag = perform_interactive_slide_out(cmd)
         if cmd == "add":
-            machete_client.read_definition_file(perform_interactive_slide_out=perform_interactive_slide_out(cmd))
+            machete_client.read_definition_file(perform_interactive_slide_out=perform_interactive_slide_out_flag)
             branch = get_branch_arg_or_current_branch(cli_opts, git)
             machete_client.add(
                 branch=branch,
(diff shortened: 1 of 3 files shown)


git_machete/client.py
```diff
@@ -82,13 +82,13 @@ def __raise_no_branches_error(self) -> None:
...
 
-    def read_definition_file(self, verify_branches: bool = True) -> None:
+    def read_definition_file(self, verify_branches: bool = True, cmd: str = None) -> None:
         with open(self._definition_file_path) as file:
             lines: List[str] = [line.rstrip() for line in file.readlines()]
 
         at_depth = {}
         last_depth = -1
-
+        high_lvl_cmds = {'traverse', 'status', 'discover'}
```
PawelLipski: Hmmm I'd rather NOT pollute `read_definition_file` with the logic to **decide** on whether to slide-out the invalid branches... I'd put this logic in cli.py instead

amalota: @PawelLipski What should happen when we don't perform interactive slide-out, should we save the definition file or not?

amalota: Personally, i would save the definition file, i see no reason to keep invalid branches.

PawelLipski: Sh*t, I didn't think of that... yeah, probably removing the invalid branch + printing a warning to **stderr** (not to pollute stdout) would be best

git_machete/utils.py
```diff
@@ -257,3 +257,8 @@ def slurp_file_or_empty(path: str) -> str:
             return file.read()
     except IOError:
         return ''
+
+
+def perform_interactive_slide_out(cmd: str) -> bool:
+    high_level_commands = {'traverse', 'status'}
```
PawelLipski: Yeah, let's stick to these two for now for simplicity... time will tell if any changes need to be done here

git_machete/utils.py
```diff
@@ -257,3 +257,8 @@ def slurp_file_or_empty(path: str) -> str:
...
+    return sys.stdout.isatty() and cmd in high_level_commands
```
PawelLipski: In fact, this TTYness check is even more important

git_machete/utils.py
PawelLipski: Maybe actually I'd call the var `interactive_slide_out_safe_commands`, coz there are other commands that can be considered _high-level_

git_machete/cli.py
```diff
@@ -532,21 +532,22 @@ def launch(orig_args: List[str]) -> None:
                     "rather than a regular file, aborting")
 
         if cmd == "add":
-            machete_client.read_definition_file()
+            machete_client.read_definition_file(perform_interactive_slide_out=perform_interactive_slide_out(cmd))
```
PawelLipski: Maybe extract this var to before the `if`, and just pass it everywhere like:
```suggestion
            machete_client.read_definition_file(perform_interactive_slide_out)
```

PawelLipski (approved):

PR #466, opened 2022-03-09
Issue #457 : Add shields.io to remark's skipUrlPatterns

### LINKED ISSUE #457: If possible, configure remark-lint-no-dead-urls so that 5xx errors are ignored
For instance, now the check fails (https://app.circleci.com/pipelines/github/VirtusLab/git-machete/1670/workflows/c5170aba-9635-4939-ba48-0644649deec4/jobs/7159) coz shields.io is down, responding with 503. This shouldn't fail the check in principle.... pls check if it's possible to set up `gotOptions` so that only errors like server not responding/DNS not resolving/4xx HTTP are taken into account, but not 5xx HTTP.

If not, let's just add shields.io to `skipUrlPatterns` as a workaround

amalota:
closes #457

### DIFF
--- a/.remarkrc.yml
+++ b/.remarkrc.yml
@@ -2,5 +2,7 @@
 plugins:
   validate-links:
   lint-no-dead-urls:
-    skipUrlPatterns: ['https://pypistats.org/packages/git-machete', 'https://github.com/r-ryantm']
+    skipUrlPatterns: ['/https:\/\/img\.shields\.io/',
+                      'https://github.com/r-ryantm',
+                      'https://pypistats.org/packages/git-machete']
     gotOptions.timeout: 60000
--- a/.remarkrc.yml
+++ b/.remarkrc.yml
@@ -2,7 +2,9 @@
 plugins:
   validate-links:
   lint-no-dead-urls:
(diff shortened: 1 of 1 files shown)


.remarkrc.yml
```diff
@@ -2,5 +2,9 @@
 plugins:
   validate-links:
   lint-no-dead-urls:
-    skipUrlPatterns: ['https://pypistats.org/packages/git-machete', 'https://github.com/r-ryantm']
+    skipUrlPatterns: [
+      '/https:\/\/img\.shields\.io/',
```
PawelLipski, line 6: Does `/.../`  match against the entire URL, or just any part of the URL? I assumed the latter?

amalota: @PawelLipski Based on this docs i infered the latter: 

skipUrlPatterns {Array} - Array of String | RegExp. A list of patterns for URLs that should be skipped. Each URL will be tested against each pattern, and will be ignored if new RegExp(pattern).test(url) === true. For example, with skipUrls: [/^http:\/\/(.*)url-to-ignore\.com/, 'https://never-check.com'], links with the URLs http://www.url-to-ignore.com/foo and https://never-check.com/foo/bar will not be checked.

PawelLipski (approved):

PR #481, opened 2022-04-04
`fork-point` no longer specially treats branches merged to its parent

PawelLipski:
Related to https://github.com/VirtusLab/git-machete-intellij-plugin/pull/803

### DIFF
--- a/git_machete/client.py
+++ b/git_machete/client.py
@@ -936,11 +936,14 @@ class MacheteClient:
             if branch in self.up_branch:
                 print_line_prefix(branch, f"{utils.get_vertical_bar()} \n")
                 if opt_list_commits:
-                    if edge_color[branch] in (EscapeCodes.RED, EscapeCodes.DIM):
-                        commits: List[GitLogEntry] = self.__git.get_commits_between(fp_sha(branch), branch.full_name()) if fp_sha(branch) else []
+                    if not fp_sha(branch):
+                        # Rare case, but can happen e.g. due to reflog expiry.
+                        commits: List[GitLogEntry] = []
+                    elif edge_color[branch] == EscapeCodes.DIM:
+                        commits = []
                     elif edge_color[branch] == EscapeCodes.YELLOW:
                         commits = self.__git.get_commits_between(self.up_branch[branch].full_name(), branch.full_name())
-                    else:  # edge_color == EscapeCodes.GREEN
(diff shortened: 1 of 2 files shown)


git_machete/client.py
```diff
@@ -932,11 +932,13 @@ def print_line_prefix(branch_: LocalBranchShortName, suffix: str) -> None:
             if branch in self.up_branch:
                 print_line_prefix(branch, f"{utils.get_vertical_bar()} \n")
                 if opt_list_commits:
-                    if edge_color[branch] in (EscapeCodes.RED, EscapeCodes.DIM):
-                        commits: List[GitLogEntry] = self.__git.get_commits_between(fp_sha(branch), branch.full_name()) if fp_sha(branch) else []
+                    if fp_sha(branch) == []:
```
PawelLipski: Rel: https://github.com/VirtusLab/git-machete-intellij-plugin/pull/803 -> `backendImpl/src/main/java/com/virtuslab/gitmachete/backend/impl/GitMacheteRepository.java`

git_machete/client.py
```diff
@@ -935,11 +935,14 @@ def print_line_prefix(branch_: LocalBranchShortName, suffix: str) -> None:
             if branch in self.up_branch:
                 print_line_prefix(branch, f"{utils.get_vertical_bar()} \n")
                 if opt_list_commits:
-                    if edge_color[branch] in (EscapeCodes.RED, EscapeCodes.DIM):
-                        commits: List[GitLogEntry] = self.__git.get_commits_between(fp_sha(branch), branch.full_name()) if fp_sha(branch) else []
+                    if not fp_sha(branch):
+                        # Rare case, but can happen e.g. due to reflog expiry.
+                        commits: List[GitLogEntry] = []
+                    elif edge_color[branch] == EscapeCodes.DIM:
+                        commits = []
```
amalota, line 943: Why not just combine both cases into one `if` like:
 ```
if not fp_sha(branch) or edge_color[branch] == EscapeCodes.DIM:
                        commits: List[GitLogEntry] = []
```

PawelLipski: These 2 cases are very distinct (completely different reasons, even though the resulting commit list is empty for both), I wanted to keep them separate e.g. so that the comment can be applied to just one of them in line 939 above

PR #483, opened 2022-04-05
Issue #474: When using `git machete github checkout-prs --by=...` switch current head branch only when one PR is checked out

### LINKED ISSUE #437: Should `git machete github checkout-prs --by=...` switch current branch when >1 PR is checked out?

### LINKED ISSUE #474: `github checkout-prs --by=...` switches to ALL checked out branches one by one
Installed version from PR #435 (commit `6ee6681e33cf037e8c86b1cf340a443b4d77fcee`).

Testing on this very repo:

```
➜ 12:29 ~/git-machete develop $ g m s
  master  v3.7.2
  │ 
  └─develop  v3.8.0
➜ 12:29 ~/git-machete develop $ gb
* develop
  master
➜ 12:30 ~/git-machete develop $ g m github checkout-prs --by=amalota
Checking for open GitHub PRs...
branch 'feature/command_clean' set up to track 'origin/feature/command_clean'.
Switched to a new branch 'feature/command_clean'
Pull request #435 checked out at local branch feature/command_clean
branch 'code_quality/github_token_msg' set up to track 'origin/code_quality/github_token_msg'.
Switched to a new branch 'code_quality/github_token_msg'
Pull request #469 checked out at local branch code_quality/github_token_msg
branch 'feature/advance_with_push_to_remote' set up to track 'origin/feature/advance_with_push_to_remote'.
Switched to a new branch 'feature/advance_with_push_to_remote'
Pull request #473 checked out at local branch feature/advance_with_push_to_remote
➜ 12:30 ~/git-machete feature/advance_with_push_to_remote PR #473 (amalota) $ 
```

So... first HEAD hsa been switched to `feature/command_clean`, then to `code_quality/github_token_msg`, finally to `feature/advance_with_push_to_remote`.

This is contrary to docs in github.rst:

```
If only one PR is given, then switch the local repository's HEAD to its head branch.
```

In fact, the doc should probably say:

```
If only one PR has been checked out, then switch the local repository's HEAD to its head branch.
```

and `github checkout-prs` should behave adequately (i.e. for the sample above, it should **never** switch HEAD).

Pls provide regression tests as well.

amalota:
closes #474
closes #437

### DIFF
--- a/git_machete/client.py
+++ b/git_machete/client.py
@@ -220,7 +220,7 @@ class MacheteClient:
             opt_as_root: bool,
             opt_yes: bool,
             verbose: bool = True,
-            switch_branch: bool = True
+            switch_head_if_new_branch: bool = True
             ) -> None:
         if branch in self.managed_branches:
             raise MacheteException(
@@ -238,7 +238,7 @@ class MacheteClient:
                 msg = common_line + f"Check out `{branch}` locally?" + get_pretty_choices('y', 'N')
                 opt_yes_msg = common_line + f"Checking out `{branch}` locally..."
                 if self.ask_if(msg, opt_yes_msg, opt_yes=opt_yes, verbose=verbose) in ('y', 'yes'):
-                    self.__git.create_branch(branch, remote_branch.full_name(), switch_branch=switch_branch)
+                    self.__git.create_branch(branch, remote_branch.full_name(), switch_head=switch_head_if_new_branch)
                 else:
                     return
                 # Not dealing with `onto` here. If it hasn't been explicitly
@@ -257,7 +257,7 @@ class MacheteClient:
                         current_branch = self.__git.get_current_branch_or_none()
                         if current_branch and current_branch in self.managed_branches:
                             opt_onto = current_branch
-                    self.__git.create_branch(branch, out_of, switch_branch=switch_branch)
+                    self.__git.create_branch(branch, out_of, switch_head=switch_head_if_new_branch)
                 else:
                     return
 
@@ -1915,7 +1915,7 @@ class MacheteClient:
                             opt_onto=None,
                             opt_yes=True,
                             verbose=verbose,
-                            switch_branch=False)
+                            switch_head_if_new_branch=False)
                     else:
                         self.add(
                             branch=branch,
@@ -1923,7 +1923,7 @@ class MacheteClient:
                             opt_as_root=False,
                             opt_yes=True,
                             verbose=verbose,
-                            switch_branch=False)
+                            switch_head_if_new_branch=False)
                     if pr not in checked_out_prs:
                         print(fmt(f"Pull request `#{pr.number}` checked out at local branch `{pr.head}`"))
                         checked_out_prs.append(pr)
--- a/git_machete/client.py
+++ b/git_machete/client.py
@@ -219,7 +219,8 @@ class MacheteClient:
             opt_onto: Optional[LocalBranchShortName],
             opt_as_root: bool,
             opt_yes: bool,
-            verbose: bool = True
+            verbose: bool = True,
+            switch_branch: bool = True
             ) -> None:
         if branch in self.managed_branches:
             raise MacheteException(
@@ -237,7 +238,7 @@ class MacheteClient:
                 msg = common_line + f"Check out `{branch}` locally?" + get_pretty_choices('y', 'N')
                 opt_yes_msg = common_line + f"Checking out `{branch}` locally..."
                 if self.ask_if(msg, opt_yes_msg, opt_yes=opt_yes, verbose=verbose) in ('y', 'yes'):
-                    self.__git.create_branch(branch, remote_branch.full_name())
+                    self.__git.create_branch(branch, remote_branch.full_name(), switch_branch=switch_branch)
                 else:
                     return
                 # Not dealing with `onto` here. If it hasn't been explicitly
@@ -256,7 +257,7 @@ class MacheteClient:
                         current_branch = self.__git.get_current_branch_or_none()
                         if current_branch and current_branch in self.managed_branches:
                             opt_onto = current_branch
-                    self.__git.create_branch(branch, out_of)
+                    self.__git.create_branch(branch, out_of, switch_branch=switch_branch)
                 else:
                     return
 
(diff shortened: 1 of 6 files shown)


PawelLipski:
Let's include it in v3.8.0 as well! since it's a bug in a feature added by v3.8.0

git_machete/git_operations.py
```diff
@@ -581,7 +581,7 @@ def get_reflog(self, branch: AnyBranchName) -> List[GitReflogEntry]:
             return self.__reflogs_cached[branch]
 
     def create_branch(self, branch: LocalBranchShortName, out_of_revision: AnyRevision) -> None:
-        self._run_git("checkout", "-b", branch, out_of_revision)
+        self._run_git("branch", branch, out_of_revision)
```
PawelLipski, line 584: Whoops this will affect quite a few places in an undesired way probably 😅

amalota: @PawelLipski Yeah, i was just testing if this indeed was causing branches to be switched. I think additional flag `switch_branch` should do the trick here

PawelLipski: Better add a separate method if possible!

amalota: @PawelLipski I think implementing it in separate method is not easily doable since this method is used by another method in at least 2 different contexts so i will still need to add some flag in order to switch between functions, wdyt?

PawelLipski: Hmmm okay let's leave a flag indeed 🤔

git_machete/git_operations.py
```diff
@@ -580,8 +580,10 @@ def get_reflog(self, branch: AnyBranchName) -> List[GitReflogEntry]:
                                                          ))
             return self.__reflogs_cached[branch]
 
-    def create_branch(self, branch: LocalBranchShortName, out_of_revision: AnyRevision) -> None:
+    def create_branch(self, branch: LocalBranchShortName, out_of_revision: AnyRevision, switch_branch: bool) -> None:
         self._run_git("branch", branch, out_of_revision)
+        if switch_branch:
+            self._run_git("switch", branch)
```
PawelLipski: Note that `git switch` has only been added in git 2.23 or so

amalota: Hmm, in that case having 2 separate functions could be better.

PawelLipski: As in, you can just use `git checkout <branch>` instead!

amalota: You mean depending on git version i can do different things?

PawelLipski: Nopes!
Prior to 2.23: `git checkout <branch>` is the sole command for switching current branch
2.23 and beyond: both `git checkout <branch>` and `git switch <branch>` do **the same thing**

That's just it

PawelLipski: Just a matter of git CLI changing... but the capabilities if it comes to switching branches are the same

git_machete/client.py
```diff
@@ -219,7 +219,8 @@ def add(self,
             opt_onto: Optional[LocalBranchShortName],
             opt_as_root: bool,
             opt_yes: bool,
-            verbose: bool = True
+            verbose: bool = True,
+            switch_branch: bool = True
```
PawelLipski: The current name will be misleading, as the switch only happens when the branch is **newly created**, not when it's already existed before (lines 264 onwards)

git_machete/git_operations.py
amalota: @PawelLipski I am still not sure what you are proposing :D Are you proposing to use git checkout / switch depending on git version?

PawelLipski: Nope, just:
```suggestion
            self._run_git("checkout", branch)
```
This way we're compatible with every git version in existence! as opposed to `git switch`, which is git 2.23+ only

amalota: Ok, i think i finally got it - i can just use `checkout` instead of `switch` since it does the same thing :D

git_machete/client.py
amalota: @PawelLipski I see, maybe `switch_head_to_new_branch` instead?

PawelLipski: Okay!

PawelLipski: Or even: `switch_head_if_new_branch`, to make it clear that it's not always that a new branch will be created

git_machete/client.py
```diff
@@ -219,7 +219,8 @@ def add(self,
             opt_onto: Optional[LocalBranchShortName],
             opt_as_root: bool,
             opt_yes: bool,
-            verbose: bool = True
+            verbose: bool = True,
```
PawelLipski: While you're here, I'd avoid default params here, let's make the caller always decide explicitly what they want to avoid surprises

git_machete/tests/functional/test_machete.py
```diff
@@ -50,7 +50,7 @@ def mock_exit_script(status_code: Optional[int] = None, error: Optional[BaseExce
 
 def mock_fetch_ref(cls: Any, remote: str, ref: str) -> None:
     branch: LocalBranchShortName = LocalBranchShortName.of(ref[ref.index(':') + 1:])
-    git.create_branch(branch, get_current_commit_hash())
+    git.create_branch(branch, get_current_commit_hash(), switch_head=False)
     git.checkout(branch)
```
PawelLipski: Nit: can be squeezed into
```suggestion
    git.create_branch(branch, get_current_commit_hash(), switch_head=True)
```

PawelLipski (approved):

PawelLipski (approved):

PR #487, opened 2022-04-08
Issue #479: `git machete github *` should work with custom GitHub URL

### LINKED ISSUE #479: `git machete github anno-prs` doesn't work with custom GitHub URL
# Overview

I've stumbled upon some issues with GitHub recently: _sometimes_ for _some_ repos, `git push` wouldn't work. With trial-and-error, I found a reliable workaround for me:

```shellsession
$ git remote -v
origin	github:VirtuslabRnD/nfs-core.git (fetch)
origin	github:VirtuslabRnD/nfs-core.git (push)

$ cat ~/.ssh/config
Host github
    Hostname        ssh.github.com
    Port            443
    User            git
    IdentityFile    ~/.ssh/id_ed25519
    IdentitiesOnly  yes
```

However, this setup breaks branch discovery:
```shellsession
$ git machete github anno-prs -v
git rev-parse --git-dir
git for-each-ref --format=%\(refname\)$'\t'%\(objectname\)$'\t'%\(tree\)$'\t'%\(committerdate:raw\) refs/remotes
git for-each-ref --format=%\(refname\)$'\t'%\(objectname\)$'\t'%\(tree\)$'\t'%\(committerdate:raw\)$'\t'%\(upstream\) refs/heads
git remote
git config --get remote.origin.url
Remotes are defined for this repository, but none of them corresponds to GitHub (see git remote -v for details)
```

```shellsession
$ git --version
git version 2.35.1

$ git machete --version
git machete version 3.7.2
```

amalota:
closes #479


git_machete/git_operations.py
```diff
@@ -337,6 +337,10 @@ def get_remotes(self) -> List[str]:
     def get_url_of_remote(self, remote: str) -> str:
         return self._popen_git("config", "--get", f"remote.{remote}.url").strip()  # 'git remote get-url' method has only been added in git v2.5.1
 
+    def get_organization_and_repository_name_of_remote(self, remote: str) -> Tuple[str, str]:
+        return self._popen_git("config", "--get", f"machete.github.{remote}.organization").strip(), \
+               self._popen_git("config", "--get", f"machete.github.{remote}.repository").strip()
```
PawelLipski: Actually! we've got `get_config_attr_or_none` above to avoid multiple calls to `git config --get`... pls replace here and wherever else found

git_machete/git_operations.py
```diff
@@ -337,6 +337,14 @@ def get_remotes(self) -> List[str]:
     def get_url_of_remote(self, remote: str) -> str:
         return self._popen_git("config", "--get", f"remote.{remote}.url").strip()  # 'git remote get-url' method has only been added in git v2.5.1
 
+    def get_organization_and_repository_name_of_remote(self, remote: str) -> Tuple[str, str]:
+        org = self.get_config_attr_or_none(f"machete.github.{remote}.organization")
+        repo = self.get_config_attr_or_none(f"machete.github.{remote}.repository")
+        if org and repo:
+            return org.strip(), repo.strip()
+        else:
+            raise MacheteException(f'Organization and repository name is unset for {remote}')
```
amalota: @PawelLipski not sure if we want exception here or just returning None is enough, wdyt?

amalota: Hmm, i think that since this is not mandatory, exception might be too much, so i personally would just return None

PawelLipski: Yeah, `None` FTW

git_machete/client.py
```diff
@@ -1363,8 +1363,13 @@ def sync_annotations_to_github_prs(self) -> None:
         if not url_for_remote:
             raise MacheteException(fmt('No remotes defined for this repository (see `git remote`)'))
 
-        optional_org_name_for_github_remote: Dict[str, Optional[Tuple[str, str]]] = {
-            remote: get_parsed_github_remote_url(url) for remote, url in url_for_remote.items()}
+        optional_org_name_for_github_remote: Dict[str, Optional[Tuple[str, str]]] = {}
```
PawelLipski: While you're here... I'd replace tuple here with some simple class with 2 named fields

amalota: @PawelLipski Yeah, was thinking the same but i just wanted to use namedTuple instead of class, wdyt?

PawelLipski: Oh! `NamedTuple`, I forgot about it &mdash; yeah, sounds like a perfect choice here

git_machete/client.py
```diff
@@ -1363,16 +1363,17 @@ def sync_annotations_to_github_prs(self) -> None:
         if not url_for_remote:
             raise MacheteException(fmt('No remotes defined for this repository (see `git remote`)'))
 
-        optional_org_name_for_github_remote: Dict[str, Optional[Tuple[str, str]]] = {}
+        optional_org_name_for_github_remote: Dict[str, OrganizationRepository] = {}
```
PawelLipski: Maybe `OrganizationAndRepository`?

git_machete/git_operations.py
```diff
@@ -165,6 +165,11 @@ class BranchPair(NamedTuple):
     local_or_remote_branch: AnyBranchName
 
 
+class OrganizationRepository(NamedTuple):
```
PawelLipski: This is GitHub-specific and should probably live in github.py. Vanilla git doesn't know anything about _organizations_.

git_machete/git_operations.py
```diff
@@ -335,15 +341,11 @@ def get_remotes(self) -> List[str]:
...
 
-    def get_organization_and_repository_name_of_remote(self, remote: str) -> Tuple[str, str]:
-        org = self.get_config_attr_or_none(f"machete.github.{remote}.organization")
-        repo = self.get_config_attr_or_none(f"machete.github.{remote}.repository")
-        if org and repo:
-            return org.strip(), repo.strip()
-        else:
-            raise MacheteException(f'Organization and repository name is unset for {remote}')
+    def get_organization_and_repository_name_of_remote(self, remote: str) -> OrganizationRepository:
+        return OrganizationRepository(organization=self.get_config_attr_or_none(f"machete.github.{remote}.organization"),
```
PawelLipski: Hmmm I'm not sure yet.... but shouldn't `machete.github.organization` and `machete.github.repository` be a "global" key in the sense that they just refer to a single GitHub repository used for all `git machete github` operations performed on the given local repository?

PawelLipski: Or in other words, making these keys remote-dependent (i.e. you can have `machete.github.origin.organization` and `machete.github.my-fork.organization`) gives more elasticity... but does the user really need to run `git machete github` against more than one GitHub repo in the given local repo?

PawelLipski: Note that e.g. `gh` CLI has this exact single-Github-repository logic: if your remotes correspond to multiple GitHub repositories, it lets you select the single GitHub remote to perform CLI operations agaist

git_machete/client.py
```diff
@@ -1383,34 +1383,41 @@ def is_excluded_reflog_subject(sha_: str, gs_: str) -> bool:
               "and branch reset events irrelevant for fork point/upstream inference): %s\n" % (", ".join(result) or "<empty>"))
         return result
 
+    def get_remote_name_and_organization_and_repository_name_of_remote(self) -> 'RemoteAndOrganizationAndRepository':
+        return RemoteAndOrganizationAndRepository(remote=self.__git.get_config_attr_or_none(f"machete.github.remote"),
```
PawelLipski: Okay! if you could provide docs for that config... I think it should be treated similarly as the docs for GitHub authentication, i.e. shared b/w multiple locations (`include`d in .rst)

git_machete/git_operations.py
```diff
@@ -312,7 +312,8 @@ def __ensure_config_loaded(self) -> None:
 
     def get_config_attr_or_none(self, key: str) -> Optional[str]:
         self.__ensure_config_loaded()
-        return self.__config_cached.get(key.lower())
+        value = self.__config_cached.get(key.lower())
+        return value.strip() if value else value
```
PawelLipski: Hmmm I wouldn't indiscriminately `strip` all retrieved config values... I can't now think of any that include leading/trailing whitespace on purpose, but still, that might be error-prone

PawelLipski: I'd rather `strip` case-by-case, as in `get_strict_remote_for_fetching_of_branch`

git_machete/client.py
```diff
@@ -1383,34 +1383,41 @@ def is_excluded_reflog_subject(sha_: str, gs_: str) -> bool:
...
+                    fmt('Remotes are defined for this repository, but none of them '
+                        'corresponds to GitHub (see `git remote -v` for details)'))
+            if len(org_name_for_github_remote) == 1:
+                _, org, repo = list(org_name_for_github_remote.values())[0]
+            elif len(org_name_for_github_remote) > 1:
+                if 'origin' in org_name_for_github_remote:
+                    _, org, repo = org_name_for_github_remote['origin']
+                else:
+                    raise MacheteException(
+                        f'Multiple non-origin remotes correspond to GitHub in this repository: '
```
PawelLipski: Here also the info about `machete.github.*` keys should be displayed to the user

git_machete/client.py
```diff
@@ -1383,34 +1383,41 @@ def is_excluded_reflog_subject(sha_: str, gs_: str) -> bool:
...
+            optional_org_name_for_github_remote: Dict[str, Optional[RemoteAndOrganizationAndRepository]] = {
+                remote: get_parsed_github_remote_url(url, remote) for remote, url in url_for_remote.items()}
+            org_name_for_github_remote: Dict[str, RemoteAndOrganizationAndRepository] = {remote: org_name for remote, org_name in
+                                                                                         optional_org_name_for_github_remote.items() if
+                                                                                         org_name}
+            if not org_name_for_github_remote:
                 raise MacheteException(
-                    f'Multiple non-origin remotes correspond to GitHub in this repository: '
-                    f'{", ".join(org_name_for_github_remote.keys())}, aborting')
+                    fmt('Remotes are defined for this repository, but none of them '
```
PawelLipski: (and here 👇🏻)

git_machete/client.py
amalota: @PawelLipski Hmm, this part is inside the if-else statement that is used when the keys are not provided and the old approach is being used so i don't think this is the right place for it, wdyt?

git_machete/client.py
amalota: @PawelLipski same situation as above

amalota:
@PawelLipski FYI docs look like this:
image

docs/source/github_config_keys.rst
```diff
@@ -0,0 +1,21 @@
+.. _github_config_keys:
+
+ .. note::
+
+    In order to use custom GitHub repository URL, you need to set 3 additional, local git keys:
```
PawelLipski:
```suggestion
        git-machete will try to infer the GitHub repository to fire GitHub API requests against from `git remote` data.
        To override this default behavior and point to the target repository explicitly, you need to set 3 additional, local git config keys:
```

docs/source/github_config_keys.rst
```diff
@@ -0,0 +1,21 @@
...
+
+        1. Remote name ``machete.github.remote``
+        2. Organization name ``machete.github.organization``
+        3. Repository name ``machete.github.repository``
+
+    You can do it in 2 ways:
+
+        1. Set each key separately with ``machete.github.<key_name> "<key_value>"``
```
PawelLipski:
```suggestion
        1. Set each key separately with ``git config machete.github.<key_name> "<key_value>"``
```

docs/source/cli_help/github.rst
```diff
@@ -20,6 +20,8 @@ Creates, checks out and manages GitHub PRs while keeping them reflected in branc
   Any existing annotations are overwritten for the branches that have an opened PR; annotations for the other branches remain untouched.
   Equivalent to ``git machete anno --sync-github-prs``.
 
+.. include:: github_config_keys.rst
```
PawelLipski: Hmmm now it looks like it only applies to `anno-prs`... while in fact this is common to all subcommands 🤔 note that it also applies to `clean -H` and `anno -H`, I'd expect it to be mentioned there as well

git_machete/client.py
PawelLipski: Okay... but isn't that one of the cases where the user might want to fall back to `machete.github.*` config keys? 🤔 as in, they see that _Multiple non-origin remotes correspond to GitHub in this repository_... but how to the pick up the right GitHub repo? I'd say that's a perfect moment to inform about this workaround

git_machete/client.py
PawelLipski: I mean... that's exactly Konrad's case from the original issue 😅 so informing about the config keys would be a good idea

docs/source/cli_help/github.rst
```diff
@@ -20,6 +20,8 @@ Creates, checks out and manages GitHub PRs while keeping them reflected in branc
...
+
```
PawelLipski, line 24: Pls also replicate the changes in docs.py... in fact I think you can use a shared string there for both `github_config_keys` and `github_api_access` to avoid repetition

git_machete/client.py
amalota: Yeah, i see your point now :D

docs/source/cli_help/github.rst
amalota: Hmm, i am not 100% sure but i think it only applies to `github anno-prs` and as you rightly mentioned `anno -H` but AFAICS `clean -H` doesn't use recently modified `sync_annotations_to_github_prs` function, @PawelLipski am i right?

PawelLipski: Oh okay, I see the source of confusion... I mean, the original Konrad's issue was indeed just about `git machete github anno-prs`... but I'd say the newly-added config should apply to all places where some interaction with GitHub API happens! Which includes all `github` subcommands, as well as `clean -H` and `anno -H`

docs/source/github_config_keys.rst
```diff
@@ -2,15 +2,16 @@
 
  .. note::
 
-    In order to use custom GitHub repository URL, you need to set 3 additional, local git keys:
+        git-machete will try to infer the GitHub repository to fire GitHub API requests against from `git remote` data.
```
PawelLipski: Is the indent correct now? 🤔

PawelLipski:
(renamed the PR to reflect the scope of the change more accurately, as not only `git machete github anno-prs` should be affected)

docs/source/github_config_keys.rst
amalota: No, I think you removed part of the indent, its fixed now.

amalota: Also i removed **from** from `git-machete will try to infer the GitHub repository to fire GitHub API requests against from 'git remote' data.` so its `git-machete will try to infer the GitHub repository to fire GitHub API requests against 'git remote' data.`, sounds better to me

PawelLipski: Or even clearer:

```
git-machete will use ``git remote`` data to infer the GitHub repository to fire GitHub API requests against.

File: docs/source/cli_help/anno.rst [hunk 16]
```diff
@@ -23,6 +23,7 @@ If invoked with ``-H`` or ``--sync-github-prs``, annotates the branches based on
 Any existing annotations are overwritten for the branches that have an opened PR; annotations for the other branches remain untouched.
 
 .. include:: github_api_access.rst
+.. include:: github_config_keys.rst
```
PawelLipski, line 26: Pls show a screenshot it looks like now in the context

git_machete/client.py
```diff
@@ -1383,41 +1383,8 @@ def is_excluded_reflog_subject(hash_: str, gs_: str) -> bool:
...
         return result
 
-    def get_remote_name_and_organization_and_repository_name_of_remote(self) -> 'RemoteAndOrganizationAndRepository':
-        return RemoteAndOrganizationAndRepository(remote=self.__git.get_config_attr_or_none("machete.github.remote"),
-                                                  organization=self.__git.get_config_attr_or_none("machete.github.organization"),
-                                                  repository=self.__git.get_config_attr_or_none("machete.github.repository"))
-
     def sync_annotations_to_github_prs(self) -> None:
-
-        url_for_remote: Dict[str, str] = {remote: self.__git.get_url_of_remote(remote) for remote in
```
PawelLipski, line 1390: Wow, good riddance

git_machete/docs.py
```diff
@@ -30,6 +30,34 @@
...
 
+github_api_access = '''To allow GitHub API access for private repositories (and also to perform side-effecting actions like opening a PR,
+        even in case of public repositories), a GitHub API token with `repo` scope is required, see https://github.com/settings/tokens.
+        This will be resolved from the first of:
+
+            1. `GITHUB_TOKEN` env var,
+            2. content of the `.github-token` file in the home directory (`~`),
+            3. current auth token from the `gh` GitHub CLI,
+            4. current auth token from the `hub` GitHub CLI.
+    '''
```
PawelLipski: Nit: redundant whitespace

git_machete/docs.py
```diff
@@ -126,12 +154,8 @@
         If invoked with `-H` or `--sync-github-prs`, annotates the branches based on their corresponding GitHub PR numbers and authors.
         Any existing annotations are overwritten for the branches that have an opened PR; annotations for the other branches remain untouched.
 
-        To allow GitHub API access for private repositories (and also to perform side-effecting actions like opening a PR, even in case of public repositories),
-        a GitHub API token with `repo` scope is required, see `https://github.com/settings/tokens`. This will be resolved from the first of:
-        1. `GITHUB_TOKEN` env var,
-        2. content of the `.github-token` file in the home directory (`~`),
-        3. current auth token from the `gh` GitHub CLI,
-        4. current auth token from the `hub` GitHub CLI.
+        {github_api_access}
```
PawelLipski: 👍🏻

git_machete/client.py
```diff
@@ -2065,6 +2032,10 @@ def retarget_github_pr(self, head: LocalBranchShortName) -> None:
             print(fmt(f'The base branch of PR #{pr.number} is already `{new_base}`'))
 
     def __derive_remote_and_github_org_and_repo(self) -> RemoteAndOrganizationAndRepository:
+        remote_and_organization_and_repository_for_custom_url = self.__get_remote_and_organization_and_repository_name_for_custom_url()
```
PawelLipski, line 2038: Pls provide some simple test that would actually retrieve the remote/org/repo data from config keys and NOT from `git remote` data

git_machete/client.py
```diff
@@ -2077,7 +2048,9 @@ def __derive_remote_and_github_org_and_repo(self) -> RemoteAndOrganizationAndRep
         if not remote_and_github_org_and_repo:
             raise MacheteException(
                 fmt('Remotes are defined for this repository, but none of them '
-                    'corresponds to GitHub (see `git remote -v` for details)\n'))
+                    'corresponds to GitHub (see `git remote -v` for details). It is possible that you are using custom GitHub URL, '
```
PawelLipski: Add some newline in this message coz it'll be very long => hard to read

docs/source/cli_help/anno.rst
amalota: @PawelLipski 
image

PawelLipski: Ok! gr8

docs/source/github_config_keys.rst
```diff
@@ -0,0 +1,22 @@
+.. _github_config_keys:
+
+ .. note::
+
+    git-machete will try to infer the GitHub repository to fire GitHub API requests against ``git remote`` data.
```
kgadek: Editorial suggestion:

```suggestion
    git-machete will try to infer GitHub API server URL from ``git remote``.
```

docs/source/github_config_keys.rst
```diff
@@ -0,0 +1,22 @@
...
+
+        1. Set each key separately with ``git config machete.github.<key_name> "<key_value>"``
+        2. Edit config file with ``git config --edit`` and add the keys like its suggested below
+
+            .. code-block:: shell
+
+                [machete "github"]
+                    organization = <organization_name>
+                    repository = <repo_name>
+                    remote = <remote_name>
```
kgadek: I suggest the following changes:

- add `--local` to `git config --edit` invocation — this will make the command fail if invoked from outside of the repo. I think that in most cases such entry in a global config file would be a mistake,
- used the RST's [definition list](https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#definition-lists) for some nice formatting
- added example…
- **…but please check if this is right, I'm not sure if `remote` is github.com or github.com/ or https://github.com or what**
- (probably very personal opinion) _I think `git config --edit` covers most cases and if somebody has a need for separate invocations then probably they already know how to do that. Thus, I would trade that point for shorter text ("There should be one-- and preferably only one --obvious way to do it" or sth)_
- code-block syntax coloring was wrong, [should be gitconfig](https://github.com/github/linguist/blob/master/lib/linguist/languages.yml)

```suggestion
    You can override this by setting the following git config keys:
    
    ``machete.github.remote``
        Remote name. Example: ``github.com``
    ``machete.github.organization``
        Organization name. Example: ``VirtusLab``.
    ``machete.github.repository``
        Repository name. Example: ``git-machete``.
        
    To do this, run ``git config --local --edit`` and add the following section:
    
    .. code-block:: gitconfig

        [machete "github"]
            organization = <organization_name>
            repository = <repo_name>
            remote = <remote_name>
```

git_machete/docs.py
```diff
@@ -30,6 +30,34 @@
...
+        You can do it in 2 ways:
+
+            1. Set each key separately with `git config machete.github.<key_name> "<key_value>"`
+            2. Edit config file with `git config --edit` and add the keys like its suggested below
+
+                [machete "github"]
+                    organization = <organization_name>
+                    repository = <repo_name>
+                    remote = <remote_name>
+'''
```
kgadek: See my comments for `docs/source/github_config_keys.rst`.

git_machete/docs.py
kgadek: This string has an embedded indentation, which is probably fine, but maybe it would be better in the long run to dedent this and use `textwrap.indent(…)` at use-site?

Same would apply to `github_config_keys`.

NOTE: In the suggestion here, I haven't applied the other comment from `docs/source/github_config_keys.rst`.

```suggestion
github_api_access = '''To allow GitHub API access for private repositories (and also to perform side-effecting actions like opening a PR,
even in case of public repositories), a GitHub API token with `repo` scope is required, see https://github.com/settings/tokens.
This will be resolved from the first of:

    1. `GITHUB_TOKEN` env var,
    2. content of the `.github-token` file in the home directory (`~`),
    3. current auth token from the `gh` GitHub CLI,
    4. current auth token from the `hub` GitHub CLI.
'''
```

git_machete/docs.py
kgadek: Pt. 2 of `dedent`/`indent` ~~_saga_~~ suggestion:

```suggestion
{textwrap.indent(github_api_access, "        ")}
```

kgadek (commented):
Per request, I've looked at the docs. Looks good overall, but I'm bit nit-picky 😅 

My only actual gripe is the lack of example, as I am not sure what value to set for `machete.github.organization`.

PawelLipski:
> Per request, I've looked at the docs. Looks good overall, but I'm bit nit-picky sweat_smile

benissimo

As usual, best reviews out there coming from Konrad

amalota:
Yes, excellent review, thanks Konrad!

amalota:
@PawelLipski @kgadek Result:
image

RELEASE_NOTES.md
```diff
@@ -1,5 +1,9 @@
 # Release notes
 
+## New in git-machete 3.9.2
```
PawelLipski: 3.10.0! as in the milestone number

git_machete/__init__.py
```diff
@@ -1 +1 @@
-__version__ = '3.9.1'
+__version__ = '3.9.2'
```
PawelLipski: Ditto

git_machete/client.py
```diff
@@ -2061,7 +2062,7 @@ def __derive_remote_and_github_org_and_repo(self) -> RemoteAndOrganizationAndRep
 
         raise MacheteException(
             f'Multiple non-origin remotes correspond to GitHub in this repository: '
-            f'{", ".join(remote_and_github_org_and_repo.keys())}, aborting.'
+            f'{", ".join(remote_and_github_org_and_repo.keys())}, aborting. \n'
             f' You can also select the repository by providing these 3 git config keys: '
```
```suggestion
            f'You can also select the repository by providing these 3 git config keys: '
```

tests/test_github.py
```diff
@@ -202,6 +202,36 @@ def test_github_anno_prs(self) -> None:
             """,
         )
 
+        # Test anno-prs using custom remote URL provided by git config keys
+        (
+            self.repo_sandbox.add_git_config_key('machete.github.remote', 'custom_origin')
```
PawelLipski: For clarity, will the test still pass if these `add_git_config_key` were not present? 🤔

docs/source/github_config_keys.rst
```diff
@@ -0,0 +1,23 @@
+.. _github_config_keys:
+
+ .. note::
+
+    **git-machete** will try to infer GitHub API server URL from ``git remote``.
```
PawelLipski: Nit: I don't think bold for `git-machete` is justfiied

git_machete/docs.py
```diff
@@ -327,18 +344,15 @@
...
 
-        To allow GitHub API access for private repositories (and also to perform side-effecting actions like opening a PR, even in case of public repositories),
-        a GitHub API token with `repo` scope is required, see `https://github.com/settings/tokens`. This will be resolved from the first of:
-        1. `GITHUB_TOKEN` env var,
-        2. content of the .github-token file in the home directory (`~`),
-        3. current auth token from the `gh` GitHub CLI,
-        4. current auth token from the `hub` GitHub CLI.
+{textwrap.indent(github_api_access, "        ")}
+
+{textwrap.indent(github_config_keys, "        ")}
```
PawelLipski, line 356: Nit: I'd add an extra empty line here, before the subcommand help sections begin

git_machete/docs.py
```diff
@@ -154,13 +175,9 @@
         No branch will be deleted unless explicitly confirmed by the user (or unless `-y/--yes` option is passed).
         Equivalent of `git machete github sync` if invoked with `-H` or `--checkout-my-github-prs`.
 
-        To allow GitHub API access for private repositories (and also to perform side-effecting actions like opening a PR, even in case of public repositories),
-        a GitHub API token with `repo` scope is required, see https://github.com/settings/tokens. This will be resolved from the first of:
+{textwrap.indent(github_api_access, "        ")}
```
PawelLipski: Whoops sth broken here:

image

PawelLipski: (as mentioned previously, this `<b>` is redundant here anyway IMO)

git_machete/docs.py
```diff
@@ -142,7 +163,7 @@
           <b>-b, --branch=<branch></b>      Branch to set the annotation for.
           <b>-H, --sync-github-prs</b>      Annotate with GitHub PR numbers and authors where applicable.
     """,
-    "clean": """
+    "clean": f"""
         <b>Usage:
```
PawelLipski: Oh! here's the problem:

image

Unclosed tag

git_machete/docs.py
```diff
@@ -154,13 +175,9 @@
...
 
-        To allow GitHub API access for private repositories (and also to perform side-effecting actions like opening a PR, even in case of public repositories),
-        a GitHub API token with `repo` scope is required, see https://github.com/settings/tokens. This will be resolved from the first of:
+{textwrap.indent(github_api_access, "        ")}
 
-            1. `GITHUB_TOKEN` env var,
-            2. content of the `.github-token` file in the home directory (`~`),
-            3. current auth token from the `gh` GitHub CLI,
-            4. current auth token from the `hub` GitHub CLI.
+{textwrap.indent(github_config_keys, "        ")}
```
PawelLipski: TBH, in docs.py I'd add the extra 2 spaces of indent for these injected sections, to make it clear they're more of _side notes_ rather than _core docs_ for the command

PawelLipski: TBH, in docs.py I'd add the extra 2 spaces of indent for these injected sections, to make it clear they're more of _side notes_ rather than _core docs_ for the command

PawelLipski: In .rst, it's already ensured by the blue-ish frame/tooltip/note/whatnot

tests/test_github.py
amalota: Yes, the test case just above this one handles that

PawelLipski: Hmmm then IIUC the test doesn't really serve its purpose... the entire `machete.github.*` feature could be **un**implemented and the test will still pass! ☹️

PawelLipski: Pls make sure that the tests is written in such a way the setting of `machete.github.*` config keys is a necessary condition for it to pass

docs/source/github_config_keys.rst
amalota: @PawelLipski While reading, it made it easier for me to instantly realise that it's the app name but i am open to revert to non-bold

git_machete/docs.py
amalota: @PawelLipski Can you try again, it works fine for me:
image

git_machete/docs.py
amalota: sure

git_machete/docs.py
amalota: Great catch, thx

docs/source/github_config_keys.rst
PawelLipski: You can also change to passive voice:
```suggestion
    GitHub API server URL will be inferred from ``git remote``.
```

git_machete/docs.py
PawelLipski: Probably will get fixed once `<b>` tag above is sorted out... will check

RELEASE_NOTES.md
```diff
@@ -1,6 +1,6 @@
 # Release notes
 
-## New in git-machete 3.9.2
+## New in git-machete 3.10.0
 
 - added: `git machete github *` subcommands now work with custom GitHub URL
```
PawelLipski: Pls also mention what config keys have been added explicitly, similarly my comment on #511

tests/test_github.py
amalota: @PawelLipski Hmm, i am not sure that we are on the same page here, to summarise: **first** part of the test case is as before, checks if it works without setting the keys since they are not mandatory, **second** part of the test tests the same scenario as the first one but this time using the newly added config keys, wdyt?

amalota: So  the **second** part of the test case will only pass if the keys are present

amalota: Oh wait, i think i understand what you mean, the **second** part of the test case will pass even when the keys are not present, which is pointless, i'll fix it asap

RELEASE_NOTES.md
```diff
@@ -1,8 +1,9 @@
 # Release notes
 
-## New in git-machete 3.9.2
+## New in git-machete 3.10.0
 
-- added: `git machete github *` subcommands now work with custom GitHub URL
+- added: 3 git config keys `machete.github.{remote,organization,repository}` that enables `git machete github *` subcommands to work
+  with custom GitHub URLs
```
PawelLipski: Crap, I'm not sure if magic script that generates a changelog for Debian from RELEASE_NOTES.md will work properly with a broken line (the Debian changelog format is brittle as hell)... pls merge to the line above

RELEASE_NOTES.md
```diff
@@ -1,8 +1,9 @@
 # Release notes
 
-## New in git-machete 3.9.2
+## New in git-machete 3.10.0
 
-- added: `git machete github *` subcommands now work with custom GitHub URL
+- added: 3 git config keys `machete.github.{remote,organization,repository}` that enables `git machete github *` subcommands to work
```
```suggestion
- added: 3 git config keys `machete.github.{remote,organization,repository}` that enable `git machete github *` subcommands to work
```

git_machete/client.py
```diff
@@ -2058,30 +2032,44 @@ def retarget_github_pr(self, head: LocalBranchShortName) -> None:
...
         }
-        if not org_and_repo_for_github_remote:
+        if not remote_and_github_org_and_repo:
             raise MacheteException(
                 fmt('Remotes are defined for this repository, but none of them '
-                    'corresponds to GitHub (see `git remote -v` for details)\n'))
+                    'corresponds to GitHub (see `git remote -v` for details). \n'
+                    'It is possible that you are using custom GitHub URL, '
+                    'if that is the case, you can provide repository information explicitly, via these 3 git config keys: '
+                    'machete.github.{remote,organization,repository}\n'))
```
```suggestion
                    'It is possible that you are using custom GitHub URL.\n'
                    'If that is the case, you can provide repository information explicitly, via 3 git config keys: '
                    '`machete.github.{remote,organization,repository}`\n'))
```

git_machete/client.py
```diff
@@ -2058,30 +2032,44 @@ def retarget_github_pr(self, head: LocalBranchShortName) -> None:
...
-            return 'origin', org_and_repo_for_github_remote['origin']
+        if 'origin' in remote_and_github_org_and_repo:
+            return remote_and_github_org_and_repo['origin']
 
         raise MacheteException(
             f'Multiple non-origin remotes correspond to GitHub in this repository: '
-            f'{", ".join(org_and_repo_for_github_remote.keys())}, aborting')
+            f'{", ".join(remote_and_github_org_and_repo.keys())}, aborting. \n'
+            f'You can also select the repository by providing these 3 git config keys: '
+            'machete.github.{remote,organization,repository}\n')
```
```suggestion
            f'You can also select the repository by providing 3 git config keys: '
            '`machete.github.{remote,organization,repository}`\n')
```

PawelLipski:
Tested locally, all seems to work as expected 🎉 Minor comments, almost ready to approve

PawelLipski (approved):

Issue #232, opened 2021-09-14
Split `test_machete.py` into command-specific test files (`test_github.py`, `test_show.py` etc.)


PawelLipski:

AABur:
I like to implement this task.

But first, I suggest connecting [Codecov](https://about.codecov.io/) to control code coverage by tests.
CircleCI doesn't provide the ability to control code coverage. 
The connection of Codecov is straightforward with GitHub Actions.
You can see the result in my [fork](https://github.com/AABur/git-machete/tree/develop). An appropriate `badge` is added there. Here you can see what [Codecov Report](https://github.com/AABur/git-machete/pull/3#issuecomment-1100237043) looks like.

Firstly, it will avoid breaking or losing some of the tests.
Secondly, it will be easy to control the coverage of the code by the tests in future updates.

PawelLipski:
Wow I see your fork, nice 🚀 

Actually, we've got _some_ automation for code coverage: `coverage` results are upload as artifacts on each CircleCI build, you can check e.g. [latest develop build's Artifacts](https://app.circleci.com/pipelines/github/VirtusLab/git-machete/1855/workflows/eb949dbb-6996-4586-a350-7165f7110970/jobs/8312/artifacts) -> [coverage/index.html](https://output.circle-artifacts.com/output/job/b7de3a97-2d6c-42c0-ac0a-0d9edb219188/artifacts/0/coverage/index.html)

But Codecov is much smoother indeed. Would you mind opening a pull request from your fork?

Also! how did you learn about this project, was it via [goodfirstissues.com](https://goodfirstissues.com/) ?

AABur:
Regarding CircleCI, I understand how it is realised. But it's nothing more than the storage of a coverage report. And it's not clear at all. Codecov provides code coverage checks that are more straightforward and transparent. Especially for Pull Request.

Yes! I found this project on [goodfirstissues.com](https://goodfirstissues.com/).

PawelLipski:
Okay! feel free to proceed with a PR

AABur:
TODO:

- [x] refactor `test_machete.py` in **pytest** style -> DONE #498 

- [x] move common classes and  functions into separate module -> DONE #500

split `test_machete.py` into command-specific test files
- [x] add #502 
- [x] advance #512
- [ ] anno
- [x] clean #503
- [ ] delete-unmanaged
- [ ] diff
- [ ] discover
- [ ] edit
- [ ] file
- [ ] fork-point
[...]

PR #501, opened 2022-04-27
test_update

AABur:
RE: #232 

Test for the `update` command

### DIFF
--- a/git_machete/tests/functional/test_update.py
+++ b/git_machete/tests/functional/test_update.py
@@ -1,0 +1,271 @@
+import io
+import os
+import re
+import subprocess
+import sys
+import textwrap
+from contextlib import redirect_stderr, redirect_stdout
+from typing import Any, Dict, Iterable, Optional
+from unittest import mock
+
+import pytest  # type: ignore
+from git_machete import cli
+from git_machete.git_operations import LocalBranchShortName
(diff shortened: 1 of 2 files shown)


git_machete/tests/functional/test_update.py
```diff
@@ -0,0 +1,135 @@
...
+from git_machete import cli
+from git_machete.tests.functional.commons import (GitRepositorySandbox,
+                                                  get_current_commit_hash, git,
+                                                  mock_run_cmd, popen)
+
+
+class TestMachete:
+
+    @staticmethod
+    def launch_command(*args: str) -> str:
```
PawelLipski, line 15: Hmmm this should rather live in some common util class 🤔

git_machete/tests/functional/test_update.py
```diff
@@ -0,0 +1,135 @@
...
+        with io.StringIO() as out:
+            with redirect_stdout(out):
+                with redirect_stderr(out):
+                    cli.launch(list(args))
+                    git.flush_caches()
+            return out.getvalue()
+
+    def setup_method(self) -> None:
```
PawelLipski, line 23: And this as well

AABur:
Agreed, but I want to split up a huge test_machete first. And then I'll reanalyse it.  It will be easier to do this with small test modules.

PawelLipski:
Okay! merging then

Issue #508, opened 2022-04-29
Allow easy mouse selection of branch in Alacritty


kgadek:
# Overview

Currently, this is the output of status:

```console
$ git machete status
  main
  │
  └─issue/abc
  │ │
  │ └─issue/def
  │   │
  │   └─issue/ghi
  │   │
  │   └─issue/jkl
  │
  └─meta/vwx
```

In iTerm, I can double-click on `issue/abc` and it highlights exactly that. However, in Alacritty double-clicking selects `└─issue/abc`. _Not too convenient._

For comparison, `tree` has the following output:

```console
$ tree -L 2
.
├── CONTRIBUTING.md
├── Dockerfile
├── ISSUE_TEMPLATE.md
# …snip…
├── blogs
│   ├── docker-ci-tricks-1
│   ├── docker-ci-tricks-2
│   ├── git-machete-1
│   └── git-machete-2
├── ci
│   ├── checks
│   ├── deb-ppa-test-install
# …snip…
```

<details><summary>I believe here is the easiest possible fix. It aligns <code>git machete status</code> with what <code>tree</code> does by indenting 4 spaces, and adding one space before branch name.</summary>

```diff
diff --git a/git_machete/client.py b/git_machete/client.py
index 8e2b6cc..811c4fe 100644
--- a/git_machete/client.py
+++ b/git_machete/client.py
@@ -936,7 +936,7 @@
             out.write("  ")
             for p in accumulated_path[:-1]:
                 if not p:
-                    out.write("  ")
+                    out.write("    ")
                 else:
                     out.write(colored(f"{utils.get_vertical_bar()} ", edge_color[p]))
             out.write(colored(suffix, edge_color[branch_]))
@@ -972,7 +972,7 @@
                                   f"{dim(commit.short_hash)}  " if opt_list_commits_with_hashes else "", dim(commit.subject),
                                   fp_suffix))
                 elbow_ascii_only: Dict[str, str] = {AnsiEscapeCodes.DIM: "m-", AnsiEscapeCodes.RED: "x-", AnsiEscapeCodes.GREEN: "o-", AnsiEscapeCodes.YELLOW: "?-"}
-                elbow: str = u"└─" if not utils.ascii_only else elbow_ascii_only[edge_color[branch]]
+                elbow: str = u"└── " if not utils.ascii_only else elbow_ascii_only[edge_color[branch]]
                 print_line_prefix(branch, elbow)
             else:
                 if branch != dfs_res[0][0]:

[...]

PR #510, opened 2022-04-30
Apply minor refactors in `def status`

PawelLipski:
Spurred by #508... doesn't address any of its suggestion (no change in actual logic)

### DIFF
--- a/git_machete/client.py
+++ b/git_machete/client.py
@@ -872,6 +872,7 @@ class MacheteClient:
                 f"The initial branch {bold(initial_branch)} has been slid out. "
                 f"Returned to nearest remaining managed branch {bold(nearest_remaining_branch)}")
 
+    # TODO (#509): tidy up this method
     def status(
             self,
             *,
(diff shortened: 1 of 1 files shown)


git_machete/client.py
```diff
@@ -971,8 +972,23 @@ def print_line_prefix(branch_: LocalBranchShortName, suffix: str) -> None:
...
+                elbow: str
+                if utils.ascii_only:
+                    elbow_ascii_only: Dict[str, str] = {
+                        AnsiEscapeCodes.DIM: "m-",
+                        AnsiEscapeCodes.RED: "x-",
+                        AnsiEscapeCodes.GREEN: "o-",
+                        AnsiEscapeCodes.YELLOW: "?-"}
+                    elbow = elbow_ascii_only[edge_color[branch]]
+                else:
+                    # Note that using the "├─"-style elbow has been considered as well:
```
PawelLipski, line 985: @kgadek now I reminded myself why I didn't choose to use three-legged elbows (as `tree` does) 4 years ago when I first implemented that:

image

This boundary b/w green and red doesn't look good :/

PR #512, opened 2022-05-02
test `advance`

AABur:
RE: #232 

Tests for `advance`

Additional: implemented `pytest-mock` instead of `unittest.mock`

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -13,6 +13,7 @@ install_command =
 deps =
   pytest
   pytest-xdist
+  pytest-mock
 allowlist_externals = mkdir
 commands =
   mkdir -p test-results/
--- a/git_machete/tests/functional/test_advance.py
+++ b/git_machete/tests/functional/test_advance.py
@@ -1,0 +1,109 @@
+from typing import Any
+
+import pytest  # type: ignore
+from git_machete.git_operations import LocalBranchShortName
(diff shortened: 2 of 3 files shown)


tox.ini
```diff
@@ -13,6 +13,7 @@ install_command =
 deps =
   pytest
   pytest-xdist
+  pytest-mock
```
PawelLipski, line 16: If it comes to rationale for the change around mocks... this is for the sake of making the tests more consistent with pytest conventions right? Or does it have some other positive side effects?

PawelLipski:
Also! what else is to be done on this PR before it's no longer a draft?

tox.ini
AABur: In short, `pytest-mock` allows to write very concise tests and reduces a lot of boilerplate.
It also provides the very useful `mocker.spy()` and `mocker.stub()`  features, something doable with `unittest.mock` but requires lots of code. For more info read [pytest-mock documentation](https://pytest-mock.readthedocs.io/en/latest/)

AABur:
> Also! what else is to be done on this PR before it's no longer a draft?

I made two commits on purpose to highlight the changes made by using `pytest-mock`. You can see them in this commit `tests with pytest-mock`

PawelLipski (approved):

PR #513, opened 2022-05-02
test `go`

AABur:
RE: #232 

Tests for `go`


PawelLipski (approved):
Wow looks that almost everything is gone from `test_machete.py` by now 🎉

AABur:
> Wow looks that almost everything is gone from `test_machete.py` by now 🎉

Yes almost! A little more tuning of common functions and #232 can be closed.

PR #514, opened 2022-05-03
test `traverse` and `status`

AABur:
RE: #232 

Tests for `traverse` and `status`


AABur:
Splitting `test_machete.py` is finished with this commit. There will be one more commit to clean up common functions and mockers.

PawelLipski (approved):

PR #515, opened 2022-05-05
refactoring tests

### LINKED ISSUE #321: Decide on whether to keep `launch_command` or switch to `os.popen` to run git machete commands in tests
Currently (after #320 is merged) there are 2 approaches to running `git machete` commands in tests:
1. using `MacheteTester.launch_command()`
2. using `os.popen()` (as context manager). 

The latter one was introduced, because if some commands were opening `less` it was not possible to read the output using `launch_command` approach.

This issue goal is to unify the approach to running `git machete` commands in tests by either updating `launch_command` so it can read all the outputs or by using `os.popen` everywhere.

AABur:
FIX #232

I got carried away, and the PR is massive. But on the other hand, this PR covers a lot of refactoring of the tests.


PawelLipski:
Wooow that's massive indeed :) I'll go through that probably tomorrow... I don't expect I'll post many comments, however, as the changes are rather clearly beneficial

AABur:
CircleCi general checks failed due to a change of tests location.
`git ls-files '*.py' ':!**/tests/**' \`

PawelLipski:
One thing that I noticed from running `tox -e mypy` locally:

```
mypy-py310 run-test: commands[0] | mypy --config-file mypy.ini git_machete
Success: no issues found in 10 source files
```

Only `git_machete` package (i.e. production code) is checked but not tests... checking tests against mypy isn't of utmost importance, but still nice to have. Pls change the last line of tox.ini to:

```
  mypy --config-file mypy.ini git_machete tests
```

PawelLipski:
> CircleCi general checks failed due to a change of tests location. `git ls-files '*.py' ':!**/tests/**' \`

Oh, you mean the failure in `prohibit-strings-split-without-delimiter.sh`... actually, I think this might be relevant error, and tests should NOT be excluded indeed 🤔 for example:

```
tests/test_advance.py:81-82: it looks that the string is split without a delimeter, words will be glued up together when printed

            "Verify that when there is only one, rebased downstream branch of a" \
            "current branch 'git machete advance' merges commits from that branch" 
```

Isn't the error message right here? I think the error will have the word `amessage` glued up together right?

PawelLipski:
Btw @AABur do you have a LinkedIn account? I'd like to talk to you in a more private setting ;)

tests/test_github.py
```diff
@@ -51,36 +44,10 @@ def mock__get_github_token_fake() -> Optional[str]:
     return 'token'
 
 
-def mock_should_perform_interactive_slide_out(cmd: str) -> bool:
-    return True
-
-
 class TestMachete:
```
PawelLipski: `TestGitHub` to keep in line with the conventions?

PawelLipski: Also found that `TestMachete` is still present in `test_status.py`

AABur:
> Btw @AABur do you have a LinkedIn account? I'd like to talk to you in a more private setting ;)

Yes https://www.linkedin.com/in/alexander-burchenko-bb829415a/

tests/test_github.py
AABur: DONE!

AABur:
> > CircleCi general checks failed due to a change of tests location. `git ls-files '*.py' ':!**/tests/**' \`
> 
> Oh, you mean the failure in `prohibit-strings-split-without-delimiter.sh`... actually, I think this might be relevant error, and tests should NOT be excluded indeed 🤔 for example:

Honestly, I don't see why this has to be checked separately for python code when linters are available. I will make a special issue on this subject where we can discuss everything and set up a style guide.

PawelLipski:
> > > CircleCi general checks failed due to a change of tests location. `git ls-files '*.py' ':!**/tests/**' \`
> > 
> > 
> > Oh, you mean the failure in `prohibit-strings-split-without-delimiter.sh`... actually, I think this might be relevant error, and tests should NOT be excluded indeed thinking for example:
> 
> Honestly, I don't see why this has to be checked separately for python code when linters are available. I will make a special issue on this subject where we can discuss everything and set up a style guide.

Okay, a linter would be a much better solution then `awk` indeed, pls open an issue

tests/test_github.py
```diff
@@ -840,7 +807,7 @@ def test_github_checkout_prs(self) -> None:
     @mock.patch('git_machete.github.GITHUB_REMOTE_PATTERNS', FAKE_GITHUB_REMOTE_PATTERNS)
     @mock.patch('urllib.request.urlopen', MockContextManager)
     @mock.patch('urllib.request.Request', git_api_state_for_test_github_checkout_prs_fresh_repo.new_request())
-    def test_github_checkout_prs_freshly_cloned(self) -> None:
+    def test_github_checkout_prs_freshly_cloned(self, tmp_path: Any) -> None:
```
PawelLipski, line 810: Would it make sense to type these test case params more specifically as e.g. `str`?

tests/test_slide_out.py
```diff
@@ -1,51 +1,15 @@
...
+from typing import Any
+
+import pytest
+
+from .mockers import (GitRepositorySandbox, assert_command,
+                      get_current_commit_hash, launch_command, mock_run_cmd,
+                      mock_run_cmd_and_forward_stdout)
+
+
+class TestSlideout:
```
PawelLipski, line 10: Nitpick: `TestSlideOut` for consistency

tests/test_slide_out.py
```diff
@@ -212,8 +176,9 @@ def test_slide_out(self) -> None:
             """,
         )
 
-    @mock.patch('git_machete.utils.run_cmd', mock_run_cmd_and_forward_stdout)
-    def test_slide_out_with_valid_down_fork_point(self) -> None:
+    def test_slide_out_with_valid_down_fork_point(self, mocker: Any) -> None:
```
PawelLipski, line 179: Same for `mocker`... does it make sense to type it more specifically?

PawelLipski (approved):
Minor comments, otherwise almost ready to merge

PawelLipski:
> > > > CircleCi general checks failed due to a change of tests location. `git ls-files '*.py' ':!**/tests/**' \`
> > > 
> > > 
> > > Oh, you mean the failure in `prohibit-strings-split-without-delimiter.sh`... actually, I think this might be relevant error, and tests should NOT be excluded indeed thinking for example:
> > 
> > 
> > Honestly, I don't see why this has to be checked separately for python code when linters are available. I will make a special issue on this subject where we can discuss everything and set up a style guide.
> 
> Okay, a linter would be a much better solution then `awk` indeed, pls open an issue

For the time being... pls either fix the strings in test cases, or just exclude tests from the `git ls-files`

PawelLipski:
Okay @AABur... any changes planned here on your part or can we merge?

AABur:
> For the time being... pls either fix the strings in test cases, or just exclude tests from the `git ls-files`

DONE!
I rewrote the `strings` in a more Pythonic style

PR #525, opened 2022-05-17
Issue #497: Parse branch name input from cli

### LINKED ISSUE #497: Problems around parsing branch names
```
$ git machete diff develop
...
AttributeError: 'str' object has no attribute 'full_name'

$ git machete log develop
(ok)

$ git machete diff refs/heads/develop
...
TypeError: LocalBranchShortName cannot accept refs/heads or refs/remotes. Given value: refs/heads/develop.
Consider posting an issue on https://github.com/VirtusLab/git-machete/issues/new

$ git machete log refs/heads/develop
...
TypeError: LocalBranchShortName cannot accept refs/heads or refs/remotes. Given value: refs/heads/develop.
Consider posting an issue on https://github.com/VirtusLab/git-machete/issues/new
```

Hell knows what other cases result in an error... let's get this sorted and provide decent regression tests.

amalota:
closes #497


amalota:
@PawelLipski Can we safely assume that the user should provide branch short name, IMO that would be the most intuitive input type, wdyt?

tests/test_diff.py
```diff
@@ -0,0 +1,43 @@
...
+            .new_repo(self.repo_sandbox.remote_path, "--bare")
+            .new_repo(self.repo_sandbox.local_path)
+            .execute(f"git remote add origin {self.repo_sandbox.remote_path}")
+            .execute('git config user.email "tester@test.com"')
+            .execute('git config user.name "Tester Test"')
+        )
+
+    def test_diff(self, mocker: Any) -> None:
+        """
+        Verify behaviour of a 'git machete add' command.
```
PawelLipski:
```suggestion
        Verify behaviour of a 'git machete diff' command.
```

tests/test_diff.py
```diff
@@ -0,0 +1,43 @@
...
+        mocker.patch('git_machete.utils.run_cmd', mock_run_cmd)  # to hide git outputs in tests
+
+        (
+            self.repo_sandbox.new_branch("master")
+                .commit()
+                .new_branch("develop")
+                .commit()
+        )
+        launch_command("discover", "-y")
+        launch_command("diff", "develop")
```
PawelLipski: Pls also include a test for `diff` without a branch name

PawelLipski:
> @PawelLipski Can we safely assume that the user should provide branch short name, IMO that would be the most intuitive input type, wdyt?

Yes... but still, we should be prepared for the case of someone passing `refs/heads/develop`... we can crash git-machete but in a controlled way (now some random error is happening)

PawelLipski:
> > @PawelLipski Can we safely assume that the user should provide branch short name, IMO that would be the most intuitive input type, wdyt?
> 
> Yes... but still, we should be prepared for the case of someone passing `refs/heads/develop`... we can crash git-machete but in a controlled way (now some random error is happening)

Although ideally the case of `git machete diff refs/heads/develop` should behave exactly as `git machete diff develop`

amalota:
> > > @PawelLipski Can we safely assume that the user should provide branch short name, IMO that would be the most intuitive input type, wdyt?
> > 
> > 
> > Yes... but still, we should be prepared for the case of someone passing `refs/heads/develop`... we can crash git-machete but in a controlled way (now some random error is happening)
> 
> Although ideally the case of `git machete diff refs/heads/develop` should behave exactly as `git machete diff develop`

@PawelLipski I can add a simple parser which would inference the type between `LocalBranchShortName` and `LocalBranchFullName`, so both would work, wdyt?

PawelLipski:
> @PawelLipski I can add a simple parser which would inference the type between `LocalBranchShortName` and `LocalBranchFullName`, so both would work, wdyt?

Okay... let's give it a shot

tox.ini
```diff
@@ -7,7 +7,7 @@ skipsdist = True
 setenv =
   VIRTUAL_ENV={envdir}
   PYTHONWARNINGS=default::DeprecationWarning,ignore:The distutils package is deprecated,ignore:The distutils.sysconfig module is deprecated,ignore:the imp module is deprecated,ignore:SelectableGroups dict interface is deprecated
-usedevelop = False
+usedevelop = True
```
PawelLipski: It should not matter now that `mock_run_cmd_and_forward_stdout` is applied right?

RELEASE_NOTES.md
```diff
@@ -1,5 +1,7 @@
 # Release notes
 
+## New in git-machete 3.10.1
+
```
PawelLipski, line 4: To be filled up (as there are user-visible fixes on this PR)

git_machete/cli.py
```diff
@@ -475,9 +475,9 @@ def set_utils_global_variables(
     utils.verbose_mode = cli_opts.opt_verbose
 
 
-def get_branch_arg_or_current_branch(
-        cli_opts: git_machete.options.CommandLineOptions, git_context: GitContext) -> LocalBranchShortName:
-    return cli_opts.opt_branch or git_context.get_current_branch()
+def get_local_branch_short_name_from_arg_or_current_branch(
+        branch_from_arg: Optional[AnyBranchName], git_context: GitContext) -> LocalBranchShortName:
+    return LocalBranchShortName.of(branch_from_arg.replace('refs/heads/', '')) if branch_from_arg else None or git_context.get_current_branch()
```
PawelLipski: 🤔 
Maybe just:
```suggestion
    return LocalBranchShortName.of(branch_from_arg.replace('refs/heads/', '')) if branch_from_arg else git_context.get_current_branch()
```
❓

git_machete/cli.py
```diff
@@ -540,7 +540,7 @@ def launch(orig_args: List[str]) -> None:
         should_perform_interactive_slide_out = MacheteClient.should_perform_interactive_slide_out(cmd)
         if cmd == "add":
             machete_client.read_definition_file(perform_interactive_slide_out=should_perform_interactive_slide_out)
-            branch = get_branch_arg_or_current_branch(cli_opts, git)
```
PawelLipski, line 543: Btw, was any other command then `diff` failing on `git machete <command> <short-branch-name>`? 😅

tests/mockers.py
```diff
@@ -78,6 +78,12 @@ def commit(self, message: str = "Some commit message.") -> "GitRepositorySandbox
         self.execute(f'git commit -m "{message}"')
         return self
 
+    def add_file_with_content_and_commit(self, file_name: str = 'file_name.txt', file_content: str = 'Some file content', message: str = "Some commit message.") -> "GitRepositorySandbox":
+        self.execute(f"echo {file_content} > {file_name}")
```
PawelLipski: I'd wrap file content in apostrophes to avoid surprises
```suggestion
        self.execute(f"echo '{file_content}' > {file_name}")
```
Or best, just use some Python API for writing the file, rather than shell

tests/test_add.py
```diff
@@ -40,13 +39,32 @@ def test_add(self, mocker: Any) -> None:
 
         self.repo_sandbox.new_branch("bugfix/feature_fail")
 
+        # Test `git machete diff` without providing the branch name
```
PawelLipski:
```suggestion
        # Test `git machete add` without providing the branch name
```

tests/test_add.py
```diff
@@ -40,13 +39,32 @@ def test_add(self, mocker: Any) -> None:
...
+            ['add', '-y', 'bugfix/some_feature'],
+            'Adding `bugfix/some_feature` onto the inferred upstream (parent) branch `develop`\n'
+            'Added branch `bugfix/some_feature` onto `develop`\n',
+            strip_indentation=False
+        )
+
+        self.repo_sandbox.check_out('develop')
+        self.repo_sandbox.new_branch("bugfix/another_feature")
+        assert_command(
+            ['add', '-y', 'refs/heads/bugfix/another_feature'],
```
PawelLipski, line 62: Gr8 🎉 curious how it affected the overall coverage... https://app.codecov.io/gh/VirtusLab/git-machete/branch/bug%2F497-parsing-branch-name doesn't seem to display the most recent data here (probably since the tests haven't been passing recently; failed tests don't count for coverage)

PawelLipski: (btw do you have access to Codecov?)

tests/test_diff.py
```diff
@@ -0,0 +1,53 @@
+from typing import Any
+
+from .mockers import (GitRepositorySandbox, assert_command, mock_run_cmd_and_forward_stdout)
+
+
+class TestDiff:
```
PawelLipski, line 6: :benissimo:

tests/test_fork_point.py
```diff
@@ -0,0 +1,45 @@
...
+            .execute('git config user.email "tester@test.com"')
+            .execute('git config user.name "Tester Test"')
+        )
+
+    def test_fork_point(self) -> None:
+        """
+        Verify behaviour of a 'git machete fork-point' command.
+        """
+        fixed_committer_and_author_date = 'Mon 20 Aug 2018 20:19:19 +0200'
+        os.environ['GIT_COMMITTER_DATE'] = fixed_committer_and_author_date
```
PawelLipski, line 27: Hmm for the sake of simplicity, how about the following:

1. Add `os.environ['GIT_AUTHOR_DATE'] = fixed_committer_and_author_date`
2. Remove both `.execute(f'git commit --amend...')` calls

My hypothesis is that you'll still get reproducible results (AFAICS you'll even get the exact same commit hashes as now)

tests/test_fork_point.py
```diff
@@ -0,0 +1,45 @@
...
+                .new_branch("develop")
+                .add_file_with_content_and_commit(file_content='develop content', message="develop commit.")
+                .execute(f'git commit --amend --no-edit --date "{fixed_committer_and_author_date}"')  # change commit author date
+                .new_branch("feature")
+                .commit('feature commit.')
+        )
+        launch_command('discover', '-y')
+
+        # Test `git machete fork-point` without providing the branch name
+        assert_command(["fork-point"], "2df2fb0e464ce06a8cc1020ed95a732df8362798\n", strip_indentation=False)
```
PawelLipski: For clarity, pls add a comment about **which commit these hashes correspond to** (i.e. a commit pointed **by which branch**)

tests/test_log.py
```diff
@@ -34,17 +33,19 @@ def test_log(self, mocker: Any) -> None:
         self.repo_sandbox.commit()
         childs_second_commit_hash = get_current_commit_hash()
 
-        log_content = launch_command('log')
+        log_content_no_branch_name = launch_command('log')
+        log_content_short_branch_name = launch_command('log', 'child')
+        log_content_full_branch_name = launch_command('log', 'refs/heads/child')
 
-        assert childs_first_commit_hash in log_content, \
+        assert childs_first_commit_hash in log_content_short_branch_name and log_content_no_branch_name and log_content_full_branch_name, \
```
PawelLipski: Huh how is it executed? 🤔 does `X in A and B and C` really check `X`'s presence in all of `A`, `B`,`C`? 🤔 Isn't it interpreted just as `(X in A) and B and C`, which would be somewhat useless here?

tests/test_anno.py
```diff
@@ -0,0 +1,72 @@
...
+            """
+            master (untracked)
+
+            develop *  Custom annotation for `develop` branch (untracked)
+            |
+            x-feature  Custom annotation for `feature` branch (untracked)
+            """,
+        )
+
+        launch_command('anno', '-b=refs/heads/feature', 'Custom annotation for `feature` branch')
```
PawelLipski, line 62: Also! for clarity, what happens if `refs/remote/origin/...` or `refs/tags/...` for an actually existing remote branch/tag is passed? And what happens if a string NOT corresponding to any branch in any way is passed?

I'm mostly concerned whether the error messages are clean (and not some obscure stack traces)

git_machete/cli.py
amalota: @PawelLipski  AFAIK only `diff` was failing on `short_name`, but a lot more were failing on `git machete <command> <full-branch-name>`

PawelLipski: Phew 😌

tests/mockers.py
amalota: Ok, i wanted to keep the style of only using shell in the function but i guess it's not necessary

tests/test_add.py
amalota: @PawelLipski yes, i do have access to Codecov but as you pointed out it doesn't display the most recent data, will come back to it once the tests pass

tests/test_fork_point.py
amalota: Yes, it works, i actually tested it firstly and it didn't work but the fault was on the difference in tree hash, forgot to come back to it

tests/mockers.py
PawelLipski: Yeah, shell invocations are still much more brittle than pure Python

tests/test_log.py
amalota: Yes great catch, it indeed is interpreted as `(X in A) and B and C` i thought it would get interpreted in a similar way as i does for comparison operators(e.g. `a<b<c` gets interpreted as `a<b and b<c`), mb

RELEASE_NOTES.md
```diff
@@ -2,6 +2,8 @@
 
 ## New in git-machete 3.10.1
 
+- added: `git machete {diff, log, add, anno, fork-point, is-managed, show}` commands now support both short (`develop`) and full (`refs/heads/develop`) branch names
```
PawelLipski: Huh right! that's indeed a MINOR-level change in that sense, lol &mdash; a user-facing feature

PawelLipski: Huh right! that's indeed a MINOR-level change in that sense, lol &mdash; a user-facing feature, so it'd be 3.11.0 ...

PawelLipski: Although I rather wouldn't consider an actual **added feature**

PawelLipski: I'd rather leave 3.10.1 (a PATCH-level update i.e. a bugfix)... but change the RELEASE_NOTES entry to
```
- fixed: git machete diff doesn't crash when ...
- fixed: git machete {log, add, anno, fork-point, is-managed, show} doesn't crash when...
```

tests/mockers.py
```diff
@@ -117,6 +117,10 @@ def add_git_config_key(self, key: str, value: str) -> "GitRepositorySandbox":
         self.execute(f'git config {key} {value}')
         return self
 
+    def write_to_file(self, file_name: str, file_content: str):
```
PawelLipski: Nit: IIRC pytest has even that piece ready... I remember that Alex Burchenko used it somewhere (although maybe a false memory)

tests/test_add.py
```diff
@@ -23,7 +23,7 @@ def test_add(self, mocker: Any) -> None:
         """
         Verify behaviour of a 'git machete add' command.
         """
-        mocker.patch('git_machete.utils.run_cmd', mock_run_cmd)  # to hide git outputs in tests
+        # mocker.patch('git_machete.utils.run_cmd', mock_run_cmd)  # to hide git outputs in tests
```
PawelLipski: To be removed for good?

tests/test_anno.py
amalota: @PawelLipski Same error for each case:

**refs/remote/origin/...:**
`➜ git machete anno -b=refs/remote/origin/develop 'test'
Branch refs/remote/origin/develop not found in the tree of branch dependencies.
Use git machete add refs/remote/origin/develop or git machete edit
`

**refs/tags/...:**
`➜ git machete anno -b=refs/tags/v3.9.1 'test'
Branch refs/tags/v3.9.1 not found in the tree of branch dependencies.
Use git machete add refs/tags/v3.9.1 or git machete edit
`

**non-existing branch:**
`➜ git machete anno -b=random_branch 'test'   
Branch random_branch not found in the tree of branch dependencies.
Use git machete add random_branch or git machete edit
`

PawelLipski: Gr8 ❤️

tests/mockers.py
amalota: I can't find it :(

tests/test_fork_point.py
```diff
@@ -34,6 +34,10 @@ def test_fork_point(self) -> None:
                 .new_branch("feature")
                 .commit('feature commit.')
         )
+        # Clean-up environment variables
+        os.environ.pop('GIT_COMMITTER_DATE', None)
```
PawelLipski: Nit: in the unlikely circumstance of the above statement failing, these 2 vars will NOT be popped... and thus will likely be propagated to other test cases. If so, then pls use `try..finally` here

PawelLipski (approved):

tests/test_fork_point.py
amalota: @PawelLipski  Hmm, i especially chose `os.environ.pop('GIT_COMMITTER_DATE', None)` so in the unlikely scenario that these variables don't get set in the previous steps, the test won't fail. I don't see any other reason for these statements to fail, wdyt?

amalota: @PawelLipski  Hmm, i especially chose `os.environ.pop('GIT_COMMITTER_DATE', None)` so in the unlikely scenario that these variables don't get set in the previous steps, the test won't fail. I don't see any other reason for these statements to fail, wdyt?

PawelLipski: No! I mean, what if `self ... commit(...)` fails? 🤔 `GIT_COMMITTER_DATE` will remain set!

PawelLipski (approved):

PawelLipski (approved):

PR #532, opened 2022-05-26
Issue #528: Add docstrings in `mockers.py` to explain the most complicated mocks

### LINKED ISSUE #528: Add docstrings in the `mockers.py` file
- add docstrings inside the mocking functions (at least `mock_run_cmd_and_forward_stdout`) explaining how they work
- add main doctring at the top of the file explaining when/why use each mocked function (example: why is it necessary to use `mock_run_cmd_and_forward_stdout` for the `log` and `diff` commands)

amalota:
closes #528


PawelLipski (approved):
:benissimo:

PR #536, opened 2022-05-27
Issue #535: Update completion's README.md for bash shell for macos

### LINKED ISSUE #535: Update completion's `README.md` for bash shell for macos - doesn't work ATM

amalota:
closes #535

### DIFF
--- a/completion/README.md
+++ b/completion/README.md
@@ -17,7 +17,7 @@ If the bash completion doesn't work:
         && . $(brew --prefix)/etc/bash_completion.d/git-completion.bash
       [ -f "$(brew --prefix)/etc/bash_completion.d/git-machete.completion.bash" ] 
         && . $(brew --prefix)/etc/bash_completion.d/git-machete.completion.bash
-  }
+    }
 ```
 
 
--- a/completion/README.md
+++ b/completion/README.md
@@ -10,13 +10,14 @@ If the bash completion doesn't work:
 1. Install git bash completions with `brew install git bash-completion`
 2. Add the below code to `~/.bash_profile` and reload the shell
 ```shell script
-[ -f /usr/local/etc/bash_completion ] && . /usr/local/etc/bash_completion || {
-    # if not found in /usr/local/etc, try the brew --prefix location
-    [ -f "$(brew --prefix)/etc/bash_completion.d/git-completion.bash" ] && \
-    . $(brew --prefix)/etc/bash_completion.d/git-completion.bash
-    [ -f "$(brew --prefix)/etc/bash_completion.d/git-machete.completion.bash" ] && \
-    . $(brew --prefix)/etc/bash_completion.d/git-machete.completion.bash
-}
+[ -f /usr/local/etc/bash_completion ] && . /usr/local/etc/bash_completion 
+  || # if not found in /usr/local/etc, try the brew --prefix location
+    {
+      [ -f "$(brew --prefix)/etc/bash_completion.d/git-completion.bash" ] 
+        && . $(brew --prefix)/etc/bash_completion.d/git-completion.bash
+      [ -f "$(brew --prefix)/etc/bash_completion.d/git-machete.completion.bash" ] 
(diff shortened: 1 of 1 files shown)


amalota:
Not sure if these changes or sth else while i was trying to get bash completion to work, but zsh completion stopped working, this PR is blocked until i fix it/figure out what's caused it.

amalota:
**Update**
Turns out that running `brew install git bash-completion` is vital for git-machete bash completion to work but at the same time brakes git-machete zsh completion, @PawelLipski wdyt? Is that to be expected or should i keep looking for a better solution?

PawelLipski:
> Turns out that running brew install git bash-completion is vital for git-machete bash completion to work but at the same time brakes git-machete zsh completion

Wow that's interesting lol 😯 

Okay so from practical point of view, I'd note that:
1. every Mac user I know uses (oh my) zsh,
2. literally every person I know, Mac or otherwise, uses just a single shell (bash/zsh/fish), never two at the same time

So I think we can safely ignore that problem!

PawelLipski (commented):
Generally approving then!

completion/README.md
```diff
@@ -6,6 +6,19 @@ Make sure you have bash completion installed (with `brew install bash-completion
 
 `brew install git-machete` automatically installs bash completion files for `git machete`.
 
+If the bash completion doesn't work:
+1. Install git bash completions with `brew install git bash-completion`
+2. Add the below code to `~/.bash_profile` and reload the shell
+```shell script
+[ -f /usr/local/etc/bash_completion ] && . /usr/local/etc/bash_completion || {
```
PawelLipski: Actually... where is it taken from? or is it a work of yours?

completion/README.md
```diff
@@ -6,6 +6,19 @@ Make sure you have bash completion installed (with `brew install bash-completion
...
+    # if not found in /usr/local/etc, try the brew --prefix location
+    [ -f "$(brew --prefix)/etc/bash_completion.d/git-completion.bash" ] && \
+    . $(brew --prefix)/etc/bash_completion.d/git-completion.bash
```
PawelLipski: Nit: indent the continuation lines like this one by one level, so that it's clear it's a continuation

PawelLipski: Or even better, extra indent + move `&&` from the previous line to the start of this line

completion/README.md
amalota: It's taken from here: https://gist.github.com/romansavrulin/41e55fba693b4025ed693559083bc3a0 then i deduced i need to to the same for git-machete bash completions to work

completion/README.md
amalota: @PawelLipski better now?

image

PawelLipski: Huh but this isn't a valid shell script now! Try executing that

PawelLipski: Huh but this isn't a valid shell script now! Try executing that

PawelLipski: Trailing `\`s are still needed

PawelLipski (changes requested):
as in comments

amalota:
> as in comments

@PawelLipski fixed

completion/README.md
```diff
@@ -6,6 +6,20 @@ Make sure you have bash completion installed (with `brew install bash-completion
 
 `brew install git-machete` automatically installs bash completion files for `git machete`.
 
+If the bash completion doesn't work:
+1. Install git bash completions with `brew install git bash-completion`
+2. Add the below code to `~/.bash_profile` and reload the shell
+```shell script
+[ -f /usr/local/etc/bash_completion ] && . /usr/local/etc/bash_completion \
+  || # if not found in /usr/local/etc, try the brew --prefix location
```
PawelLipski, line 14: Does it execute cleanly now? 🤔

amalota: Yeah, i tested it and it works OK

PawelLipski: Okay! Indeed; I thought that `\` would be necessary after `||` as well... but it isn't

PawelLipski (approved):

PR #537, opened 2022-05-27
Issue #526: `--by=<TAB>` prompts file names in Bash, should prompt nothing

### LINKED ISSUE #526: `--by=<TAB>` prompts file names in Bash, should prompt nothing
image

Let's also check if other shells have the same bug

amalota:
closes #526


completion/git-machete.completion.bash
```diff
@@ -1,5 +1,10 @@
 #!/usr/bin/env bash
 
+__gitcomp_empty ()
+{
```
PawelLipski: Nit, for consistency:

```suggestion
__gitcomp_empty() {
```

completion/git-machete.completion.bash
```diff
@@ -32,7 +37,7 @@ _git_machete() {
 
   case $cur in
     --branch=*|--onto=*) __gitcomp_nl "$(git machete list managed 2>/dev/null)" "" "${cur##--*=}" ;;
-    --checked-out-since=*) __gitcomp "" ;;
+    --checked-out-since=*|--by=*) __gitcomp_empty "" "${cur##--*=}";;
```
PawelLipski: Nit:
```suggestion
    --by=*|--checked-out-since=*) __gitcomp_empty "" "${cur##--*=}";;
```
plus move up for abc order

PawelLipski: Also... why are there params passed to `__gitcomp_empty`? 🤔 they're not picked up in the function anyway

completion/git-machete.completion.bash
```diff
@@ -32,7 +36,7 @@ _git_machete() {
 
   case $cur in
     --branch=*|--onto=*) __gitcomp_nl "$(git machete list managed 2>/dev/null)" "" "${cur##--*=}" ;;
-    --checked-out-since=*) __gitcomp "" ;;
+    --by=*|--checked-out-since=*) __gitcomp_empty;;
```
PawelLipski: Okay... are all options covered now? IIUC effectively each option should be mentioned here in this `case`

amalota: AFAICS yes

PawelLipski: Shouldn't `case $prev in` below be updated as well tho?

completion/git-machete.completion.bash
```diff
@@ -1,5 +1,9 @@
 #!/usr/bin/env bash
 
+__gitcomp_empty() {
+  COMPREPLY=('')
```
PawelLipski: Note that `COMPREPLY=('')` is used in a few other places as well... you can replace that calls with `__gitcomp_empty` for consistency, but TBH I'd just remove this function and simply use `COMPREPLY=('')` in line 39

PawelLipski (approved):

PR #538, opened 2022-05-30
Issue #527: Hide GitHub token value in `--debug` mode

### LINKED ISSUE #527: GitHub token should NOT be printed out in `--debug` logs
```
$ git machete github anno-prs --debug
...
__fire_github_api_request(method=GET, path=/user, token=ghp_<<<REDACTED>>>, request_body=None): firing a GET request to https://api.github.com/user with a bearer token and request body <none>
sync_annotations_to_github_prs(self): Current GitHub user is PawelLipski
...
```

except that the token is not REDACTED, it's printed out verbatim 😅

The logs might be attached by users when providing insight into bugs... I'd prevent cases like that in some/all of the following ways:
* specifically exclude `token` param of `__fire_github_api_request` (and probably a few other methods) from being printed
* exclude all params named `token` anywhere
* exclude all params that look like GitHub token (`ghp_` prefix, maybe some other prefixes as well - TBD from GitHub docs)

amalota:
closes #527

### DIFF
--- a/git_machete/utils.py
+++ b/git_machete/utils.py
@@ -95,18 +95,15 @@ def debug(msg: Optional[str] = None) -> None:
         function_name = bold(inspect.stack()[1].function)
         args, _, _, values = inspect.getargvalues(inspect.stack()[1].frame)
 
-        args_to_be_redacted_by_name = {'access_token', 'password', 'secret', 'token'}
-        for arg in args_to_be_redacted_by_name:
-            if arg in values.keys():
-                values[arg] = '***'
-
-        args_to_be_redacted_by_value = ['ghp_', 'gho_', 'ghu_', 'ghs_', 'ghr_']  # https://github.blog/2021-04-05-behind-githubs-new-authentication-token-formats/
+        args_to_be_redacted = {'access_token', 'password', 'secret', 'token'}
+        # https://github.blog/2021-04-05-behind-githubs-new-authentication-token-formats/
+        values_to_be_redacted = ['ghp_', 'gho_', 'ghu_', 'ghs_', 'ghr_']
         for arg, value in values.items():
-            if any(arg_ in str(value) for arg_ in args_to_be_redacted_by_value):
+            if arg in args_to_be_redacted or any(value_ in str(value) for value_ in values_to_be_redacted):
                 values[arg] = '***'
 
         excluded_args = {'self'}
-        allowed_args = [arg for arg in args if arg not in excluded_args]
+        allowed_args = excluding(args, excluded_args)
         args_and_values_list = [arg + '=' + str(values[arg]) for arg in allowed_args]
         args_and_values_str = ', '.join(args_and_values_list)
         args_and_values_bold_str = bold(f'({args_and_values_str})')
--- a/git_machete/utils.py
+++ b/git_machete/utils.py
@@ -107,14 +107,14 @@ def debug(msg: Optional[str] = None) -> None:
 
         excluded_args = {'self'}
         allowed_args = [arg for arg in args if arg not in excluded_args]
-        args_and_values = [arg + '=' + str(values[arg]) for arg in allowed_args]
-        args_and_values = ', '.join(args_and_values)
-        args_and_values = bold(f'({args_and_values})')
+        args_and_values_list = [arg + '=' + str(values[arg]) for arg in allowed_args]
+        args_and_values_str = ', '.join(args_and_values_list)
+        args_and_values_bold_str = bold(f'({args_and_values_str})')
 
         if msg is None:
-            print(f"{function_name}{args_and_values}", file=sys.stderr)
+            print(f"{function_name}{args_and_values_bold_str}", file=sys.stderr)
         else:
-            print(f"{function_name}{args_and_values}: {dim(msg)}", file=sys.stderr)
+            print(f"{function_name}{args_and_values_bold_str}: {dim(msg)}", file=sys.stderr)
 
 
 def run_cmd(cmd: str, *args: str, **kwargs: Any) -> int:
--- a/git_machete/utils.py
+++ b/git_machete/utils.py
@@ -94,7 +94,21 @@ def debug(msg: Optional[str] = None) -> None:
     if debug_mode:
         function_name = bold(inspect.stack()[1].function)
         args, _, _, values = inspect.getargvalues(inspect.stack()[1].frame)
-        function_args = bold(f"({', '.join([arg+'='+str(values[arg]) if arg != 'self' else 'self' for arg in args])})")
+
+        redact_arg_by_name = ['token']
+        for arg in redact_arg_by_name:
+            values[arg] = '***'
+
+        redact_arg_by_value = ['ghp_']
+        for arg, value in values.items():
+            if value and isinstance(value, str):
+                if any(value.startswith(arg_) for arg_ in redact_arg_by_value):
+                    values[arg] = '***'
+
+        excluded_args = {'self'}
+        allowed_args = [arg for arg in args if arg not in excluded_args]
+        function_args = bold(f"({', '.join([arg+'='+str(values[arg]) for arg in allowed_args])})")
+
         if msg is None:
             print(f"{function_name}{function_args}", file=sys.stderr)
         else:
--- a/git_machete/utils.py
+++ b/git_machete/utils.py
@@ -95,7 +95,7 @@ def debug(msg: Optional[str] = None) -> None:
         function_name = bold(inspect.stack()[1].function)
         args, _, _, values = inspect.getargvalues(inspect.stack()[1].frame)
 
-        redact_arg_by_name = ['token']
+        redact_arg_by_name = ['access_token', 'password', 'secret', 'token']
         for arg in redact_arg_by_name:
             values[arg] = '***'
 
--- a/git_machete/utils.py
+++ b/git_machete/utils.py
@@ -95,15 +95,14 @@ def debug(msg: Optional[str] = None) -> None:
         function_name = bold(inspect.stack()[1].function)
         args, _, _, values = inspect.getargvalues(inspect.stack()[1].frame)
 
-        redact_arg_by_name = ['access_token', 'password', 'secret', 'token']
+        redact_arg_by_name = {'access_token', 'password', 'secret', 'token'}
         for arg in redact_arg_by_name:
             values[arg] = '***'
 
-        redact_arg_by_value = ['ghp_']
+        redact_arg_by_value = ['ghp_', 'gho_', 'ghu_', 'ghs_', 'ghr_']
         for arg, value in values.items():
-            if value and isinstance(value, str):
-                if any(value.startswith(arg_) for arg_ in redact_arg_by_value):
-                    values[arg] = '***'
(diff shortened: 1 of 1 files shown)


git_machete/utils.py
```diff
@@ -94,7 +94,21 @@ def debug(msg: Optional[str] = None) -> None:
     if debug_mode:
         function_name = bold(inspect.stack()[1].function)
         args, _, _, values = inspect.getargvalues(inspect.stack()[1].frame)
-        function_args = bold(f"({', '.join([arg+'='+str(values[arg]) if arg != 'self' else 'self' for arg in args])})")
+
+        redact_arg_by_name = ['token']
```
PawelLipski: While you're here, I'd also add:
```suggestion
        redact_arg_by_name = ['access_token', 'password', 'secret', 'token']
```
to make it future-proof

git_machete/utils.py
```diff
@@ -94,7 +94,21 @@ def debug(msg: Optional[str] = None) -> None:
...
+        for arg in redact_arg_by_name:
+            values[arg] = '***'
+
+        redact_arg_by_value = ['ghp_']
```
PawelLipski: That's the only prefix used by GitHub tokens right?

git_machete/utils.py
```diff
@@ -94,7 +94,21 @@ def debug(msg: Optional[str] = None) -> None:
...
+        for arg, value in values.items():
+            if value and isinstance(value, str):
+                if any(value.startswith(arg_) for arg_ in redact_arg_by_value):
+                    values[arg] = '***'
+
+        excluded_args = {'self'}
+        allowed_args = [arg for arg in args if arg not in excluded_args]
```
PawelLipski: I'd say just `included_args`, as `self` isn't somehow _disallowed_, it's just _excluded_

PawelLipski: Although TBH... `self` is sometimes useful to include as well, did you exclude it due to security reasons? 🤔

git_machete/utils.py
PawelLipski: Actually, it can be a set, same for `redact_arg_by_value` below

git_machete/utils.py
PawelLipski:
```suggestion
        redacted_arg_by_prefix = {'ghp_'}
```

git_machete/utils.py
```diff
@@ -94,7 +94,21 @@ def debug(msg: Optional[str] = None) -> None:
...
-        function_args = bold(f"({', '.join([arg+'='+str(values[arg]) if arg != 'self' else 'self' for arg in args])})")
+
+        redact_arg_by_name = ['token']
+        for arg in redact_arg_by_name:
+            values[arg] = '***'
+
+        redact_arg_by_value = ['ghp_']
+        for arg, value in values.items():
+            if value and isinstance(value, str):
+                if any(value.startswith(arg_) for arg_ in redact_arg_by_value):
```
PawelLipski: I'd even make it more strict, sth like
```suggestion
            if any(arg_ in str(value) for arg_ in redact_arg_by_value):
```

git_machete/utils.py
```diff
@@ -94,7 +94,21 @@ def debug(msg: Optional[str] = None) -> None:
     if debug_mode:
         function_name = bold(inspect.stack()[1].function)
         args, _, _, values = inspect.getargvalues(inspect.stack()[1].frame)
-        function_args = bold(f"({', '.join([arg+'='+str(values[arg]) if arg != 'self' else 'self' for arg in args])})")
+
+        redact_arg_by_name = ['token']
+        for arg in redact_arg_by_name:
```
PawelLipski: Also, `s/redact/redacted`

git_machete/utils.py
amalota: No, there are more
image
but i am not sure they will be used, but just to sleep well at night i can add them, @PawelLipski wdyt?

git_machete/utils.py
amalota: No, i just did to for visual reasons but also `self` doesn't provide that much information, @PawelLipski do you want me to allow it?

git_machete/utils.py
PawelLipski: Yeah, I suppose only personal access tokens can reasonably be used by git-machete... but let's include all 5 prefixes for the sake of forgetting about this problem once and for all

git_machete/utils.py
PawelLipski: I suppose that in >90% of logs `self` with be an instance of `MacheteClient` or `GitContext` which indeed brings little value... amirite? if so, then exclude `self` indeed

git_machete/utils.py
amalota: @PawelLipski why is this wrong? `redact argument by name` sounds correct to me

PawelLipski: Huh, I don't like **verbs** for variables and classes. Verbs are for method/function names, nouns for var/class names

git_machete/utils.py
```diff
@@ -95,15 +95,14 @@ def debug(msg: Optional[str] = None) -> None:
...
             values[arg] = '***'
 
-        redact_arg_by_value = ['ghp_']
+        redact_arg_by_value = ['ghp_', 'gho_', 'ghu_', 'ghs_', 'ghr_']
         for arg, value in values.items():
-            if value and isinstance(value, str):
-                if any(value.startswith(arg_) for arg_ in redact_arg_by_value):
-                    values[arg] = '***'
+            if any(arg_ in str(value) for arg_ in redact_arg_by_value):
+                values[arg] = '***'
```
PawelLipski, line 103: Can you show how a sensitive piece of logs looks like now?

git_machete/utils.py
```diff
@@ -95,15 +95,14 @@ def debug(msg: Optional[str] = None) -> None:
         function_name = bold(inspect.stack()[1].function)
         args, _, _, values = inspect.getargvalues(inspect.stack()[1].frame)
 
-        redact_arg_by_name = ['token']
+        redact_arg_by_name = {'access_token', 'password', 'secret', 'token'}
         for arg in redact_arg_by_name:
             values[arg] = '***'
 
-        redact_arg_by_value = ['ghp_']
+        redact_arg_by_value = ['ghp_', 'gho_', 'ghu_', 'ghs_', 'ghr_']
```
PawelLipski: Pls add a link to relevant GitHub docs page here in the comment

git_machete/utils.py
```diff
@@ -94,7 +94,20 @@ def debug(msg: Optional[str] = None) -> None:
     if debug_mode:
         function_name = bold(inspect.stack()[1].function)
         args, _, _, values = inspect.getargvalues(inspect.stack()[1].frame)
-        function_args = bold(f"({', '.join([arg+'='+str(values[arg]) if arg != 'self' else 'self' for arg in args])})")
+
+        redact_arg_by_name = {'access_token', 'password', 'secret', 'token'}
+        for arg in redact_arg_by_name:
+            values[arg] = '***'
```
PawelLipski: Huh but wait, won't it add a redacted `access_token`, ..., `token` arguments to the logs of **every** method now?

git_machete/utils.py
```diff
@@ -94,7 +94,20 @@ def debug(msg: Optional[str] = None) -> None:
...
+
+        redact_arg_by_value = ['ghp_', 'gho_', 'ghu_', 'ghs_', 'ghr_']
+        for arg, value in values.items():
+            if any(arg_ in str(value) for arg_ in redact_arg_by_value):
+                values[arg] = '***'
+
+        excluded_args = {'self'}
+        allowed_args = [arg for arg in args if arg not in excluded_args]
+        function_args = bold(f"({', '.join([arg+'='+str(values[arg]) for arg in allowed_args])})")
```
PawelLipski: Huh this is too deeply nested... pls extract some intermediate var(s)

git_machete/utils.py
amalota: @PawelLipski 

`__fire_github_api_request(method=POST, path=/repos/amalota/test_repo/pulls, token=***, request_body={'head': 'g', 'base': 'd', 'title': 'cd', 'body': '', 'draft': False}): 
firing a POST request to https://api.github.com/repos/amalota/test_repo/pulls with a bearer token and request body {"head": "g", "base": "d", "title": "cd", "body": "", "draft": false}`

git_machete/utils.py
amalota: @PawelLipski Hmm, yes it actually will add them to `values` but it won't print them because we iterate over `args` and not `values`, i'll fix it.

PawelLipski: Oh, okay... but still, it sets `values` even for keys that are NOT present in `args`, which is kinda buggy

git_machete/utils.py
amalota: Ok, make sense

git_machete/utils.py
```diff
@@ -94,11 +94,27 @@ def debug(msg: Optional[str] = None) -> None:
...
+
+        args_to_be_redacted_by_name = {'access_token', 'password', 'secret', 'token'}
+        for arg in args_to_be_redacted_by_name:
+            if arg in values.keys():
+                values[arg] = '***'
+
+        args_to_be_redacted_by_value = ['ghp_', 'gho_', 'ghu_', 'ghs_', 'ghr_']  # https://github.blog/2021-04-05-behind-githubs-new-authentication-token-formats/
+        for arg, value in values.items():
+            if any(arg_ in str(value) for arg_ in args_to_be_redacted_by_value):
+                values[arg] = '***'
```
PawelLipski: Or maybe even:
```suggestion
        args_to_be_redacted_by_name = {'access_token', 'password', 'secret', 'token'}
        args_to_be_redacted_by_value = ['ghp_', 'gho_', 'ghu_', 'ghs_', 'ghr_']  # https://github.blog/2021-04-05-behind-githubs-new-authentication-token-formats/

        for arg, value in values.items():
            if arg in args_to_be_redacted_by_name or any(arg_ in str(value) for arg_ in args_to_be_redacted_by_value):
                values[arg] = '***'
```

AFAICS that's equivalent, and easier to think of (just one loop, instead of two over two different collections)

PawelLipski: Also, `arg_` in `for`-comprehension shouldn't be called this way, rather `value_`

PawelLipski: `args_to_be_redacted_by_value` should also rather be called `values_to_be_redacted` 🤔

amalota: Hmm, i specifically separated them to not get confused between the two of them in one big loop, thought it would be easier/cleaner to read/understand, @PawelLipski wdyt? :D

git_machete/utils.py
```diff
@@ -94,11 +94,27 @@ def debug(msg: Optional[str] = None) -> None:
...
+
+        excluded_args = {'self'}
+        allowed_args = [arg for arg in args if arg not in excluded_args]
```
PawelLipski: See `git_machete.utils.excluding`

git_machete/utils.py
PawelLipski: > Hmm, i specifically separated them to not get confused between the two of them in one big loop

Okay, no strong conviction here, leaving to your judgement

PawelLipski (approved):

Issue #493, opened 2022-04-19
When GitHub token exists but is expired/invalid, `git machete github ...` doesn't clearly state where it's coming from


PawelLipski:
My GitHub personal token just expired.

```
$ g m github checkout-prs --all
Checking for open GitHub PRs...
GitHub API returned 401 HTTP status with error message: Unauthorized
Make sure that the GitHub API token provided by one of the: 
	1. GITHUB_TOKEN environment variable.
	2. Content of the ~/.github-token file.
	3. Current auth token from the gh GitHub CLI.
	4. Current auth token from the hub GitHub CLI.
is valid and allows for access to GET https://https://api.github.com/user.
```

Now a _quiz_: where did git-machete take my expired token from? I.e. in which of these 4 location should I actually replace the token with a newly-generated equivalent? It isn't stated in the message, while it's a pretty useful piece of info.

mkondratek:
I ve tried to use `create-pr` with that result:

```sh
mikko@mikko-MS-7C91:~/IdeaProjects/git-machete-intellij-plugin$ git machete github create-pr
Fetching origin...
Creating a PR from hook-executors-improvements to develop...Given endpoint: https://api.github.com/repos/VirtusLab/git-machete-intellij-plugin/pulls, not found in GitHub or you don't have the permission to access it (expired token?).
```

I have no token (generated and) set so that's the reason. After that - works.

PawelLipski:
Actually... after a consideration, lemme extract this as a separate issue, this is not strictly related

PawelLipski:
@amalota actually... isn't it done already?

PawelLipski:
Yes, it seems to be long done:

```
GitHub API returned 401 HTTP status with error message: Unauthorized
Make sure that the GitHub API token provided by GITHUB_TOKEN environment variable is valid and allows for access to ...
```

PR #548, opened 2022-06-10
Issue #493: When GitHub token is invalid/expired, provide information which token provider was used

amalota:


git_machete/github.py
```diff
@@ -131,6 +139,14 @@ def get_token_from_file_in_home_directory() -> Optional[str]:
             get_token_from_hub())
 
 
+def __get_github_token() -> Optional[str]:
+    return __get_github_token_and_provider().token if __get_github_token_and_provider() else None
```
PawelLipski: Nit: I'd extract the result of `__get_github_token_and_provider()` to a variable to avoid a double call

amalota:
Example messages:
Screenshot 2022-06-10 at 19 20 41

image

@PawelLipski I was thinking i could also add info that you can change the way you provide token and display possible providers as before, wdyt?

PawelLipski:
> i could also add info that you can change the way you provide token and display possible providers as before, wdyt?

Not bad an idea.. in case token is already present, you can just mention that the user can check different providers in the help, without listing them explicitly (as too long error messages cause TL;DR reactions inevitably)

git_machete/github.py
```diff
@@ -59,9 +59,13 @@ def __parse_pr_json(pr_json: Any) -> GitHubPullRequest:
                              repository_url=pr_json['head']['repo']['html_url'] if pr_json['head']['repo'] else None)
 
 
-def __get_github_token() -> Optional[str]:
-    def get_token_from_gh() -> Optional[str]:
+class GithubTokenAndTokenProvider(NamedTuple):
```
PawelLipski, line 62: Pls bump version + mention this entire change in release notes

PawelLipski (approved):

PR #551, opened 2022-06-17
Change the default logic for inferring git machete definition file path to what is was before

### LINKED ISSUE #517: per-worktree machete option
﻿https://xkcd.com/1172/

i actually found this very helpful, would be great if there was an option for it.

amalota:


git_machete/git_operations.py
```diff
@@ -318,9 +319,11 @@ def get_config_attr_or_none(self, key: str) -> Optional[str]:
         self.__ensure_config_loaded()
         return self.__config_cached.get(key.lower())
 
-    def get_boolean_config_attr(self, key: str) -> bool:
+    def get_boolean_config_attr(self, key: str, default_value: bool = True) -> bool:
```
PawelLipski: Pls remove the default value for this param completely. The longer I live, the more convinced I am that the default param values are a cancer

image

git_machete/client.py
```diff
@@ -934,7 +934,8 @@ def fp_hash(branch_: LocalBranchShortName) -> Optional[FullCommitHash]:
         hook_path = self.__git.get_hook_path("machete-status-branch")
```
PawelLipski, line 934: Pls re-review docs for `file` and `config`, there were some discrepancies

PawelLipski (approved):

PawelLipski:
> 0.12%

Only ~270 PRs like this left for the full coverage 😅

PR #559, opened 2022-06-21
Issue #372: `github create-pr` should take branch tracking data into account

### LINKED ISSUE #372: Subtask of #552: `github create-pr` doesn't take branch tracking data into account
I've got two remetes `foo` and `bar`, both of which correspond to Github &mdash; neither is called `origin`.

I've got a branch called `qux` which I've pushed to `bar` (`git push -u bar qux`).

Now when running `git machete github create-pr` while on `qux`, I'd expect the PR to opened against `bar` repo, but instead I'm getting:

`Multiple non-origin remotes correspond to GitHub in this repository: foo, bar, aborting`

This message would make sense only if `qux` was NOT yet pushed to any of these repos.

amalota:


RELEASE_NOTES.md
```diff
@@ -4,6 +4,7 @@
 
 - fixed: `git machete` now correctly infers remote for fetching of branch when the branch is associated with more than one remote
 - fixed: `git machete github retarget-pr` now takes branch tracking data into account when finding out where (in what GitHub organization/repository) to create a PR
+- fixed: `git machete github create-pr` now takes branch tracking data into account when finding out where (in what GitHub organization/repository) to create a PR
```
PawelLipski: I'd squash them to one:
```suggestion
- fixed: `git machete github create-pr` and `retarget-pr` now take branch tracking data into account when finding out where (in what GitHub organization/repository) to create a PR
```

git_machete/client.py
```diff
@@ -2086,7 +2086,15 @@ def __get_remote_and_organization_and_repository_name_for_custom_url(self) -> Re
                                                   repository=self.__git.get_config_attr_or_none("machete.github.repository"))
 
     def create_github_pr(
-            self, *, head: LocalBranchShortName, opt_draft: bool, opt_onto: Optional[LocalBranchShortName]) -> None:
+            self,
+            *,
```
PawelLipski, line 2087: Huh I've already forgotten... what's this `*` for in this method? 🤔

git_machete/client.py
```diff
@@ -2086,7 +2086,15 @@ def __get_remote_and_organization_and_repository_name_for_custom_url(self) -> Re
...
+            head: LocalBranchShortName,
+            opt_draft: bool,
+            opt_onto: Optional[LocalBranchShortName]
+    ) -> None:
+        for remote in self.__git.get_remotes():
```
PawelLipski: Hmmm I'm skeptical if this fetch should really be here (same for `retarget-pr`)... I mean, the trade-off is as follows:
* with fetch: ++slightly better chance of finding the right org/repo on the off chance that the branch appeared in the meantime in remote repo, --well, additional time spent fetching on each `git machete github create-pr`
* without fetch: --slightly lower chance of detecting the right org/repo, ++time saved, less confusion for the user (_lol why fetch all remotes now?_)

TBH I'd vote for skipping the fetch here, that's going to become a PITA very quickly

PawelLipski: Also! note that if you're about to open a PR for a branch... then it isn't likely that this branch has been pushed by someone else in the meantime (and that's what `git fetch` would help us detect here)... it's usually just you and only you who pushes that branch! `git fetch` won't help much

tests/mockers.py
```diff
@@ -55,6 +55,9 @@ def execute(self, command: str) -> "GitRepositorySandbox":
         subprocess.check_output(command, stderr=subprocess.STDOUT, shell=True)
         return self
 
+    def execute_with_output(self, command: str) -> "GitRepositorySandbox":
```
PawelLipski:
```suggestion
    def execute_and_return_stdout(self, command: str) -> "GitRepositorySandbox":
```
?

tests/test_github.py
```diff
@@ -594,6 +598,146 @@ def test_github_create_pr_missing_base_branch_on_remote(self) -> None:
...
+    @mock.patch('git_machete.cli.exit_script', mock_exit_script)
+    @mock.patch('git_machete.client.MacheteClient.ask_if', mock_ask_if)
+    # We need to mock GITHUB_REMOTE_PATTERNS in the tests for `test_github_create_pr` due to `git fetch` executed by `create-pr` subcommand.
+    @mock.patch('git_machete.github.GITHUB_REMOTE_PATTERNS', FAKE_GITHUB_REMOTE_PATTERNS)
+    @mock.patch('git_machete.utils.run_cmd', mock_run_cmd)  # to hide git outputs in tests
+    @mock.patch('git_machete.options.CommandLineOptions', FakeCommandLineOptions)
+    @mock.patch('urllib.error.HTTPError', MockHTTPError)  # need to provide read() method, which does not actually reads error from url
+    @mock.patch('urllib.request.Request', git_api_state_for_test_github_create_pr_with_multiple_non_origin_remotes.new_request())
+    @mock.patch('urllib.request.urlopen', MockContextManager)
+    @mock.patch('builtins.input', mock_input)
```
PawelLipski, line 626: Wow I didn't know the package is called `builtins` for stuff like `input` 😯

amalota: Me neither until now :D

git_machete/client.py
amalota: @PawelLipski Parameters after `*` are keyword-only parameters and may only be passed used keyword arguments. I would say the only use case i see is when you change order of the arguments, it protects you from passing values to wrong arguments, not sure if this is really necessary here since the function is only used once and even if you don't put '*' in the function def you can still pass parameters using keyword-args, but overall pretty interesting feature.

git_machete/client.py
amalota: Ok, that last point convinced me - i'll remove the fetching.

git_machete/client.py
PawelLipski: Oh! okay, makes sense, let's leave as is, indeed such extra protection is useful (could even be applied in more places TBH... I'll open an issue)

PawelLipski: https://github.com/VirtusLab/git-machete/issues/568

tests/mockers.py
amalota: This was only temporary

git_machete/client.py
```diff
@@ -1627,7 +1627,7 @@ def __pick_remote(
               "repository, or 'n' to skip this branch, or " \
               "'q' to quit the traverse: " if is_called_from_traverse \
               else f"Select number 1..{len(rems)} to specify the destination remote " \
-                   "repository, or 'q' to quit creating pull request: "
+                   "repository, or 'q' to quit creating pull request:"
```
amalota: The space after the one before last character `:` is unnecessary and makes it difficult for tests (pep8 nad general checks will throw error about trailing whitespaces) and even if i can silence pep8 i can't do anything about general checks (well actually i can but it would involve changing general checks so that they ignore tests dir - not sure if it is a correct solution)

PawelLipski: Huh the space feels kinda necessary TBH here in the user-facing prompt... I'd leave it as is

Let's use `# noqa` + exclude tests from the trailing whitespace check (in fact you can exclude all Python  sources from that check, coz they're already covered by pep8 anyway)

PawelLipski (approved):

Issue #562, opened 2022-06-22
`git machete` functions for resolving remote of a branch or a branch's remote counterpart assume that a branch is only present in one remote 


amalota:
# Functions:

### `get_strict_remote_for_fetching_of_branch`

```   
 def get_strict_remote_for_fetching_of_branch(self, branch: LocalBranchShortName) -> Optional[str]:
        remote = self.get_config_attr_or_none(f"branch.{branch}.remote")
        return remote.rstrip() if remote else None
```
**-> returns only one remote, the one that was added last**

Used by commands:
- `discover`
- `status`
- `traverse`
- `github create-pr` through the `__sync_before_creating_pr` function

### `get_inferred_remote_for_fetching_of_branch`
```
    def get_inferred_remote_for_fetching_of_branch(self, branch: LocalBranchShortName) -> Optional[str]:
        # Since many people don't use '--set-upstream' flag of 'push', we try to infer the remote instead.
        for remote in self.get_remotes():
            if f"{remote}/{branch}" in self.get_remote_branches():
                return remote
        return None
```
**-> returns only one remote, the first one that meets the requirements from alphabetically ordered remote-branch pairs**

Used by commands:
- `discover`
- `status`
- `traverse`
- `github create-pr` through the `__sync_before_creating_pr` function

### `__get_inferred_counterpart_for_fetching_of_branch`
```
    def __get_inferred_counterpart_for_fetching_of_branch(self, branch: LocalBranchShortName) -> Optional[RemoteBranchShortName]:
        for remote in self.get_remotes():
            if f"{remote}/{branch}" in self.get_remote_branches():
                return RemoteBranchShortName.of(f"{remote}/{branch}")
        return None
```
**-> returns only one local branch remote counterpart, the first one that meets the requirements from alphabetically ordered remote-branch pairs**

Used by commands:
- `list`
- `add`
- `up`
- `discover`
- `status`
- `traverse`
- `advance`
- `fork-point`
- 
### `get_strict_counterpart_for_fetching_of_branch`
```
    def get_strict_counterpart_for_fetching_of_branch(self, branch: LocalBranchShortName) -> Optional[RemoteBranchShortName]:
        if self.__counterparts_for_fetching_cached is None:
            self.__load_branches()
        return self.__counterparts_for_fetching_cached.get(branch)
```
**-> returns only one local branch remote counterpart,, the one that was added last**

Used by commands:
- `github create-pr` through the `__sync_before_creating_pr` function
- `add`
[...]

PR #564, opened 2022-06-22
Issue #563: `get_inferred_remote_for_fetching_of_branch()` now correctly infers remote for fetching of branch when the branch is associated with more than one remote

### LINKED ISSUE #563: Subtask of #562: `get_inferred_remote_for_fetching_of_branch`  -> returns only one remote, the first one that meets the requirements from alphabetically ordered remote-branch pairs
Subtask of #562

amalota:


git_machete/git_operations.py
```diff
@@ -437,10 +437,15 @@ def get_committer_unix_timestamp_by_revision(self, revision: AnyBranchName) -> i
 
     def get_inferred_remote_for_fetching_of_branch(self, branch: LocalBranchShortName) -> Optional[str]:
         # Since many people don't use '--set-upstream' flag of 'push', we try to infer the remote instead.
-        for remote in self.get_remotes():
-            if f"{remote}/{branch}" in self.get_remote_branches():
-                return remote
-        return None
+        remotes = self.get_remotes()
+        for remote in remotes:
+            self.fetch_remote(remote)  # NOT SURE IF FETCHING REMOTES IS REALLY NECESSARY
```
PawelLipski: Nopes! I wouldn't do it here under the hood, this might lead to unexpected fetches where we don't want them

git_machete/git_operations.py
```diff
@@ -437,10 +437,15 @@ def get_committer_unix_timestamp_by_revision(self, revision: AnyBranchName) -> i
...
+        remote_branches = self.get_remote_branches()
+        remotes_containing_branch: List[str] = [remote for remote in remotes if f'{remote}/{branch}' in remote_branches]
+        if len(remotes_containing_branch) > 1 or len(remotes_containing_branch) == 0:
+            return None
```
PawelLipski, line 449: I'd add a `debug` log in case there's more than one remote

RELEASE_NOTES.md
```diff
@@ -1,5 +1,9 @@
 # Release notes
 
+## New in git-machete 3.11.2
+
+- fixed: `git machete` now correctly infers remote for fetching of branch
```
PawelLipski: I'd specify the exact case here, but ofc with the low level of detail, as adequate for release notes

git_machete/__init__.py
```diff
@@ -1 +1 @@
-__version__ = '3.11.1'
+__version__ = '3.11.2'
```
PawelLipski: Btw! this PR isn't correctly names, `get_inferred_remote_for_fetching_of_branch` still doesn't (and shouldn't!) take into account the branch tracking data (which is typically understood as `branch.{...}.remote` git config)

PawelLipski: It's the `get_strict_...` versions that rely on `branch.{...}.remote` git config

PawelLipski: `get_inferred_...` methods are a fallback in case branch tracking data is **not** set. From `get_inferred_remote_for_fetching_of_branch`:

```python
# Since many people don't use '--set-upstream' flag of 'push', we try to infer the remote instead.
```

PawelLipski: You may actually add some docstrings about that in code to make it clearer for the future devs

amalota:
@PawelLipski i am wondering if any tests are needed here, same for https://github.com/VirtusLab/git-machete/pull/566 PR. There are no tests just for git-operations and the functions that were changed are used in many commands (see https://github.com/VirtusLab/git-machete/issues/562) so even if we wanted to test it, should we just select 1 command and test a scenario that would involve newly implemented logic?

PawelLipski:
> so even if we wanted to test it, should we just select 1 command and test a scenario that would involve newly implemented logic?

As discussed... I'd cover this scenario in tests of e.g. `github create-pr`

RELEASE_NOTES.md
```diff
@@ -1,5 +1,9 @@
 # Release notes
 
+## New in git-machete 3.11.2
+
+- fixed: `git machete` now correctly infers remote for fetching of branch when the branch is associated with more than one remote
```
```suggestion
- fixed: `git machete` now correctly infers remote for fetching of branch when the branch is found in more than one remote
```

As _associated_ would suggest `branch.<branch>.remote` config being in place... while here it's just about a branch with the given name being present in multiple remotes (which can theoretically be a coincidence!)

git_machete/git_operations.py
```diff
@@ -435,18 +435,27 @@ def get_committer_unix_timestamp_by_revision(self, revision: AnyBranchName) -> i
             self.__load_branches()
         return self.__committer_unix_timestamp_by_revision_cached.get(revision.full_name(), 0)
 
-    def get_inferred_remote_for_fetching_of_branch(self, branch: LocalBranchShortName) -> Optional[str]:
-        # Since many people don't use '--set-upstream' flag of 'push', we try to infer the remote instead.
-        for remote in self.get_remotes():
-            if f"{remote}/{branch}" in self.get_remote_branches():
-                return remote
-        return None
+    def __get_remotes_associated_with_branch(self, branch: LocalBranchShortName, remotes: Optional[List[str]] = None) -> List[str]:
```
PawelLipski: Just
```suggestion
    def __get_remotes_containing_branch(self, branch: LocalBranchShortName, remotes: Optional[List[str]] = None) -> List[str]:
```

git_machete/git_operations.py
```diff
@@ -435,18 +435,27 @@ def get_committer_unix_timestamp_by_revision(self, revision: AnyBranchName) -> i
...
+        remotes = remotes if remotes else self.get_remotes()
+        remote_branches = self.get_remote_branches()
+        return [remote for remote in remotes if f'{remote}/{branch}' in remote_branches]
+
+    def get_inferred_remote_for_fetching_of_branch(self, branch: LocalBranchShortName, remotes: Optional[List[str]] = None) -> Optional[str]:
+        remotes_associated_with_branch: List[str] = self.__get_remotes_associated_with_branch(branch=branch, remotes=remotes)
```
PawelLipski: Ditto

RELEASE_NOTES.md
```diff
@@ -1,5 +1,9 @@
 # Release notes
 
+## New in git-machete 3.11.2
+
+- fixed: `git machete` now correctly infers remote for fetching of branch when the branch is found in more than one remote
```
PawelLipski, line 5: Maybe actually:
```suggestion
- fixed: `git machete` now correctly does not infer a remote for fetching of branch when the branch is found in more than one remote
```

git_machete/git_operations.py
```diff
@@ -435,18 +435,27 @@ def get_committer_unix_timestamp_by_revision(self, revision: AnyBranchName) -> i
...
+        remotes = remotes if remotes else self.get_remotes()
+        remote_branches = self.get_remote_branches()
+        return [remote for remote in remotes if f'{remote}/{branch}' in remote_branches]
+
+    def get_inferred_remote_for_fetching_of_branch(self, branch: LocalBranchShortName, remotes: Optional[List[str]] = None) -> Optional[str]:
+        remotes_containing_branch: List[str] = self.__get_remotes_containing_branch(branch=branch, remotes=remotes)
+        if len(remotes_containing_branch) > 1 or len(remotes_containing_branch) == 0:
+            debug(f'Can\'t infer remote for fetching of branch.\n'
+                  f'There are {len(remotes_containing_branch)} remotes: {",".join(remotes_containing_branch)} '
+                  f'associated with {branch} branch.')
```
```suggestion
                  f'containing branch '{branch}'.')
```

RELEASE_NOTES.md
amalota: @PawelLipski Why? The code still infers the remote, just the correct one, i would't change change it like that IMHO

PawelLipski: Okay, let's leave as is, not insisting

PawelLipski (commented):
Okay! so anything left to be done here? adding tests? or are they added collectively on the subsequent PR

PawelLipski (approved):

PR #572, opened 2022-07-08
Issue #362: `GitRepositorySandbox.new_repo` has a side effect of changing the current directory

### LINKED ISSUE #362: `GitRepositorySandbox.new_repo` has a side effect of changing the current directory
```python
    def new_repo(self, *args: str) -> "GitRepositorySandbox":
        os.chdir(args[0])
        opts = args[1:]
        self.execute(f"git init {' '.join(opts)}")
        return self
```

Some `os.chdir` back to the original position would be useful here... or, if the current behavior makes more sense, maybe a rename to this method?

amalota:


tests/mockers.py
```diff
@@ -55,10 +55,13 @@ def execute(self, command: str) -> "GitRepositorySandbox":
         subprocess.check_output(command, stderr=subprocess.STDOUT, shell=True)
         return self
 
-    def new_repo(self, *args: str) -> "GitRepositorySandbox":
+    def new_repo(self, *args: str, switch_dir_to_new_repo: bool = True) -> "GitRepositorySandbox":
```
PawelLipski, line 58: How many times is this method called across the project? TBH I'd prefer if `switch_dir_to_new_repo` param did NOT have a default value... let's make the behavior as explicit as possible

PawelLipski, line 58: How many times is this method called across the project? TBH I'd prefer if `switch_dir_to_new_repo` param did NOT have a default value... let's make the behavior as explicit as possible

amalota: It's called 45 times across tests, so a lot, because of that i would stick with default value, wdyt?

PawelLipski: Huh okay... let's live with that then

PawelLipski (approved):

PR #577, opened 2022-07-12
Issue #573: Prioritise `origin` remote when inferring remote 

### LINKED ISSUE #573: `github create-pr` and `retarget-pr` probably should NOT take branch tracking data into account, after all 😅
Follow up to #552... I haven't considered the common flow when working with forks 😅 

1. I've got an `origin` remote corresponding to the original (non-fork) repo where I can**not** push branches
2. I've got a `PawelLipski` remote corresponding to the fork
3. I push branch `foo` to `PawelLipski` (as I cannot push to `origin`)
4. I run `git machete github create-pr`... which then (after the changes from #552) picks `PawelLipski` as the remote to open a PR in
5. <no profit>, as I didn't want to open a PR in a fork, but in the original repo :////

So... reverting the changes related to #552 is probably a better idea, in fact :///

Sorry for the confusion... still, the changes #562 are still valid & useful.

TBD if `retarget-pr` is also affected (it looks to me that **yes**, also should be reverted)

amalota:


git_machete/client.py
```diff
@@ -2058,19 +2058,19 @@ def __derive_remote_and_github_org_and_repo(self, branch_used_for_tracking_data:
                     'If that is the case, you can provide repository information explicitly, via these 3 git config keys: '
                     '`machete.github.{remote,organization,repository}`\n'))
```
PawelLipski, line 2060: Okay! hope this will be some sensible compromise... I'd still expect some 1 or 2 corner cases to crop up, but nothing major

PawelLipski (approved):

PR #579, opened 2022-07-13
Issue #509: Tidy up `client.status` function

### LINKED ISSUE #509: Tidy up `def status`
The logic is pretty complicated, mostly due to the presence of the underlying DFS that collects the chain of _next sibling for ancestor_ for each branch.

This DFS can stay... but a few suggestion on what could be refactored in the logic:
* no longer use ANSI escape code for anything but the final printing of status; in particular, the logic for `opt_list_commits` should not rely on comparing ANSI escape codes 😅 instead, use some more abstract representation of colors, to be mapped into either ANSI escape codes or its ASCII equivalents (in case of `utils.ascii_only`)
* split into smaller methods
* naming of `prefix_dfs` and its local variables can be improved
* `next_sibling_of_ancestor_by_branch` can be a dictionary rather than list of pairs; the main loop `for branch, next_sibling_of_ancestor in next_sibling_of_ancestor_by_branch` can probably go over `self.managed_branches` rather than over `next_sibling_of_ancestor_by_branch`

amalota:


git_machete/client.py
```diff
@@ -747,6 +747,8 @@ def traverse(
                         self.__down_branches[upstream])
                     if branch in self.__annotations:
                         del self.__annotations[branch]
+                    if branch in self.managed_branches:
```
amalota: When trying to iterate over `managed_branches` in `status` function i encounter bug where `self.managed_branches` were not updated properly ->`slide_out` function removes slid out branches from the `self.managed_branches` but `traverse` did not and i don't see a reason why it should not, @PawelLipski wdyt?

git_machete/client.py
```diff
@@ -915,15 +916,15 @@ def fp_hash(branch_: LocalBranchShortName) -> Optional[FullCommitHash]:
         # in order to render the leading parts of lines properly.
         branch: LocalBranchShortName
         for branch in self.up_branch:
-            upstream = self.up_branch[branch]
+            up_branch = self.up_branch[branch]
```
amalota: @PawelLipski So i changed it because i think that ancestor of the branch != upstream branch

git_machete/client.py
```diff
@@ -947,7 +948,7 @@ def print_line_prefix(branch_: LocalBranchShortName, suffix: str) -> None:
             out.write(colored(suffix, edge_color[branch_]))
 
         next_sibling_of_ancestor: List[Optional[LocalBranchShortName]]
-        for branch, next_sibling_of_ancestor in next_sibling_of_ancestor_by_branch:
+        for branch, next_sibling_of_ancestor in next_sibling_of_ancestor_by_branch.items():
```
amalota, line 952: @PawelLipski https://github.com/VirtusLab/git-machete/issues/509 suggests to make `next_sibling_of_ancestor_by_branch` a dict (which i agree with and changed so) and iterate over `self.managed_branches` instead of `next_sibling_of_ancestor_by_branch`. So modified iteration works fine in tests but now `discover` does't work properly locally, i am betting that the order of the branches in `self.managed_branches` is different than in `next_sibling_of_ancestor_by_branch`, but it is still weird that all the tests pass. I will investigate it further

amalota: Yeah, as i suspected, somehow `self.managed_branches` has different branch order when run with `status` than run with `discover` (which also executes `status()` function after discovering the tree)

amalota: I guess it's to be expected as `discover` creates tree from scratch and orders branches alphabetically and `status` does dfs. To sum up i think the iteration should stay as it was before, @PawelLipski wdyt?

git_machete/client.py
PawelLipski: Huh TBH I use all the terms interchangably, `parent branch` == `upstream branch` == `up branch`... although `upstream` is somewhat brittle, as it also corresponds to `remote tracking branch` (see e.g. `git show @{upstream}`)

git_machete/client.py
PawelLipski: Pls investigate this particular change on a separate PR as discussed

git_machete/client.py
PawelLipski: https://github.com/VirtusLab/git-machete/issues/582

git_machete/client.py
PawelLipski: I wasn't aware of the caveat with `discover`... so I'm not insisting, let's stick to iterating over `next_sibling_of_ancestor_by_branch` as it used to be.

PawelLipski: `dict` -> maybe `OrderedDict` for clarity!

git_machete/utils.py
```diff
@@ -295,3 +296,12 @@ def format_choice(c: str) -> str:
         else:
             return colored(c, AnsiEscapeCodes.ORANGE)
     return f" ({', '.join(map_truthy_only(format_choice, choices))}) "
+
+
+class Color(Enum):
```
PawelLipski: Hmmm this still doesn't abstract away ANSI escape codes... it's just a 1:1 wrapper over such codes 🤔

PawelLipski: What I'd like to see would rather be an enum that can only hold values that represent **edge colors** (so `GREEN`, `RED`, `GRAY`, `YELLOW`)... that will be translated to ANSI escape codes only when needed.

The enum itself should know nothing of ANSI escape codes.

tests/test_status.py
```diff
@@ -178,8 +178,9 @@ def test_squashed_branch_recognized_as_merged(self, mocker: Any) -> None:
             """,
         )
 
-        # traverse then slides out the branch
+        # traverse then slide out the feature branch
```
PawelLipski, line 181: `feature` or `child`? 🤔

amalota: `feature`

git_machete/client.py
```diff
@@ -968,7 +968,8 @@ def print_line_prefix(branch_: LocalBranchShortName, suffix: str) -> None:
                             fp_branches_formatted: str = " and ".join(
                                 sorted(underline(lb_or_rb) for lb, lb_or_rb in fp_branches_cached[branch]))
                             fp_suffix: str = " %s %s %s seems to be a part of the unique history of %s" % \
-                                             (colored(utils.get_right_arrow(), AnsiEscapeCodes.RED), colored("fork point ???", AnsiEscapeCodes.RED),
```
PawelLipski, line 973: A nuance: here I'd leave `AnsiEscapeCodes` as it used to be. This line is concerned with **rendering**, not **logic**... and hence you're more than justified to operate on `AnsiEscapeCodes`. In fact, using `EdgeColor` doesn't make any sense here

git_machete/client.py
```diff
@@ -1016,7 +1017,8 @@ def print_line_prefix(branch_: LocalBranchShortName, suffix: str) -> None:
                     prefix = "REVERTING "
                 else:
                     prefix = ""
-                current = "%s%s" % (bold(colored(prefix, AnsiEscapeCodes.RED)), bold(underline(branch, star_if_ascii_only=True)))
+                current = "%s%s" % (bold(colored(prefix, to_ansi_escape_code(EdgeColor.RED))),
```
PawelLipski: Ditto, it's just rendering (text formatting), nothing to do with edge colors

PawelLipski: Ditto, it's just rendering (text formatting), nothing to do with edge colors

git_machete/client.py
```diff
@@ -1025,12 +1027,14 @@ def print_line_prefix(branch_: LocalBranchShortName, suffix: str) -> None:
             s, remote = self.__git.get_combined_remote_sync_status(branch)
             sync_status = {
                 SyncToRemoteStatuses.NO_REMOTES: "",
-                SyncToRemoteStatuses.UNTRACKED: colored(" (untracked)", AnsiEscapeCodes.ORANGE),
+                SyncToRemoteStatuses.UNTRACKED: colored(" (untracked)", to_ansi_escape_code(EdgeColor.ORANGE)),
```
PawelLipski: Same, nothing to do with `EdgeColor` (applies to all values in `sync_status`)

git_machete/utils.py
```diff
@@ -295,3 +296,24 @@ def format_choice(c: str) -> str:
         else:
             return colored(c, AnsiEscapeCodes.ORANGE)
     return f" ({', '.join(map_truthy_only(format_choice, choices))}) "
+
+
+class EdgeColor(Enum):
+    DIM = auto()
+    UNDERLINE = auto()
```
PawelLipski: `UNDERLINE` is not an edge color :/// `UNDERLINE` is just a way to format text... which is a different level of abstraction than edge colors!

git_machete/utils.py
```diff
@@ -295,3 +296,24 @@ def format_choice(c: str) -> str:
         else:
             return colored(c, AnsiEscapeCodes.ORANGE)
     return f" ({', '.join(map_truthy_only(format_choice, choices))}) "
+
+
+class EdgeColor(Enum):
+    DIM = auto()
```
PawelLipski: The (abstract) edge color is `GRAY`... it just happens to be implemented by `DIM` coloration in text terminal (but in IntelliJ plugin, it's implemented with the actual gray color)

git_machete/utils.py
```diff
@@ -295,3 +296,24 @@ def format_choice(c: str) -> str:
...
+    UNDERLINE = auto()
+    GREEN = auto()
+    YELLOW = auto()
+    ORANGE = auto()
```
PawelLipski: There is no such edge color in git-machete as orange :/

In fact... I'd even go further. I wouldn't use colors to name the entries in this enum at all.

I'd rather use sth like here:

https://github.com/VirtusLab/git-machete-intellij-plugin/blob/master/backend/api/src/main/java/com/virtuslab/gitmachete/backend/api/SyncToParentStatus.java

PawelLipski (changes requested):
See the linked enum in the intellij-plugin... hope that'll help make sense of the separation of concerns (logic vs rendering) we're striving to achieve here!

git_machete/utils.py
amalota: Yeah, it makes more sense to use status than color since its more meaningful

git_machete/client.py
```diff
@@ -1049,7 +1047,7 @@ def print_line_prefix(branch_: LocalBranchShortName, suffix: str) -> None:
         sys.stdout.write(out.getvalue())
         out.close()
 
-        yellow_edge_branches = [k for k, v in edge_color.items() if v == AnsiEscapeCodes.YELLOW]
+        yellow_edge_branches = [k for k, v in sync_to_parent_status.items() if v == SyncToParentStatus.InSyncButForkPointOff]
```
PawelLipski: For consistency... I'd also rename `yellow_edge_branches` and `warn_on_yellow_edges` accordingly (so that edge colors are **not** mentioned)

git_machete/client.py
```diff
@@ -1049,7 +1047,7 @@ def print_line_prefix(branch_: LocalBranchShortName, suffix: str) -> None:
...
         if yellow_edge_branches and warn_on_yellow_edges:
             if len(yellow_edge_branches) == 1:
                 first_part = f"yellow edge indicates that fork point for `{yellow_edge_branches[0]}` is probably incorrectly inferred,\n" \
```
PawelLipski: Albeit I'd still leave the `yellow` etc. naming in user-facing messages, for the sake of clarity

git_machete/utils.py
```diff
@@ -295,3 +296,25 @@ def format_choice(c: str) -> str:
         else:
             return colored(c, AnsiEscapeCodes.ORANGE)
     return f" ({', '.join(map_truthy_only(format_choice, choices))}) "
+
+
+class SyncToParentStatus(Enum):
```
PawelLipski, line 301: :benissimo:

git_machete/utils.py
```diff
@@ -295,3 +296,25 @@ def format_choice(c: str) -> str:
...
+    InSync = auto()
+    MergedToParent = auto()
+    InSyncButForkPointOff = auto()
+    OutOfSync = auto()
+
+
+sync_to_parent_status_to_edge_color_map: Dict[SyncToParentStatus, str] = {
```
PawelLipski, line 308: Nit: pls keep the order of keys the same in both dicts for easier comparison

git_machete/client.py
amalota: I guess you meant to write `(so that edge colors are not mentioned)`, right?

PawelLipski: are NOT mentioned, yes! sorry

git_machete/client.py
```diff
@@ -874,29 +874,28 @@ def traverse(
...
 
-        for upstream in self.__roots:
-            prefix_dfs(upstream, accumulated_path_=[])
+        for root in self.__roots:
+            prefix_dfs(root, accumulated_path_=[])
 
         out = io.StringIO()
-        edge_color: Dict[LocalBranchShortName, str] = {}
+        sync_to_parent_status: Dict[LocalBranchShortName, SyncToParentStatus] = {}
         fp_hash_cached: Dict[LocalBranchShortName, Optional[FullCommitHash]] = {}  # TODO (#110): default dict with None
```
PawelLipski: While you're here... pls rename such `fp_` variables/functions to `fork_point_`

git_machete/client.py
```diff
@@ -945,22 +944,23 @@ def print_line_prefix(branch_: LocalBranchShortName, suffix: str) -> None:
...
                         # Rare case, but can happen e.g. due to reflog expiry.
                         commits: List[GitLogEntry] = []
-                    elif edge_color[branch] == AnsiEscapeCodes.DIM:
+                    elif sync_to_parent_status[branch] == SyncToParentStatus.MergedToParent:
                         commits = []
-                    elif edge_color[branch] == AnsiEscapeCodes.YELLOW:
+                    elif sync_to_parent_status[branch] == SyncToParentStatus.InSyncButForkPointOff:
                         commits = self.__git.get_commits_between(self.up_branch[branch].full_name(), branch.full_name())
-                    else:  # (AnsiEscapeCodes.RED, AnsiEscapeCodes.GREEN):
+                    else:  # (SyncToParentStatus., OutOfSyncSyncToParentStatus.InSync):
```
PawelLipski: Whoops pls fix the comment

PawelLipski (approved):

PR #583, opened 2022-07-15
Fix bug in `traverse()` function where `self.managed_branches` were not updated properly after sliding out the branch

amalota:


git_machete/client.py
```diff
@@ -747,6 +747,8 @@ def traverse(
                         self.__down_branches[upstream])
                     if branch in self.__annotations:
                         del self.__annotations[branch]
+                    if branch in self.managed_branches:
```
PawelLipski, line 750: Could you provide a regression test that would've captured the bug?

amalota: @PawelLipski I don't think it's doable - the state of `self.managed_branches` is not being preserved between commands since commands read definition file and `self.managed_branches` is not used when saving branch tree (probably why it wasn't caught in the first place)

PawelLipski: Oh! okay... could you remind me how you discovered the bug in the first place?

amalota: I think i was debugging the `traverse` function and saw that `self.managed_branches` wasn't updated

PawelLipski: Ok, let's merge as is then

PawelLipski (approved):

PawelLipski:
Pls just rename the PR accordingly before merge (so that it's clear what was done from the resulting commit message after the squash)

PR #594, opened 2022-07-23
Fix fork point being incorrectly inferred when parent branch has been pushed immediately after being created

PawelLipski:
Another fix inspired by @wbar ;D tests to be added

### DIFF
--- a/git_machete/client.py
+++ b/git_machete/client.py
@@ -1395,8 +1395,11 @@ class MacheteClient:
                            # The rare case of a no-op rebase, the exact wording
                            # likely depends on git version
                            gs_ == f"rebase finished: {branch.full_name()} onto {hash_}" or
-                           gs_ == f"rebase -i (finish): {branch.full_name()} onto {hash_}"
-                           )
+                           gs_ == f"rebase -i (finish): {branch.full_name()} onto {hash_}" or
+                           # For remote branches, let's NOT include the pushes,
+                           # as a branch can be pushed directly after being created,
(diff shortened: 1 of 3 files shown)


git_machete/client.py
```diff
@@ -1395,8 +1395,11 @@ def is_excluded_reflog_subject(hash_: str, gs_: str) -> bool:
                            # The rare case of a no-op rebase, the exact wording
                            # likely depends on git version
                            gs_ == f"rebase finished: {branch.full_name()} onto {hash_}" or
-                           gs_ == f"rebase -i (finish): {branch.full_name()} onto {hash_}"
-                           )
+                           gs_ == f"rebase -i (finish): {branch.full_name()} onto {hash_}" or
+                           # For remote branches, let's NOT include the pushes,
+                           # as a branch can be pushed directly after being created,
+                           # which might lead to fork point being inferred too *late* in the history
+                           gs_ == "update by push")
```
PawelLipski, line 1402: Regression test scenario:

1. create `foo` out of `master`, commit sth on `foo`, add `foo` on top of `master` into .git/machete
2. create `bar` out of `foo`
3. push `bar` to remote
4. commit sth on `bar`

Before this fix: you'll see a yellow edge b/w `foo` and `master`: the commit pointed by `foo` will be incorrectly inferred to be its fork point, coz it appears on reflog of `bar`

After this fix: you'll see green edge b/w `master` and `foo`, as expected

PawelLipski (commented):
Approving!

PR #602, opened 2022-07-27
Issue #595: Don't use formatting codes when output is not a terminal

### LINKED ISSUE #595: Don't use formatting codes when output is not a terminal
# Overview

I just discovered `git machete`. The very first thing I did was run `git machete` by itself to see the help output, but it was much larger than a screenful of text, so I tried:

```
git machete | less
```

And it barfed terminal control codes all over my screen.

I expect to be able to pipe help output into a pager  without having to deal with visible escape codes.

See bove

```
git version 2.35.3
git machete version 3.11.3
```

amalota:


git_machete/cli.py
```diff
@@ -61,9 +61,7 @@ def get_help_description(command: str = None) -> str:
             usage_str += underline(hdr) + '\n\n'
             for cm in cmds:
                 alias = f", {alias_by_command[cm]}" if cm in alias_by_command else ""
-                usage_str += ("    %s%-18s%s%s" % (
-                    # bold(...) can't be used here due to the %-18s format specifier
-                    AnsiEscapeCodes.BOLD, cm + alias, AnsiEscapeCodes.ENDC, short_docs[cm]))
+                usage_str += f'    {bold(cm + alias) : <{18 if utils.ascii_only else 26}}{short_docs[cm]}'
```
PawelLipski, line 64: Wow I didn't know this syntax 😯

git_machete/cli.py
```diff
@@ -497,6 +491,10 @@ def launch(orig_args: List[str]) -> None:
         cli_opts.validate()
         set_utils_global_variables(cli_opts)
 
+        if not orig_args:
```
PawelLipski, line 494: Oh! so basically the problem was in the lack of call to `set_utils_global_variables`?

PawelLipski (approved):
Tested, works fine now!

git_machete/cli.py
amalota: Yes, that was the main problem + small changes above.

PR #604, opened 2022-07-28
Issue #597: Handle Temporary Redirect when a repository or organization name was changed

### LINKED ISSUE #597: Temporary Redirect when creating a new pull request
# Overview

Running `git machete github create-pr` results in an error message: `GitHub API returned 307 HTTP status with error message: Temporary Redirect`

Ability to create a PR using `git-machete`

Http request to Github API fails which means that PR is not being made

```
git --version
git version 2.32.1 (Apple Git-133)

git machete --version
git machete version 3.11.3
```

## Additional diagnostics info

```
Creating a PR from 209/plugins to develop...__fire_github_api_request(method=POST, path=/repos/<org_redacted_by_me>/<repo_redacted_by_me>/pulls, token=***, request_body={'head': '209/plugins', 'base': 'develop', 'title': 'feat(web): #209 copy lexical plugins', 'body': '', 'draft': False}): firing a POST request to https://api.github.com/repos/<org_redacted_by_me>/<repo_redacted_by_me>/pulls with a bearer token and request body {"head": "209/plugins", "base": "develop", "title": "feat(web): #209 copy lexical plugins", "body": "", "draft": false}

GitHub API returned 307 HTTP status with error message: Temporary Redirect

Please open an issue regarding this topic under link: https://github.com/VirtusLab/git-machete/issues/new
```

amalota:


amalota:
Result:
image

git_machete/github.py
```diff
@@ -214,6 +214,18 @@ def __fire_github_api_request(method: str, path: str, token: Optional[str], requ
                 f'`{method} {url}` request ended up in 404 response from GitHub. A valid GitHub API token is required.\n'
                 f'Provide a GitHub API token with `repo` access via one of the: {get_github_token_possible_providers()} '
                 'Visit `https://github.com/settings/tokens` to generate a new one.')  # TODO (#164): make dedicated exception here
+        elif err.code == http.HTTPStatus.TEMPORARY_REDIRECT:
+            current_repo_and_org = get_repo_and_org_names_by_id(err.headers['Location'].split('/')[4])
```
PawelLipski: Just in case... https://github.com/VirtusLab/git-machete/issues/597#issuecomment-1195864082 has an example of lower-case `location` header being returned. Is `err.headers[...]` case-sensitive?

git_machete/github.py
```diff
@@ -214,6 +214,18 @@ def __fire_github_api_request(method: str, path: str, token: Optional[str], requ
...
+            new_path = f'/repos/{current_repo_and_org}/{"/".join(path.split("/")[4:])}'
```
PawelLipski: Hmmm how is the `new_path` gonna look like? 🤔 maybe add an example in comment in code

git_machete/github.py
```diff
@@ -214,6 +214,18 @@ def __fire_github_api_request(method: str, path: str, token: Optional[str], requ
...
+            print(fmt(f'<yellow>\nGitHub API returned {err.code} HTTP status with error message: {err.reason}. '
+                      'It looks like the organization or repository name got changed recently and is outdated.\n'
+                      'Inferring current organization or repository... '
+                      f'New organization = {current_repo_and_org.split("/")[0]}, '
+                      f'new repository = {current_repo_and_org.split("/")[1]}.\n'
+                      'You can update your remote repository via: `git remote set-url <repository_name> <new_repository_url>`. </yellow>'),
```
PawelLipski:
```suggestion
                      'You can update your remote repository via: `git remote set-url <remote_name> <new_repository_url>`. </yellow>'),
```

git_machete/github.py
PawelLipski: And/or extract the `"/".join(path.split("/")[4:])` part to a separate variable

git_machete/github.py
```diff
@@ -214,6 +214,18 @@ def __fire_github_api_request(method: str, path: str, token: Optional[str], requ
                 f'`{method} {url}` request ended up in 404 response from GitHub. A valid GitHub API token is required.\n'
                 f'Provide a GitHub API token with `repo` access via one of the: {get_github_token_possible_providers()} '
                 'Visit `https://github.com/settings/tokens` to generate a new one.')  # TODO (#164): make dedicated exception here
+        elif err.code == http.HTTPStatus.TEMPORARY_REDIRECT:
+            current_repo_and_org = get_repo_and_org_names_by_id(err.headers['Location'].split('/')[4])
+            new_path = f'/repos/{current_repo_and_org}/{"/".join(path.split("/")[4:])}'
+            print(fmt(f'<yellow>\nGitHub API returned {err.code} HTTP status with error message: {err.reason}. '
```
PawelLipski: Maybe use `def warn`, we've got something like that in utils

git_machete/github.py
amalota: @PawelLipski Both lower and upper case first letter works, i just chose upper-case.

git_machete/github.py
amalota: @PawelLipski Hmm, warn has red color which for me identifies with the error and since this is not an error (we can handle it) i wouldn't color it with red, wdyt?

PawelLipski: Hmmm actually... maybe change the color in `def warn` to orange? Yellow might be hard to spot on white-background terminals (pls check on your Mac, actually)

git_machete/github.py
```diff
@@ -216,13 +216,15 @@ def __fire_github_api_request(method: str, path: str, token: Optional[str], requ
                 'Visit `https://github.com/settings/tokens` to generate a new one.')  # TODO (#164): make dedicated exception here
         elif err.code == http.HTTPStatus.TEMPORARY_REDIRECT:
             current_repo_and_org = get_repo_and_org_names_by_id(err.headers['Location'].split('/')[4])
-            new_path = f'/repos/{current_repo_and_org}/{"/".join(path.split("/")[4:])}'
+            pulls_or_issues_api_suffix = "/".join(path.split("/")[4:])
```
PawelLipski, line 229: +1, much clearer now, I wouldn't guess that it might be either `pulls` or `issues`...

PawelLipski: Actually, won't it always be `pulls`? 🤔 why can it end up being `issues`? 😯

git_machete/github.py
PawelLipski: Okay, probably some case-insensitive `dict` is underneath... pls add a comment in code about that (preferably pointing to an place in the respective `__getitem__` implementation)

git_machete/github.py
```diff
@@ -216,13 +216,15 @@ def __fire_github_api_request(method: str, path: str, token: Optional[str], requ
...
+            new_path = f'/repos/{current_repo_and_org}/{pulls_or_issues_api_suffix}'
+            # example when creating new PR, new_path='/repos/new_org_name/new_repo_name/pulls'
```
PawelLipski:
```suggestion
            # for example when creating a new PR, new_path='/repos/new_org_name/new_repo_name/pulls'
```

git_machete/github.py
PawelLipski: Awww okay, coz this method is generic right? can be called not just for creating a new PR but for other cases as well?

git_machete/github.py
amalota: @PawelLipski IMO yellow is nicely visible on white background 
image
so i would leave it as it is, wdyt?

git_machete/github.py
amalota: Yes, exactly, it can be used for other cases as well

git_machete/github.py
PawelLipski: Hmmm okay... still, I'd insist making the format of all warnings consistent. I'd rather use `def warn` (possibly changed to get rid of red color as you suggested ☝🏻) here as well.

git_machete/github.py
```diff
@@ -214,6 +214,22 @@ def __fire_github_api_request(method: str, path: str, token: Optional[str], requ
                 f'`{method} {url}` request ended up in 404 response from GitHub. A valid GitHub API token is required.\n'
                 f'Provide a GitHub API token with `repo` access via one of the: {get_github_token_possible_providers()} '
                 'Visit `https://github.com/settings/tokens` to generate a new one.')  # TODO (#164): make dedicated exception here
+        elif err.code == http.HTTPStatus.TEMPORARY_REDIRECT:
+            current_repo_and_org = get_repo_and_org_names_by_id(err.headers['Location'].split('/')[4])
+            # err.headers is a case-insensitive dict of class Message with the `__getitem__` function implemented in
```
PawelLipski:
```suggestion
            # err.headers is a case-insensitive dict of class Message with the `__getitem__` and `get` functions implemented in
```

git_machete/github.py
```diff
@@ -214,6 +214,22 @@ def __fire_github_api_request(method: str, path: str, token: Optional[str], requ
                 f'`{method} {url}` request ended up in 404 response from GitHub. A valid GitHub API token is required.\n'
                 f'Provide a GitHub API token with `repo` access via one of the: {get_github_token_possible_providers()} '
                 'Visit `https://github.com/settings/tokens` to generate a new one.')  # TODO (#164): make dedicated exception here
+        elif err.code == http.HTTPStatus.TEMPORARY_REDIRECT:
+            current_repo_and_org = get_repo_and_org_names_by_id(err.headers['Location'].split('/')[4])
```
PawelLipski: One more thing... to make this thing really future-proof... I'd make an explicit check:
* whether `Location` header is present and non-`None`
* whether the result of `split('/')` has 4-th element

to avoid runtime exception on some unexpected HTTP response. Unlikely, but still

git_machete/github.py
amalota: Alright, i'll update the `warn` in separate PR, here's how it will look like after: 
image
@PawelLipski ok?

PawelLipski: Okay, sure

PawelLipski (approved):

PR #623, opened 2022-08-13
Add dependabot config

### LINKED ISSUE #617: Use specific versions of packages in tox.ini to ensure reproducible builds + set up automatic updates
As for now, most `deps` in tox.ini are defined as:
```tox
deps =
  pytest
  pytest-mock
  pytest-xdist
```

This is convenient, but might lead to non-reproducible builds in the future (as just the latest version of each package is used).

Let's instead use specific version, as in
```tox
deps =
  sphinx==4.4.0
  sphinx-rtd-theme==0.5.2
```

Also, let's check if there's a tool (dependabot?) that is able to open dependency update PRs periodically. Probably it might require all dependencies to be defined in a `*-requirements.txt` file? TBD

PawelLipski:
image

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,12 @@
+# Please see the documentation for all configuration options:
+# https://docs.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+
+version: 2
+updates:
+  - package-ecosystem: "pip"
+    directory: "/"
+    schedule:
+      interval: weekly
+      day: saturday
+      time: "12:45"
+      timezone: Europe/Warsaw


PawelLipski:
Let's see if it works for our case with:
* multiple separate `requirements.*.txt` files and
* Python version markers, like `pytest==7.0.1; python_version=="3.6"`, `pytest==7.1.2; python_version>="3.7"`

PawelLipski:
As a fallback option, a manual setup for https://github.com/simion/pip-upgrader/ can be used (it doesn't support Python version markers at all, but does support multiple requirements.txt files)

Issue #603, opened 2022-07-27
Failed `assert ... == ...` in tests doesn't seem to show the actual value


PawelLipski:
A sample failure (I tried to run the test newly-added on #594 against code without `client.py` changed):

```
================================================================================================================== FAILURES ==================================================================================================================
_______________________________________________________________________________ TestStatus.test_status_when_child_branch_is_pushed_immediately_after_creation ________________________________________________________________________________
[gw0] linux -- Python 3.10.5 /home/plipski/git-machete/.tox/py310/bin/python

self = <tests.test_status.TestStatus object at 0x7f2cbe406e60>, mocker = <pytest_mock.plugin.MockerFixture object at 0x7f2cbe4105e0>

    def test_status_when_child_branch_is_pushed_immediately_after_creation(self, mocker: Any) -> None:
        mocker.patch('git_machete.utils.run_cmd', mock_run_cmd)  # to hide git outputs in tests
    
        (
            self.repo_sandbox.new_branch("master")
            .commit("master")
            .push()
            .new_branch("foo")
            .commit("foo")
            .new_branch("bar")
            .push()
            .commit("bar")
        )
        launch_command('discover', '-y')
        expected_status_output = (
            """
            master
            |
            o-foo (untracked)
              |
              o-bar * (ahead of origin)
            """
        )
>       assert_command(['status'], expected_status_output)

/home/plipski/git-machete/tests/test_status.py:309: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

cmds = ['status'], expected_result = '  master\n  |\n  o-foo (untracked)\n    |\n    o-bar * (ahead of origin)\n', strip_indentation = True, indent = '  '

    def assert_command(cmds: Iterable[str], expected_result: str, strip_indentation: bool = True, indent: str = '  ') -> None:
        expected_result = adapt(expected_result, indent) if strip_indentation else expected_result
>       assert launch_command(*cmds) == expected_result
[...]

PR #632, opened 2022-08-16
Issue #618: Add new subcommand `git machete list childless`

### LINKED ISSUE #618: Implement `git machete list childless`
As suggested by @LeeAlephium on #616.

amalota:


git_machete/docs.py
```diff
@@ -554,6 +554,7 @@
 
         Lists all branches that fall into one of the specified categories:
         * `addable`: all branches (local or remote) than can be added to the definition file,
+        * `childless`: all branches that do not posses children branches,
```
```suggestion
        * `childless`: all branches that do not possess child branches,
```

PawelLipski: (https://virtuslab.github.io/tips/#english/grammar-tips-1 -> point 4)

git_machete/client.py
```diff
@@ -68,6 +68,10 @@ def up_branch(self) -> Dict[LocalBranchShortName, Optional[LocalBranchShortName]
     def up_branch(self, val: Dict[LocalBranchShortName, Optional[LocalBranchShortName]]) -> None:
         self._up_branch = val
 
+    def get_childless_branches(self):
+        parent_branches = self.__down_branches.keys()
+        return [branch for branch in self.managed_branches if branch not in parent_branches]
```
PawelLipski: Nit: or use `def excluding`?

git_machete/client.py
```diff
@@ -68,6 +68,10 @@ def up_branch(self) -> Dict[LocalBranchShortName, Optional[LocalBranchShortName]
     def up_branch(self, val: Dict[LocalBranchShortName, Optional[LocalBranchShortName]]) -> None:
         self._up_branch = val
 
+    def get_childless_branches(self):
+        parent_branches = self.__down_branches.keys()
```
PawelLipski: Also, I'm not 100% sure if it's guaranteed (in general case, esp. after some slide-outs have been performed in the given git-machete execution) that all values in `__down_branches` are non-empty lists. I'd only leave the keys which correspond to non-empty branch lists

tests/mockers.py
```diff
@@ -353,6 +353,8 @@ def launch_command(*args: str) -> str:
 
 def assert_command(cmds: Iterable[str], expected_result: str, strip_indentation: bool = True, indent: str = '  ') -> None:
     expected_result = adapt(expected_result, indent) if strip_indentation else expected_result
+    e = expected_result
+    l = launch_command(*cmds)
     assert launch_command(*cmds) == expected_result
```
PawelLipski: Hmm `assert l == e`?

tests/test_list.py
```diff
@@ -0,0 +1,130 @@
...
+            self.repo_sandbox
+            # Create the remote and sandbox repos, chdir into sandbox repo
+            .new_repo(self.repo_sandbox.remote_path, "--bare")
+            .new_repo(self.repo_sandbox.local_path)
+            .execute(f"git remote add origin {self.repo_sandbox.remote_path}")
+            .execute('git config user.email "tester@test.com"')
+            .execute('git config user.name "Tester Test"')
+        )
+
+    def test_list(self, mocker: Any) -> None:
```
PawelLipski, line 22: :benissimo:

tests/test_list.py
```diff
@@ -0,0 +1,130 @@
...
+                .new_branch("feature_0_0_0")
+                .commit("feature_0_0_0 commit.")
+                .check_out("feature_0")
+                .new_branch("feature_0_1")
+                .commit("feature_0_1 commit.")
+                .check_out("develop")
+                .new_branch("feature_1")
+                .commit("feature_1 commit.")
+        )
+        launch_command("discover", "-y")
```
PawelLipski: Hmmm maybe better just fill up .git/machete file explicitly? It'll make it more clear how the branch layout is supposed to look

tests/test_list.py
```diff
@@ -0,0 +1,130 @@
...
+
+        (
+            self.repo_sandbox.check_out("develop")
+                             .new_branch("feature_2")
+                             .commit("feature_2 commit.")
+        )
+
+        expected_msg = """
```
PawelLipski: Here and in other places
```suggestion
        expected_output = """
```

git_machete/cli.py
```diff
@@ -253,7 +253,7 @@ def create_cli_parser() -> argparse.ArgumentParser:
     list_parser.add_argument(
         'category',
         choices=[
-            'addable', 'managed', 'slidable', 'slidable-after', 'unmanaged',
+            'addable', 'childless', 'managed', 'slidable', 'slidable-after', 'unmanaged',
```
PawelLipski, line 256: Pls also update docs (2 files) and shell completions (3 files)

tests/mockers.py
amalota: forgot to remove this, thx

PawelLipski: Actually... was the reason why you've done it has something to do with https://github.com/VirtusLab/git-machete/issues/603 ? to see the actual vs expected value printed out in case of test failure?

tests/test_list.py
amalota: Hmm, i can just add the expected layout in the comment below, @PawelLipski wdyt?

amalota: like: 
```
        # content of the machete definition file:
        """
        master
        develop
            feature_0
                feature_0_0
                    feature_0_0_0
                feature_0_1
            feature_1
        """
```

PawelLipski: Could be... but it might get outdated with the changes to the test. TBH I'd just overwrite machete file explicitly, no need to kinda-test `discover` in an otherwise non-`discover`-related test

amalota: @PawelLipski I tried overwriting def file but i still had to leave the creation of the branches and the commits so in the end it looked less clear plus it could get outdated as well, i don't see why it would be better

amalota: Hmm, my bad, it would not get updated as the tests will probably fail, will try again

PawelLipski (approved):
Let's ship it then!

Issue #323, opened 2021-10-27
Decide explicitly on `usedevelop = true/false` in each `tox` env


PawelLipski:
I.e. whether `git-machete` package should be installed in the venv used by `tox`.

PR #638, opened 2022-08-19
Issue #211: Generate command line documentation based on .rst files.

### LINKED ISSUE #211: Create one common place to store commands and help topics documentation
After delivery of #202 there will be two unsynchronized places where documentation of `commands` and `help topics` is stored - `docs/cli_help/*` and `git_machete/docs.py`. To avoid duplication and problems with one storage being updated and the other one not it would be good to choose one place to store such information and than via some scripts generate documentation in a particular formats and put them in correct locations.

Things to consider while preparing script/s to generate .rst and docs help:
1. What tags are needed to break the line in cli help and what in .rst?
2. How to mark a code block?
3. How to color text?
4. How/if mark <branch>?
5. Should there be a separate general storage (that is not directly used, but is a base for all other) and from this one we would generate documentation that is used in other places?

amalota:


git_machete/long_docs.py
```diff
@@ -0,0 +1,934 @@
+from typing import Dict
```
PawelLipski: Actually, a high-level question: since these docs are auto-generated, should they be tracked under git at all? I'd either expect:
1. that these files are git-ignored, only created once `tox -e docs` runs OR
2. that these files are tracked as under git, but then CI only checks if they're identical to what you'd get from the existing .rst docs (e.g. just checks if they're kept up to date with .rst)

Actually, I'd even say that the latter option will give us more insight/control over what's kept in these files... although it's rather unorthodox (typically what I see in various ecosystems is rather (1.))

amalota: Alright, lets go with second option

PawelLipski: Yeah, we've agreed to (2.) coz it's more future-proof (what if some yet-unhandled formatting in RST ends up looking weird in docs.py?)

.circleci/config.yml
```diff
@@ -105,7 +105,7 @@ jobs:
       - codecov/upload:
           file: coverage.xml
 
-  "sphinx docs and python 3_10 git 2_33_1":
+  "sphinx and command line docs and python 3_10 git 2_33_1":
```
PawelLipski: Maybe just
```suggestion
  "docs and python 3_10 git 2_33_1":
```
for brevity (so that these are easier to display by GitHub)

git_machete/short_docs.py
```diff
@@ -0,0 +1,31 @@
+from typing import Dict
+
```
PawelLipski: Actually, would be nice if (at least for the time of review), `short_docs.py` and `long_docs.py` are located into a single file, called just `docs.py` as before this change... this way we'll clearly see on PR review what's changed

docs/generate_docs.py
```diff
@@ -0,0 +1,169 @@
...
+
+def skip_prefix_new_lines(txt: str) -> str:
+    txt = re.sub(r'\A[\n]+', '', txt)
+    return txt
+
+
+if __name__ == '__main__':
+    output_docs_path = 'git_machete/long_docs.py'
+    docs_source_path = 'docs/source'
+    output_text = 'from typing import Dict\n\nlong_docs: Dict[str, str] = {\n'
```
PawelLipski: Actually, while I remember: pls place a comment in the autogenerated file(s) that they're NOT supposed to be edited directly, but instead regenerated via `tox -e docs`

docs/generate_docs.py
```diff
@@ -0,0 +1,169 @@
...
+
+        rst = skip_or_replace_unparseable_directives(rst)
+        rst = resolve_includes(rst=rst, docs_source_path=docs_source_path)
+        html = rst2html(rst)['body']
+        # print(html)
+        # print('\n\n\n)
+        plain_text = html2txt(html)
+        plain_text = skip_prefix_new_lines(plain_text)
+        # print(plain_text)
+        output_text += f'    "{command}": """\n' + indent(plain_text, '        ') + '\n   """,\n'
```
PawelLipski: Maybe even just build a dict here, and add its `repr` into `output_text`? instead of appending pieces of Python code manually

amalota: @PawelLipski I tried but there is no easy way to write the dict as string in the readable format so i would stick with how it is atm

ci/checks/enforce-docs-up-to-date.sh
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env bash
+
+pip install bs4
```
PawelLipski: Pls extract to a dedicated requirements.*.txt with fixed versions (as I've done with all other dependencies recently)

ci/checks/run-all-checks.sh
```diff
@@ -24,6 +24,7 @@ enforce-newline-at-eof.sh
 enforce-release-notes-up-to-date.sh
 enforce-shell-scripts-pass-shellcheck.sh
 enforce-version-bumped.sh
+enforce-docs-up-to-date.sh
```
PawelLipski: Pls note that this script is supposed to be wired up as pre-commit hook (see CONTRIBUTING.md)... I'm not sure if running `pip install` on each commit is safe/reasonable (as the packages might end up getting installed in a venv pointed by the current shell etc.)

docs/source/cli_help/config.rst
```diff
@@ -11,25 +11,8 @@ Note: ``config`` is not a command as such, just a help topic (there is no ``git
     * ``machete.github.{remote,organization,repository}``:
 
         When executing ``git machete github <subcommand>`` command, GitHub API server URL will be inferred from ``git remote``.
-        You can override this by setting the following local git config keys:
```
PawelLipski, line 14: Wow this one survived the recent refactors ;)

ci/checks/enforce-docs-up-to-date.sh
```diff
@@ -0,0 +1,20 @@
...
+
+pwd
+ls
+previous_pwd=$(pwd)
+#cd ../../
+
+current_docs=$(cat git_machete/docs.py)
+generated_docs=$(python docs/generate_docs.py)
+if [[ $current_docs != "$generated_docs" ]]; then
+  cd $previous_pwd || return
```
PawelLipski: Hmm are the `cd`s needed? Current directory is confined to the given bash process, and in `ci/checks/run-all-checks.sh` we execute each script (including this one) in a separate shell process

git_machete/docs.py
```diff
@@ -728,128 +848,146 @@
...
-          - by commit equivalency (default), or by strict detection of merge commits (if `--no-detect-squash-merges` passed),
-          - if so, asks the user whether to slide out the branch from the dependency tree (typically branches are no longer needed after they're merged);
-        * otherwise, if the branch has a <b><red>red</red></b> or <b><yellow>yellow</yellow></b> edge to its parent/upstream (see `status`):
-          - asks the user whether to rebase (default) or merge (if `--merge` passed) the branch onto into its upstream branch - equivalent to `git machete update` with no `--fork-point` option passed;
-
+        * detects if the branch is merged (grey edge) to its parent (aka upstream):
+           - by commit equivalency (default), or by strict detection of merge commits (if `--no-detect-squash-merges` passed),
+           - if so, asks the user whether to slide out the branch from the dependency tree (typically branches are no longer needed after they're merged);
+        * otherwise, if the branch has a [31mred[0m or [33myellow[0m edge to its parent/upstream (see status):
```
PawelLipski: Whoops

ci/checks/run-all-checks.sh
PawelLipski: Actually... it looks to me that this task better fits into `tox` model (with deps installed into a fresh virtual env) than into a vanilla shell script as now

git_machete/docs.py
amalota: I mean this was to be expected since there were no colors in the cmd line docs before and we decided to add them

PawelLipski: Oh! okay, makes sense, I see

ci/checks/run-all-checks.sh
amalota: Yes, great idea, i think it should work well within tox

ci/checks/enforce-docs-up-to-date.sh
amalota: Yes, you are right, the 'cd's are indeed unnecessary

ci/tox/.env-sample
```diff
@@ -3,6 +3,7 @@ GIT_VERSION=2.25.0  # note that generally the older the git version, the faster
 PYTHON_VERSION=3.8
 
 # Vars for `docker-compose run`
-BUILD_DOCS=false
+BUILD_SPHINX_DOCS=false
+CHECK_CMD_DOCS_UP_TO_DATE=false
```
PawelLipski: I'd rather call them `PY_DOCS` or `PYTHON_DOCS`, `CMD` is unclear in this context

diff.txt
```diff
@@ -0,0 +1,4 @@
+1c1
```
PawelLipski: To be removed right?

requirements.cmd_docs.txt
```diff
@@ -0,0 +1,3 @@
+docutils==0.17.1
```
PawelLipski: Same here, rename the file for consistency

tox.ini
```diff
@@ -53,14 +53,32 @@ commands =
...
+[testenv:build_cmd_docs]
+description = "Build command line documentation"
+usedevelop=true
+deps =
+  -r{toxinidir}/requirements.cmd_docs.txt
+commands =
+  python docs/generate_docs.py
+
+
+[testenv:check_docs]
```
PawelLipski: Looks that the convention is to use `-` rather than `_` in env names

tox.ini
```diff
@@ -53,14 +53,32 @@ commands =
...
+description = "Check if command line documentation is up to date with sphinx documentation"
+usedevelop=true
+allowlist_externals = /usr/local/bin/bash
```
PawelLipski: Hmm `bash` typically lives under `/usr/bin/bash`; `/usr/local/bin/` is for stuff that doesn't come from official package repos: https://unix.stackexchange.com/a/4187

docs/source/cli_help/config.rst
```diff
@@ -71,15 +54,12 @@ Note: ``config`` is not a command as such, just a help topic (there is no ``git
...
 
-    * ``GIT_MACHETE_EDITOR``
-
-        Name of the editor used by ``git machete e[dit]``, example: ``vim`` or ``nano``.
-
-    * ``GIT_MACHETE_REBASE_OPTS``
-
-        Used to pass extra options to the underlying ``git rebase`` invocation (called by the executed command, such as: ``reapply``, ``slide-out``, ``traverse``, ``update``)
-        Example: ``GIT_MACHETE_REBASE_OPTS="--keep-empty --rebase-merges" git machete update``.
+``GIT_MACHETE_EDITOR``
```
PawelLipski, line 31: Is the formatting consistent with other .rst files where env vars are documented?

git_machete/short_docs.py
PawelLipski: (resolved?)

tox.ini
```diff
@@ -53,14 +53,32 @@ commands =
...
+description = "Build sphinx documentation"
 usedevelop=true
 deps =
-  -r{toxinidir}/requirements.docs.txt
+  -r{toxinidir}/requirements.sphinx_docs.txt
 commands =
   sphinx-build -W --keep-going -b html docs/source docs/html
 
+[testenv:build_cmd_docs]
+description = "Build command line documentation"
```
PawelLipski: Yeah, both Sphinx and Python docs are, in some sense, a "command line documentation"... I mean, yeah, only the Python docs are **available via command line**, but both **are documentation of command line interface**. Using the "Python docs" name should be less confusing IMO

tox.ini
```diff
@@ -53,14 +53,32 @@ commands =
...
+usedevelop=true
+deps =
+  -r{toxinidir}/requirements.cmd_docs.txt
+commands =
+  python docs/generate_docs.py
+
+
+[testenv:check_docs]
+description = "Check if command line documentation is up to date with sphinx documentation"
+usedevelop=true
```
PawelLipski, line 77: I don't think `true` (i.e. installing the package into the venv) is needed here... there's an entire task to smooth that out, #323

docs/enforce-docs-up-to-date.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
+
+set -e -o pipefail -u
+
+current_docs=$(cat git_machete/docs.py)
+generated_docs=$(python docs/generate_docs.py dont_save)
```
PawelLipski: I'd rather make the script print to stdout by default; saving can be done by a simple `> git_machete/docs.py`

docs/enforce-docs-up-to-date.sh
```diff
@@ -0,0 +1,11 @@
...
+
+if [[ "$current_docs" != "$generated_docs" ]]; then
```
PawelLipski, line 8: What's the limit for env var contents (on Mac/on Linuxes/in Docker)? pls check if we're ever likely to hit that

PawelLipski: Alternatively, just use https://virtuslab.github.io/tips/#shell/output-comparing

docs/enforce-docs-up-to-date.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
+
+set -e -o pipefail -u
+
+current_docs=$(cat git_machete/docs.py)
```
PawelLipski: Nit:
```suggestion
current_docs=$(<git_machete/docs.py)
```

tox.ini
```diff
@@ -53,14 +53,32 @@ commands =
   # Saves a report to coverage.xml (uploaded to Codecov from CI, as Codecov apparently does not accept .coverage binary files)
   coverage xml
 
-[testenv:docs]
-description = "Build documentation"
+[testenv:build_sphinx_docs]
```
PawelLipski: Pls keep `envlist` up to date

tox.ini
amalota: @PawelLipski I tried using `usr/bin/bash` but got this warning: 
```
WARNING: test command found but not installed in testenv
  cmd: /usr/local/bin/bash
  env: /Users/amalota/Projects/git-machete/.tox/check-docs
Maybe you forgot to specify a dependency? See also the allowlist_externals envconfig setting.

DEPRECATION WARNING: this will be an error in tox 4 and above!
```

docs/source/cli_help/config.rst
amalota: Yes, now it is

tox.ini
amalota: @PawelLipski Checking if documentation is up-to-date uses docs generating script which in turn uses git_machete package so it has to stay.

docs/enforce-docs-up-to-date.sh
amalota: @PawelLipski Sure, it's done but i had to create separate bash file just for saving because tox doesn't support output redirections.

docs/enforce-docs-up-to-date.sh
amalota: @PawelLipski Hmm, it looks like it's not that simple since it depends on how you set the env var value: https://devblogs.microsoft.com/oldnewthing/20100203-00/?p=15083 but per internet:

Linux: 128KiB -> https://unix.stackexchange.com/questions/336934/raise-128kib-limit-on-environment-variables-in-linux but some guy had env var with the size of 256 MB

MacOs: based on this post it is difficult to say and it also depends on the system resources > https://aplawrence.com/Unixart/variable_limits.html but the guy managed to get up to hundreds of MB

I couldn't find anything on docker limits, to sum up it's impossible to answer but it could be hundreds of MBs.

tox.ini
PawelLipski: Hmmm is it possible to just add `bash`, without specifying the exact path? As the path might differ b/w systems/machines

docs/enforce-docs-up-to-date.sh
PawelLipski: Hmmm does it support piping to `tee`? Like `python ... | tee ...` (which is kinda equivalent to `python ... > ...`)

docs/enforce-docs-up-to-date.sh
PawelLipski: Okay, looks safe generally then

tox.ini
PawelLipski: Right... pls add a comment here in tox.ini then so that we don't need to re-research when doing #323

tox.ini
amalota: Oh yeah, i didn't think about it, it is indeed possible.

docs/enforce-docs-up-to-date.sh
amalota: @PawelLipski It does support `tee`, the issue was raised here https://github.com/tox-dev/tox/issues/73, the proposed workaround -> `commands =
    bash -c \'pylint --rcfile=.pylintrc implib2 |tee pylint.out\'`
unfortunately doesn't work but `bash -c "python docs/generate_py_docs.py |tee git_machete/docs.py"` works although it not only saves the the output but also prints it, is that acceptable?

amalota: @PawelLipski Oh yeah, i got it, `bash -c "python docs/generate_py_docs.py > git_machete/docs.py"` redirects without printing, "ladies and gentlemen, we got him" :D

PawelLipski: Haha yeah, that's what I thought as well ;)

docs/enforce-docs-up-to-date.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
+
+set -e -o pipefail -u
+
+current_docs=$(<git_machete/docs.py)
+generated_docs=$(python docs/generate_py_docs.py)
+
+if [[ "$current_docs" != "$generated_docs" ]]; then
+  echo "Command line docs are not up-to-date. Please regenerate docs via tox -e py_docs"
```
PawelLipski: `py-docs` or `py_docs`?

git_machete/short_docs.py
```diff
@@ -0,0 +1,31 @@
...
+short_docs: Dict[str, str] = {
```
PawelLipski: Okay... so in the target state of this repo, we'll have both short_docs.py and docs.py, the latter with**out** `short_docs` dict? in fact I'm for just having single docs.py, and removing short_docs.py completely. Single file is easier to manage

PawelLipski: Or actually... are these short docs generated out of .rst? or to be filled up manually

requirements.py_docs.txt
```diff
@@ -0,0 +1,3 @@
+docutils==0.17.1
```
PawelLipski: Nit: use `...py-docs.txt`, for consistency with tox env name; same for `sphinix-docs`

tox.ini
```diff
@@ -1,5 +1,5 @@
 [tox]
-envlist = pep8,mypy-py{36,37,38,39,310},py{36,37,38,39,310},coverage,docs
+envlist = pep8,mypy-py{36,37,38,39,310},py{36,37,38,39,310},coverage,sphinx-docs,py-docs,check-py-docs
```
PawelLipski: It doesn't make much sense to include both `py-docs` and `check-py-docs` in `envlist` (i.e. the stuff that's executed on simple `tox` invocation with no `-e`)... I'd leave just `py-docs`

git_machete/short_docs.py
amalota: So i created separate file short_docs to be modified manually and then used in the docs.py for the review process but ultimately i would not copy short_docs.py into docs.py as it is unnecessary, @PawelLipski ok?

docs/enforce-docs-up-to-date.sh
amalota: `py-docs`

tox.ini
amalota: @PawelLipski Hmm, i think it has to stay since modified Ci for docs executes both `check-py-docs` and `sphinx-docs`

git_machete/short_docs.py
amalota: For clarification i can even rename the files to: short_docs_manually.py and docs_generated.py or sth like that, @PawelLipski wdyt?

PawelLipski: +1, I'd make it as clear as possible what's generated and what's manual

PawelLipski: Plus a comment at the beginning of each file to say whether it should or should not be modified manually

tox.ini
PawelLipski: But does an env need to be included in `envlist` to be executable? Doesn't `envlist` just control the behavior of no-argument `tox` (which is never called this way in CI, IIRC)?

amalota: Yeah, you are right, env doesn't need to be included in envlist to be executable

tox.ini
```diff
@@ -53,14 +53,34 @@ commands =
...
 usedevelop=true
 deps =
-  -r{toxinidir}/requirements.docs.txt
+  -r{toxinidir}/requirements.sphinx-docs.txt
 commands =
   sphinx-build -W --keep-going -b html docs/source docs/html
 
+[testenv:py-docs]
+description = "Build Python documentation"
+# The generation of the pythons docs uses git_machete package
```
PawelLipski:
```suggestion
# The generation of the python docs uses git_machete package
```

tox.ini
```diff
@@ -53,14 +53,34 @@ commands =
...
+usedevelop=true
+allowlist_externals = bash
+deps =
+  -r{toxinidir}/requirements.py-docs.txt
+commands =
+  bash -c "python docs/generate_py_docs.py > git_machete/generated_docs.py"
+
+[testenv:check-py-docs]
+description = "Check if Python documentation is up to date with Sphinx documentation"
+# The generation of the pythons docs uses git_machete package
```
PawelLipski:
```suggestion
# The generation of the python docs uses git_machete package
```

docs/generate_py_docs.py
```diff
@@ -0,0 +1,174 @@
+import re
+import os
+from textwrap import dedent, indent
```
PawelLipski: Nit: sort imports in abc order

docs/generate_py_docs.py
```diff
@@ -0,0 +1,174 @@
...
+    for tag in html_elements.select('td'):
+        # remove non-breaking spaces
+        if tag.text == u'\xa0':
+            tag.decompose()
+        else:
+            # add new line and indent for the cells inside a table
+            if 'class' not in tag.attrs:
+                tag.insert_before('\n      ')
+
+    # substitute double apostrophe with a single apostrophe
```
PawelLipski: +1 for the comments on what happens in each step

PawelLipski: Although... here it looks it doesn't describe what happens in this step ;D

docs/generate_py_docs.py
```diff
@@ -0,0 +1,174 @@
...
+    for tag in html_elements.select('style'):
+        tag.decompose()
+
+    for tag in html_elements.select('td'):
+        # remove non-breaking spaces
+        if tag.text == u'\xa0':
+            tag.decompose()
+        else:
+            # add new line and indent for the cells inside a table
+            if 'class' not in tag.attrs:
```
PawelLipski, line 36: Hmm is having `class` attribute equivalent to being in a table in Sphinx-generated HTML?

git_machete/handmade_short_docs.py
```diff
@@ -0,0 +1,35 @@
+from typing import Dict
+
+# ---------------------------------------------------------------------------------------------------------
+# Warning: This file is supposed to be modified manually
```
PawelLipski: Not a warning ;) just
```suggestion
# This file is supposed to be modified manually
```
Warnings make sense where the file is NOT supposed to be edited manually

git_machete/generated_docs.py
```diff
@@ -0,0 +1,938 @@
...
+   """,
+    "advance": """
+        <b>Usage:</b><b>
+          git machete advance [-y|--yes]</b>
+
+        Fast forwards (as in `git merge --ff-only`) the current branch `C` to match its downstream `D`, pushes `C`
+        and subsequently slides out `D`. All three steps require manual confirmation unless `-y/--yes` is provided.
+
+        The downstream `D` is selected according to the following criteria:
+        * if `C` has exactly one downstream (child) branch `d` connected with a [32mgreen edge[0m (see status for definition) to `C` or is overridden, then `d` is selected as `D`,
```
PawelLipski: Hmmm there's a major inconsistency: while bold is kept in this file as `<b>` tags, colors are rendered directly with ANSI escape codes. Is there a strong reason for that? If not, I'd make it all consistent, TBD in which direction (all ANSI codes or no ANSI codes)

Also... what if the help is printed out to a non-terminal (see https://github.com/VirtusLab/git-machete/issues/595)? do ANSI codes for colors stay as they are? 🤔

docs/source/cli_help/status.rst
```diff
@@ -80,28 +81,7 @@ With ``--color=always``, git machete always emits colors and with ``--color=auto
 
 --no-detect-squash-merges         Only consider *strict* (fast-forward or 2-parent) merges, rather than rebase/squash merges, when detecting if a branch is merged into its upstream (parent).
 
- .. note::
-
-    To make it easier to select branch name from the ``status`` output on certain terminals
-    (e.g. `Alacritty <https://github.com/alacritty/alacritty>`_), you can add an extra
-    space between ``└─`` and ``branch name`` by setting ``git config machete.status.extraSpaceBeforeBranchName true``.
```
PawelLipski, line 87: Nit: maybe I'd keep the elbow un-quoted:

image

this underline looks weird in terminal

docs/source/cli_help/status.rst
```diff
@@ -80,28 +81,7 @@ With ``--color=always``, git machete always emits colors and with ``--color=auto
 
 --no-detect-squash-merges         Only consider *strict* (fast-forward or 2-parent) merges, rather than rebase/squash merges, when detecting if a branch is merged into its upstream (parent).
 
- .. note::
-
-    To make it easier to select branch name from the ``status`` output on certain terminals
-    (e.g. `Alacritty <https://github.com/alacritty/alacritty>`_), you can add an extra
```
PawelLipski, line 86: Pls open a task for automagically selecting line lengths (might be non-trivial tho, not a high priority definitely)

docs/generate_py_docs.py
amalota: No, it's just that these <td> tags without the 'class' attribute need indent

docs/generate_py_docs.py
amalota: Hmm, which line are you referring to? @PawelLipski

git_machete/generated_docs.py
amalota: @PawelLipski Right, great catch, so since most of the formats are based on tags except the colors and also it is safer as it doesn't face issues with terminal vs non-terminal dilemma i will stick with tags and NO ANSI codes at all.

docs/generate_py_docs.py
PawelLipski: Hmm as the piece of code below this comment:
```python
    # substitute double apostrophe with a single apostrophe
    for tag in html_elements.select('tt'):
        tag.insert_before('`')
        tag.insert_after('`')
```
seems to insert **backticks** around preformatted text (`tt`), nothing to do with apostrophes 🤔

PawelLipski: Oh, you refer to double **backticks** from .rst being effectively replaced with single **backticks** in the generated python hmmm

docs/source/cli_help/status.rst
amalota: https://github.com/VirtusLab/git-machete/issues/652

docs/generate_py_docs.py
amalota: @PawelLipski yeah, exactly that, i'll replace `apostrophes` with `backticks` in the comment to make it clearer

git_machete/generated_docs.py
```diff
@@ -0,0 +1,938 @@
...
+    "config": """
+        Documentation about available `git machete` config keys and environment variables that change the command's default behavior.
+
+        Note: `config` is not a command as such, just a help topic (there is no `git machete config` command).
+
+        <b>Config keys:</b>
+           `machete.github.{remote,organization,repository}`:
+               When executing `git machete github <subcommand>` command, GitHub API server URL will be inferred from `git remote`.
+
+              Note
```
PawelLipski: Such loose _Note_-s look somewhat weird:

image

I'd rather get rid of them in Python docs

git_machete/generated_docs.py
```diff
@@ -0,0 +1,938 @@
...
+              GitHub API token.
+   """,
+    "config": """
+        Documentation about available `git machete` config keys and environment variables that change the command's default behavior.
+
+        Note: `config` is not a command as such, just a help topic (there is no `git machete config` command).
+
+        <b>Config keys:</b>
+           `machete.github.{remote,organization,repository}`:
+               When executing `git machete github <subcommand>` command, GitHub API server URL will be inferred from `git remote`.
```
PawelLipski: Nit: extra space of indent... is it intentional?

git_machete/generated_docs.py
```diff
@@ -0,0 +1,938 @@
...
+              To do this, run `git config --local --edit` and add the following section:
+
+              [machete "github"]
+                  organization = <organization_name>
+                  repository = <repo_name>
+                  remote = <remote_name>
+
+           `machete.overrideForkPoint.<branch>.{to,whileDescendantOf}`:
+              Executing `git machete fork-point --override-to=<revision> [<branch>]` sets up a fork point override for <branch>.
+              The override data is stored under `machete.overrideForkPoint.<branch>.to` and `machete.overrideForkPoint.<branch>.whileDescendantOf` git config keys.
```
PawelLipski: A newline before the next config key would be useful

PawelLipski: Actually, a new line right after the config key (and before its description) would also be _pog_

git_machete/generated_docs.py
```diff
@@ -0,0 +1,938 @@
...
+           `machete.status.extraSpaceBeforeBranchName`:
+
+              Note
+
+              To make it easier to select branch name from the `status` output on certain terminals
+              (e.g. Alacritty), you can add an extra
+              space between `└─` and `branch name` by setting `git config machete.status.extraSpaceBeforeBranchName true`.
```
PawelLipski: Whoops underlined elbow, pls sort all of them out

PawelLipski: (generally, all the comments to the specific format in generated_docs.py are obviously comments regarding the entire algorithm of generating docs out of .rst, not the formatting for a specific command ;) )

git_machete/generated_docs.py
```diff
@@ -0,0 +1,938 @@
...
+          master
+          |
+          m-develop *
+            |
+            | Enable adding remote branch in the manner similar to git checkout
+            o-feature/add-from-remote
+            | |
+            | | Add support and sample for machete-post-slide-out hook
+            | o-feature/post-slide-out-hook
+            |
```
PawelLipski: Actually.. while we're here, I'd simplify this example. Let's just remove `chore/v3` and `feature/types` completely from both trees, they only obfuscate the point while adding little value

tox.ini
```diff
@@ -32,7 +32,7 @@ commands = flake8
 ignore = W504
 show-source = True
 max-line-length = 140
-exclude = ./.*,build,dist,*egg,venv, git_machete/docs.py
+exclude = ./.*,build,dist,*egg,venv, git_machete/handmade_short_docs.py, git_machete/generated_docs.py
```
PawelLipski: `handmade_short_docs` excluded due to line length right?

docs/generate_py_docs.py
```diff
@@ -0,0 +1,170 @@
...
+    # add new line before included NOTE class of rst documentation, example: github_api_access.rst
+    for tag in html_elements.select('div'):
+        if 'class' in tag.attrs:
+            if 'admonition note' in ' '.join(tag.attrs['class']):
+                tag.insert_before('\n')
+
+    # format code examples
+    for tag in html_elements.select('pre'):
+        if 'class' in tag.attrs:
+            if ' '.join(tag.attrs['class']) in ['code shell literal-block', 'code awk literal-block']:
```
PawelLipski: Would it be possible not to rely on specific languages like `shell` or `awk` (as they were selected somewhat randomly IIRC), but apply the transformation to all `pre`s with `code literal-block` classes?

docs/generate_py_docs.py
```diff
@@ -0,0 +1,170 @@
...
+                tag.string = '<b>' + indent(tag.text.rstrip('\n'), '  ') + '</b>'
+            elif 'literal-block' in ' '.join(tag.attrs['class']):
+                tag.string = '\n<dim>' + indent(tag.text, '  ') + '</dim>'
+
+    # substitute color classes with respective ANSI codes
```
PawelLipski: Outdated ☹️

docs/generate_py_docs.py
```diff
@@ -0,0 +1,170 @@
...
+    for tag in html_elements.select('span'):
+        if 'class' in tag.attrs:
+            color = ' '.join(tag.attrs['class']).replace('gray', 'dim')
+            if color in ['green', 'yellow', 'red', 'dim']:
```
PawelLipski: Actually... pls add a test for `help` that would check `help` will not print any ANSI escape sequence in non-terminal mode (should be enough to just check for `\033` character)

docs/generate_py_docs.py
```diff
@@ -0,0 +1,170 @@
...
+                tag.insert_before(f'<{color}>')
+                tag.insert_after(f'</{color}>')
+
+    # build plain text output out of the previously formatted string elements
```
PawelLipski: Hmmm it's not clear to me what happens here... what would happen **without** this step in place?

tox.ini
amalota: @PawelLipski Yes

docs/generate_py_docs.py
amalota: Hmm, without this step there won't be any output docs text because previous steps just modified the html docs but they did not build actual plain text docs, maybe:

```
# build python docs string out of the previously formatted html tags
```
is better?

PawelLipski: Aww okay 🤦🏻 yes, pls change

docs/generate_py_docs.py
amalota: Yes, will do

git_machete/generated_docs.py
amalota: Yes

git_machete/generated_docs.py
amalota: @PawelLipski Okay, so adding the new line before the next config key can be done but not before its description.

docs/source/cli_help/advance.rst
```diff
@@ -40,10 +40,6 @@ As an example, if ``git machete status --color=never --list-commits`` is as foll
       |
       | Remove support for Python 2
       | Remove support for Python 2 --- 1st round of fixes
```
PawelLipski: Whoops these commits (and the downward edge to the left) also to be removed

docs/source/cli_help/advance.rst
```diff
@@ -60,10 +56,6 @@ After ``advance`` completes, ``status`` will show:
       |
```
PawelLipski: Same here, these commits to be removed

git_machete/generated_docs.py
```diff
@@ -0,0 +1,944 @@
...
+          git machete advance [-y|--yes]</b>
+
+        Fast forwards (as in `git merge --ff-only`) the current branch `C` to match its downstream `D`, pushes `C`
+        and subsequently slides out `D`. All three steps require manual confirmation unless `-y/--yes` is provided.
+
+        The downstream `D` is selected according to the following criteria:
+        * if `C` has exactly one downstream (child) branch `d` connected with a <green>green edge</green> (see status for definition) to `C` or is overridden, then `d` is selected as `D`,
+        * if `C` has no downstream branches connected with a <green>green edge</green> to `C`, then `advance` fails,
+        * if `C` has more than one downstream branch connected with a <green>green edge</green> to `C`,
+        then user is asked to pick the branch to fast-forward merge into (similarly to what happens in `git machete go down`). If `--yes` is specified, then `advance` fails.
```
PawelLipski: Whoops this line should be indented so that it's clear that it's continuation of the 3rd case (_if `C` has more than one downstream branch connected with a <green>green edge</green> to `C`,_) 🤔 is it doable easily?

git_machete/generated_docs.py
```diff
@@ -0,0 +1,944 @@
...
+   """,
+    "advance": """
+        <b>Usage:</b><b>
+          git machete advance [-y|--yes]</b>
+
+        Fast forwards (as in `git merge --ff-only`) the current branch `C` to match its downstream `D`, pushes `C`
+        and subsequently slides out `D`. All three steps require manual confirmation unless `-y/--yes` is provided.
+
+        The downstream `D` is selected according to the following criteria:
+        * if `C` has exactly one downstream (child) branch `d` connected with a <green>green edge</green> (see status for definition) to `C` or is overridden, then `d` is selected as `D`,
```
PawelLipski: I'd an extra level of indent before bullet points, like:
```suggestion
          * if `C` has exactly one downstream (child) branch `d` connected with a <green>green edge</green> (see status for definition) to `C` or is overridden, then `d` is selected as `D`,
```

git_machete/generated_docs.py
```diff
@@ -0,0 +1,944 @@
...
+        <b>Options:</b>
+           <b>-c</b>, <b>--checkout-my-github-prs</b>
+              Checkout your open PRs into local branches.
+           <b>-y</b>, <b>--yes</b>
+              Don't ask for confirmation when deleting branches from git.
+
+        <b>Environment variables:</b>
+           `GITHUB_TOKEN`
+              GitHub API token.
+
```
PawelLipski: 2 newlines at the end... can be just 1 left?

PawelLipski: Here and in other help topics (probably wherever `Environment variables` are at the end?)

git_machete/generated_docs.py
```diff
@@ -0,0 +1,944 @@
...
+        Documentation about available `git machete` config keys and environment variables that change the command's default behavior.
+
+        Note: `config` is not a command as such, just a help topic (there is no `git machete config` command).
+
+        <b>Config keys:</b>
+           `machete.github.{remote,organization,repository}`:
+ 
+              When executing `git machete github <subcommand>` command, GitHub API server URL will be inferred from `git remote`.
+
+              GitHub API server URL will be inferred from `git remote`.
```
PawelLipski: Whoops repeated sentence

git_machete/generated_docs.py
```diff
@@ -0,0 +1,944 @@
...
+        <b>Options:</b>
+           <b>-y</b>, <b>--yes</b>
+              Don't ask for confirmation.
+   """,
+    "diff": """
+        <b>Usage:</b><b>
+          git machete d[iff] [-s|--stat] [<branch>]</b>
+
+        Runs `git diff` of the given branch tip against its fork point or, if none specified, of the current working tree against the fork point of the currently checked out branch.
+        See fork-point for more details on meaning of the fork point.
```
PawelLipski:
```suggestion
        See help for `fork-point` for more details on meaning of the fork point.
```

PawelLipski: (here and in other places, like `delete-unmanaged`)

git_machete/generated_docs.py
```diff
@@ -0,0 +1,944 @@
...
+        <b>Usage:</b><b>
+          git machete file</b>
+
+        Outputs the absolute path of machete definition file.
+        The file is always called `machete` and is located in the git directory of the project.
+
+        Three cases are possible:
+        * if `git machete` is executed from a regular working directory (not a worktree or submodule), the file is located under `.git/machete`,
+        * if `git machete` is executed from a <b>worktree</b>, the file path depends on the `machete.worktree.useTopLevelMacheteFile` config key value:
+           - if `machete.worktree.useTopLevelMacheteFile` is true (default), the file is located under `.git/machete`
```
PawelLipski: AFAICS the indent is sometimes 2 spaces (like after `Usage:`), and sometimes 3 spaces (like here)... pls make consistent (best to use a single constant in `generate_py_docs.py` like `INDENT_WIDTH`)

git_machete/generated_docs.py
amalota: @PawelLipski Unfortunately it's not easily doable, i already spent some time on this issue but i'll give it another shot

git_machete/generated_docs.py
```diff
@@ -0,0 +1,944 @@
...
+              hotfix/receipt-trigger PR #236
+        </dim>
+
+        In the above example `develop` and `master` are roots of the tree of branch dependencies.
+        Branches `adjust-reads-prec`, `edit-margin-not-allowed` and `grep-errors-script` are direct downstream branches for `develop`.
+        `block-cancel-order` is a downstream branch of `adjust-reads-prec`, `change-table` is a downstream branch of `block-cancel-order` and so on.
+
+        Every branch name can be followed (after a single space as a delimiter) by a custom annotation - a PR number in the above example.
+        The annotations don't influence the way `git machete` operates other than that they are displayed in the output of the `status` command.
+        Also see anno command.
```
PawelLipski:
```suggestion
        Also see help for `anno` command.
```

git_machete/generated_docs.py
```diff
@@ -0,0 +1,944 @@
...
+        Equivalent to `git machete anno --sync-github-prs`.
+
+        `checkout-prs [--all | --by=<github-login> | --mine | <PR-number-1> ... <PR-number-N>]`:
+
+        Check out the head branch of the given pull requests (specified by numbers or by a flag),
+        also traverse chain of pull requests upwards, adding branches one by one to git-machete and check them out locally.
+        Once the specified pull requests are checked out locally, annotate local branches with corresponding pull request numbers.
+        If only one PR has been checked out, then switch the local repository's HEAD to its head branch.
+
+        <b>Options:</b>
```
PawelLipski: Huh would it be doable to indent the `Options` for subcommands? or best, indent the entire description for each subcommand

PawelLipski: As it's pretty unclear now where one subcommand ends and another begins ☹️

git_machete/generated_docs.py
```diff
@@ -0,0 +1,944 @@
...
+        * what is going to be the new base for the rebased commits,
+        * what is the fork point - the place where the rebased history diverges from the upstream history,
+        * what branch is rebased.
+
+        If the hook returns a non-zero exit code, an error is raised and the entire rebase is aborted.
+
+        Note: this hook is independent from git's standard `pre-rebase hook`.
+        If machete-pre-rebase returns zero, the execution flow continues to `git rebase`, which may also run `pre-rebase hook` if present.
+        `machete-pre-rebase` is thus always launched before `pre-rebase`.
+        *  `machete-status-branch <branch-name>`
```
PawelLipski: Huh it's very hard to notice where the section for each hook beings :/// indent of description + some newlines would be awesome

git_machete/generated_docs.py
```diff
@@ -0,0 +1,944 @@
...
+
+        The hook that is executed for each branch displayed during `discover`, `status` and `traverse`.
+
+        The standard output of this hook is displayed at the end of the line, after branch name, (optionally) custom annotation and (optionally) remote sync-ness status.
+        Standard error is ignored. If the hook returns a non-zero exit code, both stdout and stderr are ignored, and printing the status continues as usual.
+
+        Note: the hook is always invoked with `ASCII_ONLY` variable passed into the environment.
+        If `status` runs in ASCII-only mode (i.e. if `--color=auto` and stdout is not a terminal, or if `--color=never`), then `ASCII_ONLY=true`, otherwise `ASCII_ONLY=false`.
+
+        Please see hook_samples for examples.
```
PawelLipski:
```suggestion
        Please see `hook_samples` directory in git-machete project for examples.
```

git_machete/generated_docs.py
amalota: Oh, this sounds like a great idea, will implement.

git_machete/generated_docs.py
amalota: @PawelLipski Yes, it's done, also i made some changes to the `github` .rst file

git_machete/generated_docs.py
amalota: @PawelLipski I wasn't able to do that but i managed to improve indentation for nested list's bullet points, from:

image

to

image

PawelLipski: Gr8 🎉

docs/source/cli_help/advance.rst
```diff
@@ -37,9 +37,6 @@ As an example, if ``git machete status --color=never --list-commits`` is as foll
       | |
       | | Add support and sample for machete-post-slide-out hook
       | o-feature/post-slide-out-hook
```
PawelLipski:
```suggestion
        |
        | Add support and sample for machete-post-slide-out hook
        o-feature/post-slide-out-hook
```

docs/source/cli_help/advance.rst
```diff
@@ -19,7 +19,7 @@ and subsequently slides out ``D``. All three steps require manual confirmation u
 
 The downstream ``D`` is selected according to the following criteria:
 
-    * if ``C`` has exactly one downstream (child) branch ``d`` connected with a :green:`green edge` (see :ref:`status` for definition) to ``C`` or is overridden, then ``d`` is selected as ``D``,
+    * if ``C`` has exactly one downstream (child) branch ``d`` connected with a :green:`green edge` (see help for :ref:`status` for definition) to ``C`` or is overridden, then ``d`` is selected as ``D``,
```
PawelLipski:
```suggestion
    * if ``C`` has exactly one downstream (child) branch ``d`` connected with a :green:`green edge` (see help for :ref:`status`) to ``C`` or is overridden, then ``d`` is selected as ``D``,
```

docs/source/cli_help/config.rst
```diff
@@ -9,7 +9,7 @@ Note: ``config`` is not a command as such, just a help topic (there is no ``git
 **Config keys:**
 
 ``machete.github.{remote,organization,repository}``:
-    When executing ``git machete github <subcommand>`` command, GitHub API server URL will be inferred from ``git remote``.
+    When executing ``git machete github <subcommand>`` command, following will happen:
```
PawelLipski:
```suggestion
    When executing ``git machete github <subcommand>`` command, the following will happen:
```

docs/source/cli_help/diff.rst
```diff
@@ -9,7 +9,7 @@ diff
     git machete d[iff] [-s|--stat] [<branch>]
 
 Runs ``git diff`` of the given branch tip against its fork point or, if none specified, of the current working tree against the fork point of the currently checked out branch.
-See :ref:`fork-point` for more details on meaning of the *fork point*.
+See help for :ref:`fork-point` for more details on meaning of the *fork point*.
```
PawelLipski:
```suggestion
See help for :ref:`fork-point` for more details on the meaning of *fork point*.
```

docs/source/cli_help/update.rst
```diff
@@ -12,7 +12,7 @@ Synchronizes the current branch with its upstream (parent) branch either by reba
 
 If updating by rebase, interactively rebases the current branch on the top of its upstream (parent) branch.
 The chunk of the history to be rebased starts at the fork point of the current branch, which by default is inferred automatically, but can also be set explicitly by ``--fork-point``.
-See :ref:`fork-point` for more details on meaning of the *fork point*.
+See help for :ref:`fork-point` for more details on meaning of the *fork point*.
```
PawelLipski:
```suggestion
See help for :ref:`fork-point` for more details on the meaning of *fork point*.
```

git_machete/generated_docs.py
```diff
@@ -7,12 +7,12 @@
 long_docs: Dict[str, str] = {
```
PawelLipski: One more thing... for consistency with `handmade_short_docs.py`, maybe rename the file to `generated_long_docs.py`

docs/source/cli_help/advance.rst
PawelLipski: @amalota resolved but looks unapplied yet 🤔

PawelLipski: Huh actually:
1. the change is applied in advance.rst
2. the change is **not** reflected in generated_docs.py
3. `tox -e check-py-docs` is failing on my machine for the current commit (`ffc42ce62af46ce58730f216fa09953fe3c84e59`)
4. **still, the CI somehow passed**... which is not expected 🤔

PawelLipski: Awww I see why, `TRUE` instead of `true` in CI config... lemme comment a suggestion there; once applied, it should fail the build (assuming `generated_docs.py` remain un-regenerated)

.circleci/config.yml
```diff
@@ -85,13 +85,13 @@ jobs:
...
 
-  "sphinx docs and python 3_10 git 2_33_1":
+  "docs and python 3_10 git 2_33_1":
     executor: ubuntu_executor
     steps:
       - checkout
       - run:
           name: Run tests against Python 3.10 and git 2.33.1 (with building docs)
-          command: PYTHON_VERSION=3.10 GIT_VERSION=2.33.1 BUILD_DOCS=true  bash ci/tox/ci-run.sh
+          command: PYTHON_VERSION=3.10 GIT_VERSION=2.33.1 BUILD_SPHINX_DOCS=true CHECK_PY_DOCS_UP_TO_DATE=TRUE bash ci/tox/ci-run.sh
```
PawelLipski:
```suggestion
          command: PYTHON_VERSION=3.10 GIT_VERSION=2.33.1 BUILD_SPHINX_DOCS=true CHECK_PY_DOCS_UP_TO_DATE=true bash ci/tox/ci-run.sh
```

.circleci/config.yml
```diff
@@ -191,7 +191,7 @@ mandatory_jobs: &mandatory_jobs
   - python 3_7 git 2_7_6
   - pep8 and python 3_8 git 2_25_0
   - coverage and python 3_9 git 2_30_1
-  - sphinx docs and python 3_10 git 2_33_1
+  - docs and python 3_10 git 2_33_1
```
PawelLipski, line 194: FYI, I've just changed the required checks for PR to be mergable to `develop`/`master` to accommodate for the new name

docs/source/cli_help/config.rst
```diff
@@ -8,78 +8,32 @@ Note: ``config`` is not a command as such, just a help topic (there is no ``git
...
 
-    * ``GIT_MACHETE_EDITOR``
-
-        Name of the editor used by ``git machete e[dit]``, example: ``vim`` or ``nano``.
-
-    * ``GIT_MACHETE_REBASE_OPTS``
-
-        Used to pass extra options to the underlying ``git rebase`` invocation (called by the executed command, such as: ``reapply``, ``slide-out``, ``traverse``, ``update``)
-        Example: ``GIT_MACHETE_REBASE_OPTS="--keep-empty --rebase-merges" git machete update``.
+``GIT_MACHETE_EDITOR``:
```
PawelLipski: Nit: pls use or not use colons after env var names consistently. Other files don't have colons here, and I'd remove them everywhere indeed

docs/source/cli_help/advance.rst
PawelLipski: Failed indeed, gr8: https://app.circleci.com/pipelines/github/VirtusLab/git-machete/2526/workflows/71d89164-a735-4439-9ade-b232838a2db5/jobs/12280

git_machete/generated_docs.py
```diff
@@ -0,0 +1,960 @@
...
+           <b>-s</b>, <b>--stat</b>
+              Makes `git machete diff` pass `--stat` option to `git diff`, so that only summary (diffstat) is printed.
+   """,
+    "discover": """
+        <b>Usage:</b><b>
+           git machete discover [-C|--checked-out-since=<date>] [-l|--list-commits] [-r|--roots=<branch1>,<branch2>,...] [-y|--yes]</b>
+
+        Discovers and displays tree of branch dependencies using a heuristic based on reflogs and asks whether to overwrite the existing definition `file` with the new discovered tree.
+        If confirmed with a `y[es]` or `e[dit]` reply, backs up the current definition file (if it exists) as `$GIT_DIR/machete~` and saves the new tree under the usual `$GIT_DIR/machete` path.
+        If the reply was `e[dit]`, additionally an editor is opened (as in `git machete edit<edit>`) after saving the new definition file.
```
PawelLipski: Whoops leftover `<edit>` from link

docs/source/cli_help/config.rst
```diff
@@ -9,7 +9,7 @@ Note: ``config`` is not a command as such, just a help topic (there is no ``git
 **Config keys:**
 
 ``machete.github.{remote,organization,repository}``:
-    When executing ``git machete github <subcommand>`` command, the following will happen:
+    When executing ``git machete github <subcommand>`` command, following will happen:
```
PawelLipski:
```suggestion
    When executing ``git machete github <subcommand>`` command, the following will happen:
```
(reverted change, probably some unpulled changes from GitHub)

git_machete/generated_long_docs.py
```diff
@@ -0,0 +1,960 @@
...
+
+        Outputs the absolute path of machete definition file.
+        The file is always called `machete` and is located in the git directory of the project.
+
+        Three cases are possible:
+           * if `git machete` is executed from a regular working directory (not a worktree or submodule), the file is located under `.git/machete`,
+           * if `git machete` is executed from a <b>worktree</b>, the file path depends on the `machete.worktree.useTopLevelMacheteFile` config key value:
+              - if `machete.worktree.useTopLevelMacheteFile` is true (default), the file is located under `.git/machete`
+              - if `machete.worktree.useTopLevelMacheteFile` is false, the file is located under `.git/worktrees/.../machete`,
+
```
PawelLipski, line 341: Nit: any chance to remove this newline, or to add a newline b/w 1st and 2nd top-level bullets for consistency?

git_machete/generated_long_docs.py
```diff
@@ -0,0 +1,960 @@
...
+           Repository name
+              E.g. `machete.github.repository` = `git-machete`
+
+        To do this, run `git config --local --edit` and add the following section:
+        <dim>
+          [machete "github"]
+              organization = <organization_name>
+              repository = <repo_name>
+              remote = <remote_name>
+        </dim>
```
PawelLipski, line 459: I'd add some header here like `Subcommands`

PawelLipski: Coz in Python docs there's a sudden transition from the sample git config to `anno-prs`

git_machete/generated_long_docs.py
```diff
@@ -0,0 +1,960 @@
...
+
+        Roughly equivalent to `git checkout $(git machete show <direction>)`.
+   """,
+    "help": """
+        <b>Usage:</b><b>
+           git machete help [<command>]</b>
+
+        Prints a summary of this tool, or a detailed info on a command if provided.
+   """,
+    "hooks": """
```
PawelLipski, line 550: :benissimo:, looks really neat now

docs/source/cli_help/github.rst
```diff
@@ -15,62 +15,55 @@ Creates, checks out and manages GitHub PRs while keeping them reflected in branc
 .. include:: github_api_access.rst
 .. include:: github_config_keys.rst
 
-``anno-prs``:
+**Subcommands:**
```
PawelLipski, line 18: :benissimo: now

docs/source/cli_help/advance.rst
PawelLipski: (okay resolving)

PawelLipski (approved):
That was a big one 😅 kudoz for the hard work!
One nitpick left, otherwise :shipit:

git_machete/generated_long_docs.py
amalota: @PawelLipski Unfortunately it's not easily doable, i will look into it next week.

PawelLipski: Okay, pls open a separate issue for that so as not to block this PR

amalota: https://github.com/VirtusLab/git-machete/issues/657

PR #667, opened 2022-09-29
Issue #610: `git machete diff`, contrary to the the docs, diffs HEAD (and not working directory) against fork-point

### LINKED ISSUE #610: `git machete diff`, contrary to the the docs, diffs HEAD (and not working directory) against fork-point
https://git-machete.readthedocs.io/en/stable/#diff says:

> Runs git diff of the given branch tip against its fork point or, if none specified, of the current working tree against the fork point of the currently checked out branch.

This is no longer the behavior of `git machete diff` however. Instead of calling `git diff <fork-point>`, it calls `git diff <fork-point> <current-branch>`. This is different because the latter does NOT take changes to working tree into account (and I'd expect to see them as well when doing `git machete diff`, analogously to `git diff`).

amalota:


git_machete/cli.py
```diff
@@ -463,7 +463,12 @@ def set_utils_global_variables(
 
 def get_local_branch_short_name_from_arg_or_current_branch(
         branch_from_arg: Optional[AnyBranchName], git_context: GitContext) -> LocalBranchShortName:
-    return LocalBranchShortName.of(branch_from_arg.replace('refs/heads/', '')) if branch_from_arg else git_context.get_current_branch()
+    return get_local_branch_short_name_from_arg(branch_from_arg) if branch_from_arg else git_context.get_current_branch()
+
+
+def get_local_branch_short_name_from_arg(
+        branch_from_arg: Optional[AnyBranchName]) -> LocalBranchShortName:
```
```suggestion
        branch_from_arg: Optional[AnyBranchName]) -> Optional[LocalBranchShortName]:
```

PawelLipski (approved):
Pls also update RELEASE_NOTES.md!

PR #712, opened 2022-10-24
Issue #645: Add `machete.traverse.push` config key with default value `true`

### LINKED ISSUE #645: Config keys for default settings for subcommands
Hi! Firstly, thanks for maintaining `git-machete` :)

Is it possible to configure "default settings" for subcommands? Specifically, I want to ensure that I never push by default, and rather than remembering `--no-push` every time, it'd be great to have that in my `.gitconfig`.

Thanks in advance for your advice!

amalota:
- [x] Implement
- [x] Update documentation
- [x] Add test case


RELEASE_NOTES.md
```diff
@@ -6,6 +6,7 @@
 - added: `git machete help <command>` now accepts command aliases
 - fixed: removed stack trace when terminating `git machete` command prompt with `Ctrl+D`
 - added: support for Python 3.11
+- added: machete config key `machete.traverse.push` that controls `--push` flag of `traverse` command
```
PawelLipski: I wouldn't say it controls the flag really 🤔 It just controls the default behavior of `traverse`... and `--push` also controls the behavior of traverse. If both are present, then the flag takes precedence. Compare e.g. `git commit`, `--author` flag and `git config user.*`: the flag takes precedence over config; it's not that either flag controls the config, or config controls the flag.

docs/source/cli_help/traverse.rst
```diff
@@ -84,7 +84,8 @@ Unlike with e.g. ``git rebase``, there is no special ``--continue`` flag, as ``t
 
 --no-push-untracked          Do not push untracked branches to remote, re-enable via ``--push-untracked``.
 
---push                       Push all (both tracked and untracked) branches to remote --- default behavior.
+--push                       Push all (both tracked and untracked) branches to remote --- default behavior. Default behaviour can be changed
```
PawelLipski, line 87: Flag, if present, should take precedence over config 🎏

git_machete/options.py
```diff
@@ -1,12 +1,12 @@
...
 
 from git_machete.exceptions import MacheteException
-from git_machete.git_operations import AnyBranchName, AnyRevision, LocalBranchShortName
+from git_machete.git_operations import AnyBranchName, AnyRevision, GitContext, LocalBranchShortName
 
 
 class CommandLineOptions:
 
-    def __init__(self) -> None:
+    def __init__(self, git: GitContext) -> None:
```
PawelLipski: Huh I wouldn't introduce that dependency. Traverse should use `CommandLineOptions` and `git config` as two independent inputs, and resolve their precedence on its own. Such logic shoudln't reside in `CommandLineOptions` &mdash; it should be just a dumb container for options, and nothing else

amalota: Hmm, in order for `traverse` to receive `CommandLineOptions` and `git config` as two independent inputs the hole options providing would need some major refactoring or some new unnecessary params, here's why:
- We want to achieve this options override order:
   1️⃣  Default values - CommandLineOptions
   2️⃣  Config values - only exists for some options
   3️⃣  Passed flags -> values - may be all maybe be none
- ATM, traverse (actually all commands) receive command options that follow logic where 1️⃣ is overridden by 3️⃣ 
- Now, applying 2️⃣ in between them in `traverse` is difficult:
   - we can get default values pretty easily but traverse doesn't know which flags were actually passed (it only receives 
   updated default values). So in order to know which flags were actually passed you have 2 options (1 - add some 
   parameter that stores list of options that were passed, 2 - find diff between updated values and default values) -> 
   both options are IMO bad

That's why i resolved config keys straight in the CommandLineOptions. But i understand that it's just a container so i propose resolving default options from config here (`update_cli_options_using_config_keys(cli_opts, git)`):

image

This solutions can also be generalised into other commands that might in the future use default values stored in config. @PawelLipski wdyt?

RELEASE_NOTES.md
PawelLipski: Resolved? 🤔

git_machete/options.py
PawelLipski: Okay! let's proceed with `update_cli_options_using_config_keys` then

docs/source/traverse_config_key.rst
```diff
@@ -0,0 +1,4 @@
+ .. note::
```
PawelLipski, line 1: Docs for `config` are now (even before this PR) inconsistent:

image

Some parts as Note, some as plain text. I'd unify towards **plain text**. Can be done on a separate PR ofc

RELEASE_NOTES.md
amalota: Yes

docs/source/traverse_config_key.rst
amalota: Hmm, yes, it's not perfect, the Note format is being used since we wanted to store the doc text in one place and use it across several commands as a Note, will have to see if there is a straight forward way to convert Note to plain text -> https://github.com/VirtusLab/git-machete/issues/719

RELEASE_NOTES.md
PawelLipski: But... but... as discussed, I don't think the config controls the **flag**, it controls the **behavior** (which also happens to be controlled by the flag). And if anything, it's the flag that has the higher priority than the config, so one could even argue that it's `--push` that controls `machete.traverse.push` and not the other way round 😅

PawelLipski: So... the current formulation is deeply confusing ☹️

git_machete/options.py
PawelLipski: For clarity: are you planning to apply this change on this PR?

amalota: @PawelLipski Yes, i just did

RELEASE_NOTES.md
amalota: Oh, a commit must have disappeared, i remember that i changed it

git_machete/client.py
```diff
@@ -22,6 +22,7 @@
     create_pull_request, checkout_pr_refs, derive_pull_request_by_head, derive_pull_requests,
     get_github_token_possible_providers, get_parsed_github_remote_url, get_pull_request_by_number_or_none, GitHubPullRequest,
     is_github_remote_url, RemoteAndOrganizationAndRepository, set_base_of_pull_request, set_milestone_of_pull_request)
+from git_machete.options import CommandLineOptions
```
PawelLipski: Hmmm needed now even though it wasn't previously? 🤔

git_machete/options.py
```diff
@@ -1,7 +1,7 @@
 from typing import Optional, List
 
 from git_machete.exceptions import MacheteException
-from git_machete.git_operations import AnyBranchName, AnyRevision, LocalBranchShortName
+from git_machete.git_operations import AnyBranchName, AnyRevision, GitContext, LocalBranchShortName
```
PawelLipski: Redundant import

git_machete/git_operations.py
```diff
@@ -352,6 +352,12 @@ def get_boolean_config_attr(self, key: str, default_value: bool) -> bool:
             return self.__config_cached.get(key.lower()) == 'true'
         return default_value
 
+    def get_boolean_config_attr_or_none(self, key: str) -> Optional[bool]:
```
PawelLipski, line 354: Implement `get_boolean_config_attr` in terms of this method to keep DRY

PawelLipski (approved):

git_machete/git_operations.py
```diff
@@ -347,10 +347,15 @@ def get_config_attr_or_none(self, key: str) -> Optional[str]:
         return self.__config_cached.get(key.lower())
 
     def get_boolean_config_attr(self, key: str, default_value: bool) -> bool:
+        if self.get_boolean_config_attr_or_none(key) is not None:
```
PawelLipski, line 350: Extract a var for the result of `get_boolean_config_attr_or_none`

PR #718, opened 2022-10-27
Issue #711: Update docs with information that flag order defines their precedence over each other

### LINKED ISSUE #711: Update docs with information that flag order defines their precedence over each other
To be determined whether do it with a note type of information that's added to each command's documentation or just add it once in some general place (it might get easily missed so i would go with the former option)

amalota:
- [x] Update traverse `push`
- [x] Check for any other situations and update them

### DIFF
--- a/docs/source/cli_help/traverse.rst
+++ b/docs/source/cli_help/traverse.rst
@@ -27,10 +27,6 @@ Traverses the branch tree in pre-order (i.e. simply in the order as they occur i
 By default ``traverse`` starts from the current branch.
 This behaviour can, however, be customized using options: ``--start-from=``, ``--whole`` or ``-w``, ``-W``.
 
-By default ``traverse`` asks if the branch should be pushed, this behaviour can, however, be changed with the ``machete.traverse.push`` configuration key.
-It can also be customized using options: ``--[no-]push`` or ``--[no-]push-untracked`` --- the order of the flags defines their precedence over each other
-(the one on the right overriding the ones on the left). More on them in the **Options** section below.
-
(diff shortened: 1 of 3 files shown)


docs/source/cli_help/traverse.rst
```diff
@@ -27,6 +27,10 @@ Traverses the branch tree in pre-order (i.e. simply in the order as they occur i
 By default ``traverse`` starts from the current branch.
 This behaviour can, however, be customized using options: ``--start-from=``, ``--whole`` or ``-w``, ``-W``.
 
+By default ``traverse`` asks if the branch should be pushed, this behaviour can, however, be changed with the ``machete.traverse.push`` configuration key.
```
PawelLipski: I'd move it below, to right before `If the traverse flow is stopped..`

PawelLipski: Or maybe even better, as a Note within `if the branch is not tracked on a remote...` bullet point?

docs/source/cli_help/traverse.rst
```diff
@@ -56,6 +56,10 @@ For each branch, the command:
 
       - prints the updated ``status``.
 
+By default ``traverse`` asks if the branch should be pushed, this behaviour can, however, be changed with the ``machete.traverse.push`` configuration key.
```
PawelLipski, line 59:
```suggestion
By default ``traverse`` asks if the branch should be pushed. This behaviour can, however, be changed with the ``machete.traverse.push`` configuration key.
```

PawelLipski (approved):

PR #732, opened 2022-11-14
Add linter to check f-strings usage

### LINKED ISSUE #612: Wire up a linter that would detect would-be interpolated strings that lack the leading `f`
Such cases were common in the codebase, most notably in one user-facing message:

```
git_machete/client.py:1627:                "Consider running:\n  `git machete fork-point --unset-override {branch}`\n"))
```

I've used some basic regexes to find them, but they spit out a lot of false positives, so they aren't suitable for CI:
```
git grep -En "(^|[^f])'[^']*\{[^']+\}[^']*'" -- '*.py'
git grep -En '(^|[^f])"[^"]*\{[^"]+\}[^"]*"' -- '*.py'
```

Maybe there's some standard linter tool that would capture such cases with less false positives? and where uninterpolated `{...}` are really needed (e.g. in git revs like `HEAD@{tree}`), they could be marked with some `# noqa`

AABur:
fix #612 

Some strings are false positives and #noqa-ed with F003

See [flake8-use-fstring](https://github.com/MichaelKim0407/flake8-use-fstring) for more details


AABur:
https://app.circleci.com/pipelines/github/VirtusLab/git-machete/2779/workflows/fe004b8f-5e0d-4b8d-a6d4-9e2ded36240d/jobs/13931?invite=true#step-103-34

https://app.circleci.com/pipelines/github/VirtusLab/git-machete/2779/workflows/fe004b8f-5e0d-4b8d-a6d4-9e2ded36240d/jobs/13931?invite=true#step-103-40

These two errors of the "strings split without delimiter" checker are wrong. For f-strings it's correct.

git_machete/client.py
```diff
@@ -979,17 +979,17 @@ def print_line_prefix(branch_: LocalBranchShortName, suffix: str) -> None:
                             # the above call to 'fork_point_hash'.
                             fp_branches_formatted: str = " and ".join(
                                 sorted(underline(lb_or_rb) for lb, lb_or_rb in fork_point_branches_cached[branch]))
-                            fp_suffix: str = " %s %s %s seems to be a part of the unique history of %s" % \
```
PawelLipski, line 982: Wow we've still got some old-school `%` format 😯

git_machete/client.py
```diff
@@ -979,17 +979,17 @@ def print_line_prefix(branch_: LocalBranchShortName, suffix: str) -> None:
...
-                                             (colored(utils.get_right_arrow(), AnsiEscapeCodes.RED),
-                                              colored("fork point ???", AnsiEscapeCodes.RED),
-                                              "this commit" if opt_list_commits_with_hashes else f"commit {commit.short_hash}",
-                                              fp_branches_formatted)
+                            right_arrow = colored(utils.get_right_arrow(), AnsiEscapeCodes.RED)
+                            fork_point = colored("fork point ???", AnsiEscapeCodes.RED)
+                            fp_suffix: str = f' {right_arrow} {fork_point}\
```
PawelLipski, line 984: Hmmm won't there be a space missing between `???` and `this commit`? which would be a change from the original impl 🤔

git_machete/client.py
```diff
@@ -2134,7 +2136,7 @@ def __derive_remote_and_github_org_and_repo(self,
                     'corresponds to GitHub (see `git remote -v` for details). \n'
                     'It is possible that you are using custom GitHub URL.\n'
                     'If that is the case, you can provide repository information explicitly, via these 3 git config keys: '
-                    '`machete.github.{remote,organization,repository}`\n'))
+                    '`machete.github.[remote,organization,repository]`\n'))
```
PawelLipski: Hmm I wanted to use the "brace expansion" syntax to evoke the [analogous mechanism in shell](https://www.gnu.org/software/bash/manual/html_node/Brace-Expansion.html)... square brackets are more typically use to signify optional stuff. Maybe leave braces as they were, just add some `noqa`?

git_machete/client.py
```diff
@@ -2153,7 +2155,7 @@ def __derive_remote_and_github_org_and_repo(self,
             f'Multiple non-origin remotes correspond to GitHub in this repository: '
             f'{", ".join(remote_and_github_org_and_repo.keys())} -> aborting. \n'
             f'You can also select the repository by providing 3 git config keys: '
-            '`machete.github.{remote,organization,repository}`\n')
+            f'`machete.github.[remote,organization,repository]`\n')
```
PawelLipski: Ditto

PawelLipski:
If it comes to the failing CI... pls replace the contents of `ci/checks/prohibit-strings-split-without-delimiter.awk` with the following:

```awk
#!/usr/bin/env awk

BEGIN { exit_code = 0 }

FNR == 1 { prev_is_unsafe_terminated_string = 0 }

prev_is_unsafe_terminated_string && /^ *f?["'][^ ]/ {
  print FILENAME ":" FNR-1 "-" FNR ": it looks that the string is split without a delimiter, words will be glued up together when printed"
  print ""
  print prev
  print $0
  print "\n"
  exit_code = 1
}

                      { prev_is_unsafe_terminated_string = 0 }

/[^ ]["'] *\\?$/      { prev_is_unsafe_terminated_string = 1 }

/(\\n|\})["'] *\\?$/  { prev_is_unsafe_terminated_string = 0 }

{ prev = $0 }

END { exit exit_code }
```

git_machete/client.py
AABur: No problem. To use {  } in f-string you have to use {{  }} 
I fixed it

git_machete/client.py
```diff
@@ -2155,7 +2155,7 @@ def __derive_remote_and_github_org_and_repo(self,
             f'Multiple non-origin remotes correspond to GitHub in this repository: '
             f'{", ".join(remote_and_github_org_and_repo.keys())} -> aborting. \n'
             f'You can also select the repository by providing 3 git config keys: '
-            f'`machete.github.[remote,organization,repository]`\n')
+            f'`machete.github.{{remote,organization,repository}}`\n')
```
PawelLipski, line 2158: Or just drop `f` and add `noqa`!

PawelLipski (approved):

git_machete/client.py
AABur: `noqa` should only be used where it is not at all possible without it.
If the f-string style is adopted, it should be used. Using {{ to denote { in f-string is the f-string-style

PR #793, opened 2023-01-19
Add support for `slide-out=no` in branch qualifiers

### LINKED ISSUE #749: Add support for `slide-out=no` in branch qualifiers
Alongside the existing `rebase=no` and `push=no`.

I just now realized it'd be very useful for branches like `develop` (i.e. an "almost top-level" branch that still has a parent, like `master` or `main`).

amalota:
TODO:
- [x] implement
- [x] update docs
- [x] add test cases


git_machete/annotation.py
```diff
@@ -13,8 +13,10 @@ def __init__(self, annotation: str):
         self._annotation_without_qualifiers = annotation
         self._rebase_text = ''
         self._push_text = ''
+        self._slide_out_text = ''
         self.rebase = True
         self.push = True
+        self.slide_out = None
```
PawelLipski, line 18: `True`?

git_machete/annotation.py
```diff
@@ -31,11 +33,17 @@ def __init__(self, annotation: str):
...
+            self.slide_out = False
+            self._slide_out_text = 'slide-out=no'
+            self._annotation_without_qualifiers = re.sub(sub_pattern('slide-out'), ' ', self._annotation_without_qualifiers)
+
     def get_annotation_text_without_qualifiers(self) -> str:
         return self._annotation_without_qualifiers.strip()
 
     def get_qualifiers_text(self) -> str:
-        return f'{self._rebase_text.strip()} {self._push_text.strip()}'.replace('  ', ' ').strip()
+        return f'{self._rebase_text.strip()} {self._push_text.strip()} {self._slide_out_text.strip()}'.replace('  ', ' ').strip()
```
PawelLipski: Actually... is there a need to strip any of these? given what values can be assigned to them

tests/test_traverse.py
```diff
@@ -472,3 +472,48 @@ def test_traverse_qualifiers_no_rebase_no_push(self, mocker: Any) -> None:
...
+            master
+            \thotfix/add-trigger
+            \t\tignore-trailing
+            """
+
+        body = dedent(body).strip()
+        rewrite_definition_file(body)
+        (
+            self.repo_sandbox.check_out('develop')
+            .merge('call-ws')
```
PawelLipski: Nit: join lines, no need to wrap in parens

git_machete/annotation.py
amalota: @PawelLipski  didn't want to assume `True` since traverse by default doesn't slide-out the branch, it does so only when certain conditions are met.

git_machete/annotation.py
amalota: @PawelLipski Right, there is no need to strip these values since we are hard coding them.

git_machete/annotation.py
PawelLipski: >  it does so only when certain conditions are met.

Hmmm but the same holds true of rebase and push. The case when slide-out happens might be less frequent than the case for rebase or push... but still, none of these 3 action happens _by default_

PawelLipski: So I can't see a reason to treat slide-out differently!

git_machete/annotation.py
```diff
@@ -13,8 +13,10 @@ def __init__(self, annotation: str):
         self._annotation_without_qualifiers = annotation
         self._rebase_text = ''
         self._push_text = ''
+        self._slide_out_text = ''
         self.rebase = True
         self.push = True
+        self.slide_out = None
 
         match_pattern: Callable[[str], str] = lambda text: f'.*\\b{text}=no\\b.*'
         sub_pattern: Callable[[str], str] = lambda text: f'[ ]?{text}=no[ ]?'
```
PawelLipski: Nit: maybe just nested defs instead of lambdas? easier to read

docs/source/cli_help/traverse.rst
```diff
@@ -65,16 +65,16 @@ It will pick up the walk from the current branch (unless ``--start-from=`` or ``
 Unlike with e.g. ``git rebase``, there is no special ``--continue`` flag, as ``traverse`` is stateless
 (doesn't keep a state of its own like ``git rebase`` does in ``.git/rebase-apply/``).
 
-The rebase and push behavior of ``traverse`` can also be customized for each branch separately using *branch qualifiers*.
-There are ``rebase=no`` and ``push=no`` qualifiers that can be used to opt out of default behavior (rebasing and pushing).
-The qualifier can appear anywhere in the annotation but needs to be separated by a whitespace from any other character, e.g. ``some_annotation_text rebase=no push=no``.
+The rebase, push and slide out behavior of ``traverse`` can also be customized for each branch separately using *branch qualifiers*.
```
PawelLipski: Nit: for grammatical clarity:
```suggestion
The rebase, push and slide-out behaviors of ``traverse`` can also be customized for each branch separately using *branch qualifiers*.
```

docs/source/cli_help/traverse.rst
```diff
@@ -65,16 +65,16 @@ It will pick up the walk from the current branch (unless ``--start-from=`` or ``
...
+There are ``rebase=no``, ``push=no`` and ``slide-out=no`` qualifiers that can be used to opt out of default behavior (rebasing, pushing and sliding the branch out of the dependency tree).
+The qualifier can appear anywhere in the annotation but needs to be separated by a whitespace from any other character, e.g. ``some_annotation_text rebase=no push=no slide-out=no``.
 Qualifiers can only be overwritten by manually editing ``.git/machete`` file or modifying it with ``git machete e[dit]`` or by updating annotations with ``git machete anno``.
```
PawelLipski: Uber-nit:
```suggestion
Qualifiers can only be overwritten by manually editing ``.git/machete`` file or modifying it with ``git machete e[dit]``, or by updating annotations with ``git machete anno``.
```

git_machete/client.py
```diff
@@ -709,6 +709,9 @@ def traverse(
 
             needs_slide_out: bool = self.__is_merged_to_upstream(
                 branch, opt_no_detect_squash_merges=opt_no_detect_squash_merges)
+            if branch in self.annotations:
+                if self.annotations[branch].qualifiers.slide_out is not None:
+                    needs_slide_out = self.annotations[branch].qualifiers.slide_out
```
PawelLipski: Hmmm but the current logic can set `needs_slide_out` to `True` even if it was `False` originally, right? 🤔 the qualifiers should only ever cause do the opposite change: to `False` even if it was `True`

amalota: @PawelLipski No, the slide-out qualifier can only force the change from `True` or `False` to `False`. To explain the logic you are referring to, the value of `self.annotations[branch].qualifiers.slide_out` can only be `None` if the qualifier is not present in which case the value of `needs_slide_out` remains unchanged. The value of `self.annotations[branch].qualifiers.slide_out` can be `False` if the qualifier is present and equals exactly to `slide-out=no` and in that case it would change the value of `needs_slide_out` to `False` whether it was `True` or `False` beforehand.

git_machete/annotation.py
amalota: Yeah, i was 50/50 on this one, decided that it's nicer to have one liners than defs but will change

git_machete/annotation.py
amalota: @PawelLipski Yes, you're right, but still, i don't like the way it would work when assuming `True` value in this example:
1. The conditions were not met so `needs_slide_out` value is `False`, annotations are present for that branch (e.g. `rebase=no`) so it would assign `True` to `needs_slide_out` which would be wrong IMO. To avoid this i added extra check + default `None` value.

git_machete/client.py
PawelLipski: Hmmm still... pls note that the current approach introduces an inconsistency, compared with how `needs_parent_sync`/`qualifiers.rebase` and `needs_remote_sync`/`qualifiers.push` is handled 🤔 

Neither `qualifiers.push` nor `qualifiers.rebase` can ever be `None`, only `False` and then:

```python
                if needs_parent_sync and branch in self.annotations:
                    needs_parent_sync = self.annotations[branch].qualifiers.rebase
```

git_machete/annotation.py
PawelLipski: > annotations are present for that branch (e.g. rebase=no) so it would assign True to needs_slide_out

Hmmm for clarity... how would the existence of `rebase=no` qualifier (as opposed to `slide-out=no`) cause `needs_slide_out` to be `True`? 🤔

git_machete/client.py
```diff
@@ -710,6 +710,9 @@ def traverse(
...
```
PawelLipski: I'd suggest just sth like:
```suggestion
                if needs_slide_out and branch in self.annotations:
                    needs_slide_out = self.annotations[branch].qualifiers.slide_out
```

PawelLipski: A different logic can ofc be applied... but I'd then see it applied consistently for all three qualifiers 🤔

git_machete/annotation.py
amalota: Well, using logic used in existing qualifiers i.e. 
```
if branch in self.annotations:
      needs_slide_out = self.annotations[branch].qualifiers.slide_out
```
would cause `needs_slide_out` to be set `True` if any of the three qualifiers were present (it would't even need to be `slide-out=no`) and that would be wrong, @PawelLipski do you see it now?

git_machete/client.py
amalota: @PawelLipski Hmm, this implementation would also fail in the example case in the comment two comments above 🆙

PawelLipski: > @PawelLipski Hmm, this implementation would also fail in the example case in the comment two comments above 🆙

Pls re-specify the exact case and what would be the actual vs expected behavior 🤔

amalota: Ok, lets consider 3 cases for qualifiers values (no qualifiers, qualifiers but without `slide-out=no`, `slide-out=no` qualifier) and 2 cases for initial value of `needs_slide_out`:
`needs_slide_out`=`True`:
    - no qualifiers
          - current implementation:
             -  no change in the value of `needs_slide_out` -> correct
          - your proposed implementation:
             -  no change in the value of `needs_slide_out` -> correct
    - `rebase=no` qualifier
          - current implementation:
             -  no change in the value of `needs_slide_out` -> correct
          - your proposed implementation:
             - would change `needs_slide_out` from `True` to `True` -> correct
    - `slide-out=no` qualifier
          - current implementation:
             -  would change the value of `needs_slide_out` from `True` to `False` -> correct
          - your proposed implementation:
             - would change `needs_slide_out` from `True` to `False` -> correct
`needs_slide_out`=`False`:
   easier example - no need to dig deeper, both implementations are correct.

So to sum up, both are correct (my bad, i though your proposition fails in one case but turns out it doesn't), but taking into account that your proposition follows the same pattern as already existing implementation for rebase/push qualifiers - it will be used instead of my current one.

amalota: Ach, formatting doesn't work, i'll add the screenshot below:
image

PawelLipski: ok! gr8

git_machete/client.py
```diff
@@ -710,6 +710,9 @@ def traverse(
 
             needs_slide_out: bool = self.__is_merged_to_upstream(
                 branch, opt_no_detect_squash_merges=opt_no_detect_squash_merges)
+            if branch in self.annotations:
+                if needs_slide_out and branch in self.annotations:
```
PawelLipski: Whoops one of the checks for `branch in self.annotations` seems redundant 🤔

PawelLipski (approved):

Issue #763, opened 2023-01-02
Provide tests for retrieving GitHub token


PawelLipski:
A test for `git_machete.github.github_remote_url_patterns`, with access to files like `~/.config/hub` or `~/.github-token` mocked.

PR #800, opened 2023-01-31
Extract `GitHubClient` class

### LINKED ISSUE #761: Extract `GitHubClient` class
So that there's no need to pass values like `domain`, `org`, `repo` to almost every method in github.py separately. Instead, one would build a `GitHubClient` object for the given domain/org/repo, and call its methods

amalota:


git_machete/client.py
```diff
@@ -1471,12 +1472,12 @@ def is_excluded_reflog_subject(hash_: str, gs_: str) -> bool:
         return result
 
     def sync_annotations_to_github_prs(self) -> None:
-        domain = self.__derive_github_domain()
-        _, org, repo = self.__derive_remote_and_github_org_and_repo(domain)
+        _, org, repo = self.__derive_remote_and_github_org_and_repo()
+        github_client = self.__get_github_client(org, repo)
```
PawelLipski: Hmmm wouldn't it make sense to also pass `domain` to `__get_github_client`?

git_machete/github.py
```diff
@@ -76,6 +76,12 @@ class GithubTokenAndTokenProvider(NamedTuple):
     token_provider: str
 
 
+class GitHubClient(NamedTuple):
```
PawelLipski: Whoops that's not a service client class... I'd rather expect the functions like `create_pull_request` to become its methods!

PawelLipski: (or in case this is WIP, pls mark the PR as draft)

amalota:
@PawelLipski Ready for review

git_machete/github.py
```diff
@@ -104,282 +163,296 @@ def get_token_from_gh() -> Optional[GithubTokenAndTokenProvider]:
...
+    def get_repo_and_org_names_by_id(self,
+                                     repo_id: str
+                                     ) -> str:
+        repo = self.__fire_github_api_request(path='GET',
+                                              method=f'/repositories/{repo_id}')
+        return str(repo['full_name'])
 
 
-def checkout_pr_refs(git: GitContext, remote: str, pr_number: int, branch: LocalBranchShortName) -> None:
+def checkout_pr_refs(git: GitContext,
```
PawelLipski: Nit: I'd extract a method like `get_remote_ref_for_github_pr` which would return `'pull/{pr_number}/head:{branch}'`, but move the method `checkout_pr_refs` to `MacheteClient`

PawelLipski (approved):
:benissimo:

git_machete/github.py
amalota: @PawelLipski I will move the method to the class but i don't think that extracting a new method is necessary here since all the paths for each `MacheteClient` method are coded inside the method so i would stick to it here as well.

git_machete/github.py
```diff
@@ -104,282 +163,296 @@ def get_token_from_gh() -> Optional[GithubTokenAndTokenProvider]:
...
                 for line in file.readlines():
-                    if line.endswith(" " + domain):
+                    if line.endswith(" " + self.__domain):
                         token = line.split(" ")[0]
-                        return GithubTokenAndTokenProvider(token=token, token_provider=token_provider)
-                    elif domain == DEFAULT_GITHUB_DOMAIN and " " not in line.rstrip():
-                        return GithubTokenAndTokenProvider(token=line.rstrip(), token_provider=token_provider)
-        return None
-
-    return (get_token_from_env() or
```
PawelLipski, line 159: Aaaaaaa 😱

git_machete/github.py
```diff
@@ -41,57 +28,129 @@ def __init__(self,
...
+class GitHubToken:
+    GITHUB_TOKEN_ENV_VAR = 'GITHUB_TOKEN'
 
+    def __init__(self,
+                 domain: str
+                 ) -> None:
+        self.__domain = domain
+        self.__value: Optional[str] = None
+        self.__provider: Optional[str] = None
+        for token_retrieval_method in [self.__get_token_from_hub, self.__get_token_from_gh, self.__get_token_from_env,
```
PawelLipski, line 106: The order has changed ‼️ I'll fix that now, but generally https://github.com/VirtusLab/git-machete/issues/763 FTW

git_machete/github.py
```diff
@@ -104,282 +163,296 @@ def get_token_from_gh() -> Optional[GithubTokenAndTokenProvider]:
...
-                # - user: {username2}
-                #   oauth_token: *******************
-                #   protocol: {protocol}
-                found_host = False
-                for line in config_hub.readlines():
-                    if line.rstrip() == domain + ":":
-                        found_host = True
-                    elif found_host and line.lstrip().startswith("oauth_token:"):
-                        result = re.sub(' *oauth_token: *', '', line).rstrip().replace('"', '')
-                        return GithubTokenAndTokenProvider(token=result,
```
PawelLipski, line 133: Whoops another problem introduced with this PR... there's a change in logic: now the loop continues after the oauth token is found, which might leads to picking token for another host

PR #802, opened 2023-02-01
Provide tests for GitHub API pagination

### LINKED ISSUE #762: Provide tests for GitHub API pagination
Missing test coverage for PR #754.

amalota:


tests/test_github.py
```diff
@@ -52,6 +53,38 @@ def mock_input(msg: str) -> str:
...
+        link = f'<https://api.github.com/repositories/1300192/pulls?page={mock_info.counter + 2}>; rel="next"'  # type: ignore[attr-defined]
+    else:
+        link = ''
+    mock_info.counter += 1  # type: ignore[attr-defined]
+    return {"link": link}
+
+
+mock_info.counter = mock_read.counter = 0  # type: ignore[attr-defined]
+
+
```
amalota, line 87: I mocked these two functions to simulate 3 page long (with 100 PRs per page) GitHub API response.

tests/test_github.py
```diff
@@ -52,6 +54,51 @@ def mock_input(msg: str) -> str:
...
+        return int(text) if text.isdigit() else text
+
+    def natural_keys(text: str) -> List[Any]:
+        return [atoi(c) for c in re.split(r'(\d+)', text)]
+
+    if self._local_branches_cached is None:
+        self._load_branches()
+    return sorted(self._local_branches_cached, key=natural_keys)
+
+
```
amalota: The tests where not passing when number of prs per page was > 10 because of branch name sorting. e.g
```
feature_1
feature_11
feature_12
feature_2
itd
```
which i changed into natural sorting e.g.
```
feature_1
feature_2
...
feature_10
feature_11
itd
```
I actually think that we can add this sorting to the code base in separate PR since it seems to be useful, @PawelLipski wdyt?

PawelLipski: Huh... this sorting would apply in what places specifically? `discover` and `github checkout-prs` right? I mean... in real-life/non-test scenarios it's extremely rare that natural sorting would ever yield a different result than plain ASCII sorting; I wouldn't give it a high priority 🤔

PawelLipski (approved):

PR #815, opened 2023-02-13
Fix zsh slide-out completion

domesticsimian:
Move the TODO comment above the multi-line _arguments function call so that subsequent lines are part of the call.


PawelLipski:
Whoops good spot! 😅 actually... lemme check why the existing CI check didn't capture this error...

PawelLipski:
Release notes to be updated as well 😅

domesticsimian:
so many checks! 😅  I hope it's good now.

PawelLipski (approved):

PR #825, opened 2023-02-17
Make `GitHubToken` class immutable

### LINKED ISSUE #819: Rework `GitHubToken` to be immutable again
https://github.com/VirtusLab/git-machete/pull/800 changed the logic of `GitHubToken` so that:
1. an instance of `GitHubToken` can correspond to **lack** of token, which is confusing
2. properties are re-assigned (from the initial `None` to a non-`None` value)

Both are antipatterns IMO, and contributed to bugs later hotfixed in #816.

Let's restore certain pre-#800 aspects of the implementation:
1. `GitHubToken` is instantiated via a factory method (used to be `__get_github_token_and_provider`, can be a class method of `GitHubToken` now) that can return `None` if no token can be found in the env var/on the disk
2. properties are non-`None`, immutable, assigned only in c'tor

Let's, however, first provide test coverage (#763) to make sure that this refactor doesn't break anything.

amalota:


git_machete/github.py
```diff
@@ -296,7 +295,7 @@ def __fire_github_api_request(self,
                 else:
                     raise MacheteException(
                         first_line + f'You might not have the required permissions for this repository.\n'
-                                     f'Provide a GitHub API token with `repo` access via {self.__token.provider}.\n'
+                                     f'Provide a GitHub API token with `repo` access.\n'
```
amalota, line 298: Logical error, we wanted to print token provider in the else statement which is only entered when token is None

git_machete/github.py
```diff
@@ -86,47 +86,54 @@ def from_url(cls,
...
 
-    def __get_token_from_env(self) -> None:
-        debug(f"1. Trying to authenticate via `{self.GITHUB_TOKEN_ENV_VAR}` environment variable...")
-        github_token = os.environ.get(self.GITHUB_TOKEN_ENV_VAR)
+    @property
+    def provider(self) -> Optional[str]:
+        return self.__provider
+
+    @classmethod
+    def from_domain(cls, domain: str) -> Optional["GitHubToken"]:
```
PawelLipski: Nit: I'd rather say
```suggestion
    def for_domain(cls, domain: str) -> Optional["GitHubToken"]:
```

git_machete/github.py
PawelLipski: Whoops ☹️

PawelLipski (approved):
_dobra zmiana_

git_machete/github.py
```diff
@@ -86,47 +86,54 @@ def from_url(cls,
...
+                   organization=None,
+                   repository=None)
 
 
 class GitHubToken:
     GITHUB_TOKEN_ENV_VAR = 'GITHUB_TOKEN'
 
     def __init__(self,
-                 domain: str
+                 value: Optional[str],
```
PawelLipski: Hmmm actually... shouldn't these properties be non-nullable tho? 🤔 is there any case where we'd want to pass `None` as either of these? See the original issue #819

PawelLipski (changes requested):
As in the comments

git_machete/github.py
amalota: Yes, after the changes these values are non-nullable

amalota:
@PawelLipski

PawelLipski (approved):

PR #828, opened 2023-02-21
Deprecate `machete.overrideForkPoint.<branch>.whileDescendantOf` key

### LINKED ISSUE #611: `whileDescendantOf` config for fork point override isn't just redundant, it's probably erroneous
Repro:
1. 2 unique commits on a branch Y
2. delete its base branch X
3. override fork-point for Y to Y~2
4. `git machete squash` while on Y
5. fork point override to Y~2 no longer applies, due to `whileDescendantOf` config, even though that's still logically the fork point (and is pointed by `to`)

Time to deprecate the setting: newer version of git-machete doesn't need to take this config into account at all. To retain backward compatibility (esp. with IntelliJ plugin, rather than earlier versions of the CLI), let's still set the `whileDescendantOf` to the same commit as `to`.

PawelLipski:
Closes #611
See https://github.com/VirtusLab/git-machete-intellij-plugin/pull/1581


amalota (approved):
Looks good

PR #860, opened 2023-03-17
Pass `--empty=drop` to the underlying `git rebase`

### LINKED ISSUE #853: Underlying `git rebase` fails on effectively-empty commits
Probably passing `--empty=drop` should solve the issue. See https://github.com/VirtusLab/git-machete-intellij-plugin/issues/969 for details. Note that the problem only occurs in case of `--interactive` rebases: https://git-scm.com/docs/git-rebase#Documentation/git-rebase.txt---emptydropkeepask

PawelLipski:


git_machete/git_operations.py
```diff
@@ -918,7 +918,7 @@ def do_rebase() -> None:
             try:
                 if not opt_no_interactive_rebase:
                     rebase_opts.append("--interactive")
-                self._run_git("rebase", *rebase_opts, "--onto", onto, from_exclusive, branch)
+                self._run_git("rebase", *rebase_opts, "--empty=drop", "--onto", onto, from_exclusive, branch)
```
PawelLipski: The flag is only to be included if the underlying git supports it (version 2.26.0 or later)

amalota: @PawelLipski i see, so the solution you proposed would to be execute the tests only if the version supports the flag as you implemented in 'test_update_drops_empty_commits()'. But i see that there are a lot more tests that do not pass after the changes and they would't be checked for most of the checks if i were to use the same solution, is that acceptable? Because if not than maybe i can mock `rebase()` for theses tests like this:
```
git = GitContext()
if git.get_git_version() < (2, 26, 0):
    mocker.patch('git_machete.git_operations.rebase', old_rebase_with_no_empty_drop_flag)
``` 
wdyt?

PawelLipski: > But i see that there are a lot more tests that do not pass after the changes and they would't be checked for most of the checks if i were to use the same solution, is that acceptable?

Hmmm could you specify what tests fail and with what error? 🤔

> would to be execute the tests only

Actually,
1. to execute tests only for git >= 2.26.0, but also
2. to pass the flag to `git rebase` only if git >= 2.26.0 ‼️

tests/test_update.py
```diff
@@ -52,6 +54,44 @@ def test_update_with_fork_point_not_specified(self, mocker: Any) -> None:
              "'git rebase' to the parent branch of the current branch."
              )
 
+    def test_update_drops_empty_commits(self, mocker: Any) -> None:
+        """Verify behaviour of a 'git machete update' command.
+
+        Verify that 'git machete update' drops effectively-empty commits.
+        """
+        git = GitContext()
+        if git.get_git_version() < (2, 26, 0):
```
PawelLipski: Actually... maybe even I'd handle the case of git < 2.26.0 separately in the test. `git machete update` should fail, but then after `git rebase --continue`, the desired state should still be achieved 🤔

git_machete/git_operations.py
```diff
@@ -918,7 +918,10 @@ def do_rebase() -> None:
             try:
                 if not opt_no_interactive_rebase:
                     rebase_opts.append("--interactive")
-                self._run_git("rebase", *rebase_opts, "--empty=drop", "--onto", onto, from_exclusive, branch)
+                if self.get_git_version() < (2, 26, 0):
```
PawelLipski: Nit: commit message is incorrect:

image

Should be `>=`, not `>`

git_machete/git_operations.py
```diff
@@ -918,7 +918,10 @@ def do_rebase() -> None:
...
+                    self._run_git("rebase", *rebase_opts, "--onto", onto, from_exclusive, branch)
+                else:
+                    self._run_git("rebase", *rebase_opts, "--empty=drop", "--onto", onto, from_exclusive, branch)
```
PawelLipski: I'd just use `rebase_opts.append(...)`, as in line 920 above (instead of repeating all params twice)

PR #864, opened 2023-03-30
Do not assume that local branch and its remote counterpart has the same name when using `anno-prs`

### LINKED ISSUE #852: `github anno-prs`: remote tracking branch might have a different name than the local branch
`github anno-prs` assumes the name of local and remote branch is the same... and puts the warning `WRONG PR BASE or MACHETE PARENT?` unnecessarily

amalota:


git_machete/client.py
```diff
@@ -1538,11 +1538,6 @@ def __sync_annotations_to_definition_file(self,
                 anno: str = f'PR #{pr.number}'
                 if pr.user != current_user:
                     anno += f' ({pr.user})'
-                upstream: Optional[LocalBranchShortName] = self.up_branch.get(LocalBranchShortName.of(pr.head))
-                if pr.base != upstream:
-                    warn(f'branch {bold(pr.head)} has a different base in PR #{bold(str(pr.number))} ({bold(pr.base)}) '
-                         f'than in machete file ({bold(upstream) if upstream else "<none, is a root>"})')
-                    anno += f" WRONG PR BASE or MACHETE PARENT? PR has {pr.base}"
```
PawelLipski, line 1545: Hmmm okay but in general we want this case to be covered 🤔 it's just that we want to check for the tracking branch name first before doing the check

PawelLipski: Specifically, we want to extract the case like:
* PR in GitHub is from `foo` to `bar`
* locally, there's `baz` branch whose remote tracking branch is called `origin/bar`
* machete file is just `foo` under `baz`

Now if we weren't checking remote tracking branches, we'd say that `foo` has a wrong PR base or machete parent (`bar` vs `baz`)... while in fact everything's in sync.

amalota: @PawelLipski, what do you mean by tracking branch?

PawelLipski: `git_machete.git_operations.GitContext.get_combined_counterpart_for_fetching_of_branch` ;)

git_machete/client.py
```diff
@@ -1539,7 +1539,14 @@ def __sync_annotations_to_definition_file(self,
                 if pr.user != current_user:
                     anno += f' ({pr.user})'
                 upstream: Optional[LocalBranchShortName] = self.up_branch.get(LocalBranchShortName.of(pr.head))
-                if pr.base != upstream:
+                if upstream is not None:
+                    assert upstream is not None  # Cast away 'Optional'
```
PawelLipski: Hmm IIRC from how mypy behaves... this shouldn't be necessary (i.e. mypy should "remember" that `upstream` is non-None after the `if`-check)

RELEASE_NOTES.md
```diff
@@ -1,5 +1,7 @@
 # Release notes
 
+## New in git-machete 3.16.3
+
```
PawelLipski, line 8: Pls add an entry to release notes (plus I've released 3.16.3 in the meantime + moved this one to 3.16.4)

tests/mockers.py
```diff
@@ -95,9 +95,10 @@ def commit_amend(self, message: str) -> "GitRepositorySandbox":
         self.execute(f'git commit --amend -m "{message}"')
         return self
 
-    def push(self, remote: str = 'origin', set_upstream: bool = True) -> "GitRepositorySandbox":
+    def push(self, remote: str = 'origin', set_upstream: bool = True, tracking_branch: Optional[str] = None) -> "GitRepositorySandbox":
         branch = popen("git symbolic-ref -q --short HEAD")
-        self.execute(f"git push {'--set-upstream' if set_upstream else ''} {remote} {branch}")
+        tracking_branch = popen("git symbolic-ref -q --short HEAD") if not tracking_branch else tracking_branch
```
PawelLipski: Huh that's the same `popen` as in the line above 🤔

git_machete/client.py
amalota: @PawelLipski I though so too but unfortunately `assert ...` is necessary

tests/mockers.py
amalota: Yeah, nice catch, will fix

tests/test_github.py
```diff
@@ -1979,3 +1979,58 @@ def test_get_token_from_hub(self, mocker: Any) -> None:
...
+                .commit('introduce feature')
+                .push(tracking_branch='feature_repo')
+                .new_branch('feature_1')
+                .commit('introduce feature')
+                .push()
+                .delete_branch('feature_repo')
+                .add_remote('new_origin', 'https://github.com/user/repo.git')
+        )
+
+        launch_command("discover", "-y")
```
PawelLipski: General remark: I'd avoid using `discover` in tests, unless `discover` itself is tested.

1. This makes analyzing the test harder as you can't see what's the branch layout directly from the test
2. This decreases the performance of the tests without increasing the coverage or any other major benefit

tests/test_github.py
```diff
@@ -1979,3 +1979,58 @@ def test_get_token_from_hub(self, mocker: Any) -> None:
...
+        launch_command("anno", "-H")
```
PawelLipski:
```suggestion
        launch_command("github", "anno-prs")
```

That's equivalent... but let's stick to `github` command since this is `test_github.py`

tests/mockers.py
```diff
@@ -95,9 +95,10 @@ def commit_amend(self, message: str) -> "GitRepositorySandbox":
         self.execute(f'git commit --amend -m "{message}"')
         return self
 
-    def push(self, remote: str = 'origin', set_upstream: bool = True) -> "GitRepositorySandbox":
+    def push(self, remote: str = 'origin', set_upstream: bool = True, tracking_branch: Optional[str] = None) -> "GitRepositorySandbox":
         branch = popen("git symbolic-ref -q --short HEAD")
-        self.execute(f"git push {'--set-upstream' if set_upstream else ''} {remote} {branch}")
+        tracking_branch = branch if not tracking_branch else tracking_branch
```
PawelLipski:
```suggestion
        tracking_branch = tracking_branch or branch
```

git_machete/client.py
PawelLipski: Hmmm I can't reproduce that, at least under Python 3.9... even without that `assert`, `mypy` passes for me. What's the error you're getting/under what Python version?

amalota: Wow, i don't know how but it passes now, i don't remember which Python version it was

PawelLipski (approved):

PR #875, opened 2023-04-04
Revamp `hook_samples/machete-post-slide-out` to take advantage of `git machete github retarget-pr`

### LINKED ISSUE #596: Revisit `hook_samples/machete-post-slide-out`
It's been neglected, mostly coz I relied on GitHub automatically retargeting the child PRs once parent PR is merged... but I've recently realized that this behavior only applies when the setting 

image

is enabled. 

This is, however, **not** the default in GitHub... and there are certain cases where it **cannot** be enabled (e.g. cloud gitops tool [ArgoCD](https://argo-cd.readthedocs.io/en/stable/) points to certain branches from its config). 

As a first step, let's research when actually GitHub retargets PRs automatically:
- [ ] is this behavior officially documented anywhere?
- [ ] does this behavior have anything to do with closing a PRs for the base branch, or just solely with deleting the base branch? it seems that the latter
- [ ] does GitHub take PR structure into consideration when retargeting, or just retargets to repos' default branch (typically `master`/`main`)? i.e. assuming `A` is **not** the default branch, in case of `A <- B <- C`, when `B` is deleted... is PR for `C` retargeted to `A` or to the default branch?

The hook should be revisited... in particular:
- [ ] the docs about the hook could be updated to elaborate on this GitHub behavior
- [ ] the sample hook itself should be simplified to use `git machete github retarget-pr` instead of some low-level hacks with `curl`

PawelLipski:


hook_samples/machete-post-slide-out
```diff
@@ -13,34 +13,8 @@
...
-    -H "Content-Type: application/vnd.github.v3+json" \
-    "https://api.github.com/repos/$org_and_repo/pulls/$pr_number" \
-    -d "{ \"base\": \"$new_upstream\" }" \
-    --fail \
-    --silent \
-    --show-error \
-    -o/dev/null \
-    -w " HTTP status %{http_code}\n"
+for new_downstream in "${new_downstreams[@]}"; do
+  git machete github retarget-pr --ignore-if-missing "$new_downstream"
```
PawelLipski: Requires #876 and #877

PR #883, opened 2023-04-07
Replace usage of `git machete discover` in tests with explicit machete file definition

### LINKED ISSUE #873: Replace executions of `git machete discover -y` in tests with saving the branch layout directly
General remark: I'd avoid using `discover` in tests, unless `discover` itself is tested.

1. This makes analyzing the test harder as you can't see what's the branch layout directly from the test
2. This decreases the performance of the tests without increasing the coverage or any other major benefit

_Originally posted by @PawelLipski in https://github.com/VirtusLab/git-machete/pull/864#discussion_r1157328153_

amalota:


tests/test_github.py
```diff
@@ -2084,7 +2216,6 @@ def test_local_branch_name_different_than_tracking_branch_name(self, mocker: Any
                 feature
```
PawelLipski, line 2213: There are 3 occurrences left in this file:

```
tests/test_github.py:1981:        launch_command('discover', '--checked-out-since=1 day ago')
tests/test_github.py:1989:        launch_command('discover', '--checked-out-since=1 day ago')
tests/test_github.py:1994:        launch_command('discover', '--checked-out-since=1 day ago')
```

Probably to be replaced as well

tests/test_github.py
```diff
@@ -342,7 +348,14 @@ def test_github_retarget_pr_multiple_non_origin_remotes(self, mocker: Any) -> No
                 .push(remote='origin_2', set_upstream=False)
```
PawelLipski, line 347: Nit: a stray `dedent` left in line 282:

```python
        body = dedent(body)
        rewrite_definition_file(body)
```

tests/test_github.py
amalota: Yeah, first two instances can be replaced but the last one is used to check if the result of the executed commands is correct.

PawelLipski (approved):

PawelLipski:
Btw, while you're here... pls check if there's a visible change in test execution time before and after this change (I suppose it should be observable) 🤔

amalota:
@PawelLipski 
```
env      before(s)     after
py311      73            71  
py 310     80            68
py39       95            82
```

PR #889, opened 2023-04-14
Do not checkout branches that are `untracked/ahead of remote/diverged from remote`  in `traverse --no-push` mode 

### LINKED ISSUE #855: `traverse`: even in `--no-push` mode, branches that are ahead of/diverged from remote are checked out
Not a major problem... it's just redundant to check out these branches only to step over them with no action.

amalota:


git_machete/client.py
```diff
@@ -731,6 +731,8 @@ def traverse(
                 needs_remote_sync = True
                 if branch in self.annotations:
                     needs_remote_sync = self.annotations[branch].qualifiers.push
+                if not opt_push_tracked and not opt_push_untracked:
```
PawelLipski, line 734: Pls add a test that would capture the difference b/w the old and the new behavior

amalota:
It does not seem to be worth mentioning in the release notes, @PawelLipski wdyt?

PawelLipski:
>  It does not seem to be worth mentioning in the release notes

Yeah, let's skip that from release notes, that's a very subtle change

PawelLipski (approved):

PR #900, opened 2023-04-24
Pass `--force-if-includes` to `git push` alongside `--force-with-lease` if git >= 2.30.0

### LINKED ISSUE #329: Consider adding `--force-if-includes` to all `git push` invocations if git version >= 2.30
Thanks @migmit for pointing that out!

PawelLipski:


debian/source/options
```diff
@@ -1,4 +1,5 @@
 tar-ignore = AUTHORS
+tar-ignore = BADGES.md
```
PawelLipski, line 2: unrel ofc ;)

PR #913, opened 2023-04-27
Mock `ask_if` in a few tests instead of passing `-y` flag to improve coverage

PawelLipski:


PawelLipski:
Whoops no, zero change in coverage :trollface:

PR #920, opened 2023-05-02
Fix use from within venvs on Windows

### LINKED ISSUE #919: Use entry_points.txt to declare bin entrypoint, to avoid breaking in venvs
# Overview

When installing git-machete globally via pip on Windows, and then working on a python project in a virtualenv, `git machete` fails with a cryptic error: 

```
(.venv) PS C:\Users\cspot\Documents\dev\project> git machete
  File "C:\Users\cspot\scoop\apps\python\current\Scripts\git-machete", line 19, in <module>
    from git_machete import cli
ModuleNotFoundError: No module named 'git_machete'
```

I suspect this is because git-machete does not allow pip to create an `.exe` shim like it does for most other packages.  So `git` runs the `git-machete` script using the venv's `python.exe`, ignoring the shebang.  Thus `sys.path` is still pointing to the virtual env where the `git_machete` module is _not_ installed.

I think the fix is to use `entry_points.txt` to tell pip and other package managers how to create the necessary .exe shim on Windows.

https://packaging.python.org/en/latest/specifications/entry-points/

`git machete` works within a virtual env shell, calling out to the globally installed git-machete.

The error I pasted above

```
git --version
git version 2.38.0.windows.1
git machete --version
git machete version 3.17.1
```

## Additional diagnostics info

Here is one way to reproduce this issue on Windows.  Open Powershell:

```
# run in powershell
pip install git-machete
git machete # this should work
mkdir project
cd project
python -m venv .venv
./.venv/Scripts/activate.ps1
git machete # this should fail with the error described above
```

If you modify the `git-machete` script to `print(sys.executable)` you will see that git is executing `git-machete` with the venv's `python.exe`.

cspotcode:
Fixes #919


PawelLipski:
Thanks @cspotcode for pushing the fix! I've changed the base to `develop` (see [FAQ](https://github.com/VirtusLab/git-machete/blob/master/CONTRIBUTING.md#faq-about-pull-requests)). This should fix some of the CI issues.

Also, pls add the following line:
```ini
exclude = git_machete/bin.py
```
to mypy.ini ✍🏻

PawelLipski:
Oh, sorry, I forgot to ask: pls rebase your changes on top of `develop`. Both mypy.ini and git-machete were modified in the meantime (b/w `master` and `develop`)... but these are rather simple changes:
* mypy.ini: entries were sorted in abc order
* git-machete: indent has been changed from 2 to 4 spaces to make flake8 happy

cspotcode:
Thanks, flake8 is complaining about the use of `.format` in `bin.py`, but I'm guessing we want to keep `.format` so that bin.py is compatible with older python versions?  Otherwise users would not get the helpful diagnostic message.

I'm unfamiliar with flake8, what's the best way to suppress that lint failure?

PawelLipski:
> but I'm guessing we want to keep .format so that bin.py is compatible with older python versions?

Lemme check quickly for the workaround 👁️

cspotcode:
I think I fixed it: `# flake8: noqa`

PawelLipski:
> I think I fixed it: `# flake8: noqa`

Yeah, I was checking `# noqa: FS002` but it's pretty much the same

PawelLipski:
Okay! ready to go, I'm merging soon (and release v3.7.2 later in the day) 🚢

PawelLipski (approved):

PR #929, opened 2023-05-05
Use sh instead of bash in tests

Ikke:
8adfd5f (Add missing tests for yellow-edge cases, among others, 2023-04-29) and 27183a2 (Add missing tests - vol. 3, 2023-05-01) adds tests that refer to bash, while it does not use any functionality that requires bash. This results in test failures when bash is not installed.

Use sh instead which is more commonly available and more portable.

### DIFF
--- a/tests/test_edit.py
+++ b/tests/test_edit.py
@@ -1,7 +1,7 @@
 from .base_test import BaseTest
 from .mockers import assert_success, launch_command, overridden_environment
 
-dummy_editor = "bash -c 'echo foo > $1' 'ignored_$0'"
+dummy_editor = "sh -c 'echo foo > $1' 'ignored_$0'"
 
 
 class TestEdit(BaseTest):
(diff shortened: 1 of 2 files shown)


PawelLipski:
Wow indeed... how have you discovered that? 😯

PawelLipski (approved):

Ikke:
> Wow indeed... how have you discovered that? 😯

On Alpine Linux bash is not installed unless explicitly added as a dependency, so I had 2 test failures when upgrading the git-machete package.

PawelLipski:
Okay... I wasn't aware that Alpine package for git-machete runs the tests on installation. Thanks for the fix, I'll get this released as v3.17.3 soon

Ikke:
FYI, this is when building the package, not when installing it, but the builders only have a minimum of packages installed and everything else needs to be explicitly added as a dependency, so these kinds of issues pop up (though I already ran into it when building it locally).

PR #968, opened 2023-06-12
Make lint-no-dead-urls check non-blocking

PawelLipski:


hook_samples/machete-post-slide-out
```diff
@@ -11,7 +11,7 @@
 
 [[ -n "$1" || -n "$2" ]] || { echo "usage: $(basename "$0") <new-upstream> <slid-out-branch> [<new-downstreams>...]"; exit 1; }
 
-set -e -o pipefail -u
```
PawelLipski, line 14: `-u` causes a failure on `${new_downstreams[@]}` if it's empty in case of older bash versions (see https://stackoverflow.com/a/7577209)

PR #990, opened 2023-06-26
Get rid of kwargs in popen/run functions

PawelLipski:


git_machete/git_operations.py
```diff
@@ -218,23 +218,20 @@ def flush_caches(self) -> None:
...
+                f"`{utils.get_cmd_shell_repr('git', git_cmd, *args, env=None)}` returned {exit_code}")
         return exit_code
 
-    def _popen_git(self, git_cmd: str, *args: str, **kwargs: Any) -> CommandResult:
-        kwargs_ = kwargs.copy()
-        allow_non_zero = kwargs_.pop("allow_non_zero", False)
-        exit_code, stdout, stderr = utils.popen_cmd("git", git_cmd, *args, **kwargs_)
+    def _popen_git(self, git_cmd: str, *args: str,
+                   allow_non_zero: bool = False, env: Optional[Dict[str, str]] = None) -> CommandResult:
+        exit_code, stdout, stderr = utils.popen_cmd("git", git_cmd, *args)
```
PawelLipski: Whoops env **not** passed... let's add a test that would've detected it (easiest: for `squash`, check if the author of the squashed commit is as expected)

PR #1045, opened 2023-09-05
Fix reading tokens from `~/.github-token` for GitHub Enterprise domains

PawelLipski:


git_machete/github.py
```diff
@@ -102,7 +103,7 @@ def __get_token_from_file_in_home_directory(cls, domain: str) -> Optional["GitHu
                 # ghp_yetanothertoken_for_git_example_com git.example.com
 
                 for line in file.readlines():
-                    if line.endswith(" " + domain):
+                    if line.rstrip().endswith(" " + domain):
```
PawelLipski, line 106: That was the actual bug, `readlines()` returns lines with trailing `\n`...

tests/test_github.py
```diff
@@ -227,9 +227,7 @@ def test_github_get_token_from_file_in_home_directory(self, mocker: MockerFixtur
         github_token_contents = ('ghp_mytoken_for_github_com\n'
                                  'ghp_myothertoken_for_git_example_org git.example.org\n'
                                  'ghp_yetanothertoken_for_git_example_com git.example.com')
-        _mock_open = mock_open(read_data=github_token_contents)
-        _mock_open.return_value.readlines.return_value = github_token_contents.split('\n')
```
PawelLipski, line 231: ...but the test didn't capture that, coz `split('\n')` removes `\n`, unlike `readlines()`.

In fact, this extra mocking of `readlines` was unnecessary, `mock_open` already fills up methods like `readlines`.

tests/test_github.py
```diff
@@ -238,12 +236,20 @@ def test_github_get_token_from_file_in_home_directory(self, mocker: MockerFixtur
         assert github_token.provider == f'auth token for {domain} from `~/.github-token`'
         assert github_token.value == 'ghp_mytoken_for_github_com'
 
+        # Line ends with \n
```
PawelLipski, line 239: This was the test case that would've captured the bug.

tests/test_github.py
```diff
@@ -238,12 +236,20 @@ def test_github_get_token_from_file_in_home_directory(self, mocker: MockerFixtur
...
         domain = 'git.example.org'
         github_token = GitHubToken.for_domain(domain=domain)
         assert github_token is not None
         assert github_token.provider == f'auth token for {domain} from `~/.github-token`'
         assert github_token.value == 'ghp_myothertoken_for_git_example_org'
 
+        # Last line, doesn't end with \n
```
PawelLipski, line 246: Just an extra added test case, was already passing.

PR #1056, opened 2023-09-23
Use GitHub PR template

raylu:
GH repos can have multiple PR templates; this uses the default one if it's available.

`.git/info/description` is checked first and is used if present (even if empty).


PawelLipski:
Nice, good idea :) I'm gonna bump the version + add test coverage, and then release :)

PawelLipski:
Merged as #1058 

To be released soon as v3.19.0

Thanks for the contribution! 🙇🏻

PR #1058, opened 2023-09-23
Use GitHub PR template

PawelLipski:
#1056 + some changes of mine


git_machete/client.py
```diff
@@ -2338,8 +2338,15 @@ def create_github_pr(
         debug(f'organization is {org_repo_remote.organization}, repository is {org_repo_remote.repository}')
         debug('current GitHub user is ' + (current_user or '<none>'))
 
-        description_path = self.__git.get_main_git_subpath('info', 'description')
-        description: str = utils.slurp_file_or_empty(description_path)
+        machete_description_path = self.__git.get_main_git_subpath('info', 'description')
+        # https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/creating-a-pull-request-template-for-your-repository
+        github_template_path = os.path.join(self.__git.get_root_dir(), '.github', 'pull_request_template.md')
+        if os.path.isfile(machete_description_path):
```
PawelLipski, line 2344: Split into individual if-else'd cases to better track test coverage

git_machete/client.py
```diff
@@ -2356,7 +2363,10 @@ def create_github_pr(
         print(fmt(f'{ok_str}, see `{pr.html_url}`'))
 
         milestone_path: str = self.__git.get_main_git_subpath('info', 'milestone')
-        milestone: str = utils.slurp_file_or_empty(milestone_path).strip()
```
PawelLipski, line 2359: Same reason, to easily check if both the case with existing and non-existent file is covered with tests

Issue #98, opened 2020-07-25
Add support for automatic branch deletion on slide-out


PawelLipski:
As implemented in git-machete-intellij-plugin (see https://github.com/VirtusLab/git-machete-intellij-plugin/pull/336):

```
git config machete.slideOut.deleteLocalBranch true
```

PR #1094, opened 2023-11-18
Add `--all-merged` flag to `slide-out`

### LINKED ISSUE #1051: Add `--untracked` option to `slide-out`
Slide out all untracked branches. Replacement for functionality `clean`.

Note that `clean` removes untracked & **childless** branches.
`slide-out --untracked` should slide out (not remove!) branches.
TBD if the condition of childlessness should be retained.

Hmmm removing on the childless branches has the funny side effect: in case of chains of untracked branches, subsequent calls to `slide-out --untracked` will "peel off" the down-most layer 🤔

raylu:
see discussion in https://github.com/VirtusLab/git-machete/discussions/1088

should we call this `delete-merged` instead? might be too confusing to have `clean` and `prune`


tests/test_prune.py
```diff
@@ -0,0 +1,75 @@
...
+            main
+                unpushed
+                not_deleted_remotely
+                has_downstream
+                    downstream
+                should_be_pruned
+            """
+        rewrite_branch_layout_file(body)
+
+        launch_command('prune')
```
raylu: how come I don't need to pass `--yes` here? `test_clean.py` doesn't do it either

docs/source/short_docs.rst
```diff
@@ -20,6 +20,7 @@
 * :ref:`is-managed`       -- Check if the current branch is managed by git machete (mostly for scripts)
 * :ref:`list`             -- List all branches that fall into one of pre-defined categories (mostly for internal use)
 * :ref:`log`              -- Log the part of history specific to the given branch
+* :ref:`prune`            -- Delete local branches that have been deleted remotely
```
PawelLipski: There's even been `prune-branches` a long time ago (v1.x), later removed in favor of `delete-unmanaged` in the 1.x->2.0 version bump, as a more precise alternative 😅 so yeah, indeed I'd rather **not** use this name

git_machete/cli.py
```diff
@@ -39,7 +39,7 @@
     ("Build, display and modify the tree of branch dependencies",
      ["add", "anno", "discover", "edit", "status"]),
     ("List, check out and delete branches",
-     ["delete-unmanaged", "go", "is-managed", "list", "show"]),
+     ["clean", "delete-unmanaged", "go", "is-managed", "list", "prune", "show"]),
```
PawelLipski: `clean` is left out deliberately... it's not yet formally deprecated, but I'll deprecate it soon (and later remove in 3.x->4.0 transition, if it ever happens 😅)

tests/test_prune.py
PawelLipski: Thanks to 

```python 
self.patch_symbol(mocker, 'builtins.input', mock_input_returning_y)
```

;)

git_machete/git_operations.py
```diff
@@ -544,8 +553,6 @@ def __load_branches(self) -> None:
 
         for line in raw_local:
             values = line.split("\t")
-            if len(values) != 5:
-                continue  # pragma: no cover; invalid, shouldn't happen
```
PawelLipski, line 548: Hmmm removed? 🤔 there's ofc also an argument to **not** silently skip occurrences like that &mdash; after all, I'd better get an issue on github if it ever happens 🤔

git_machete/git_operations.py
```diff
@@ -198,6 +198,7 @@ def __init__(self) -> None:
         self.__is_equivalent_tree_reachable_cached: Dict[Tuple[FullCommitHash, FullCommitHash], bool] = {}
         self.__local_branches_cached: Optional[List[LocalBranchShortName]] = None
         self.__merge_base_cached: Dict[Tuple[FullCommitHash, FullCommitHash], Optional[FullCommitHash]] = {}
+        self.__missing_tracking_branch: Optional[set[str]] = None
```
PawelLipski: Nitpick:
```suggestion
        self.__missing_tracking_branch: Optional[Set[str]] = None
```
for consistency (if `mypy` accepts that)

git_machete/client.py
```diff
@@ -1182,6 +1182,23 @@ def delete_unmanaged(self, *, opt_yes: bool) -> None:
         branches_to_delete = excluding(self.__git.get_local_branches(), self.managed_branches)
         self.__delete_branches(branches_to_delete=branches_to_delete, opt_yes=opt_yes)
 
+    def __delete_branches_rearrange_layout(self, branches_to_delete: List[LocalBranchShortName], opt_yes: bool) -> None:
```
PawelLipski: Hmmm AFAICS this command first slides out, then deletes the branches... how about packing that functionality into `slide-out --merged [--delete]`? I can see the following benefits:

1. Consistent behavior &mdash; slide-out serves the purpose of removing the branches from `.git/machete`; then whether to delete them can be controlled with the existing `--delete` flag
2. Actually, deleting branches right after sliding out, while ofc possible with `--delete` flag (let alone with `git branch -d`), is kind of discouraged due to the reasons outlined in [FAQ](https://github.com/VirtusLab/git-machete#sometimes-when-i-run-update-or-traverse-too-many-commits-are-taken-into-the-rebase-how-to-fix-that). So better not make it the default behavior straight away 😅 
3. Arguably less CLI bloat 😅 git-machete got already has (too) many top-level commands, it feels better to keep related stuff in long-established commands like `slide-out` where the users can expected them,

PawelLipski:
Nice one, thanks :)

docs/source/short_docs.rst
raylu: suggestions? `delete-merged`?

EDIT: oh, I see below, `slide-out --merged [--delete]`

git_machete/git_operations.py
raylu: yeah. it "shouldn't happen" so it seems preferable to error rather than fail silently

git_machete/git_operations.py
raylu: ah, that should fix py 3.6-3.8

git_machete/client.py
raylu: one awkward thing about `slide-out --merged` is, if you don't pass `--delete`, it will remove the branch from the layout. if you realize you wanted the branches deleted, you can't immediately `slide-out --merged --delete` because those branches are now unmanaged

to automatically find and delete them at this point you probably have to `discover` and then `slide-out --merged --delete` or something

I'm not opposed to it if that's what you want. whatever invocation ends up doing this has a good chance of ending up as an alias for me anyway

PawelLipski: > to automatically find and delete them at this point you probably 

Note that there's `git machete delete-unmanaged` (already mentioned in this conversation)... and there's also a long-standing task (https://github.com/VirtusLab/git-machete/issues/98) to implement support for `git config machete.slideOut.deleteLocalBranch true` that would delete branches on slide-out by default. That's actually one of the few features present in [git-machete-intellij-plugin](https://github.com/VirtusLab/git-machete-intellij-plugin) but not in this CLI 😅 

So yeah, if you could proceed with `slide-out --merged`... I can smooth out all the details like shell completions etc. at the end ✍🏻

tests/test_prune.py
```diff
@@ -0,0 +1,75 @@
...
+            """
+        )
+        assert_success(['status'], expected_status_output)
+
+        branches = self.repo_sandbox.get_local_branches()
+        assert 'unmanaged' in branches
+        assert 'unpushed' in branches
+        assert 'not_deleted_remotely' in branches
+        assert 'has_downstream' in branches
+        assert 'sould_be_pruned' not in branches
```
raylu: oof there's a typo here so this always passes

git_machete/client.py
raylu: I went with `slide-out --all-merged` since there's already a `slide-out --merge` flag

it's... kinda awkward though. there are a bunch of other args to slide-out that don't make any sense. the whole pass a chain thing doesn't apply here either

docs/source/cli/slide-out.rst
```diff
@@ -6,9 +6,11 @@ slide-out
 
 .. code-block:: shell
 
-    git machete slide-out [-d|--down-fork-point=<down-fork-point-commit>] [--delete] [-M|--merge] [-n|--no-edit-merge|--no-interactive-rebase] [<branch> [<branch> [<branch> ...]]]
+    git machete slide-out [-d|--down-fork-point=<down-fork-point-commit>] [--delete] [--all-merged] [-M|--merge] [-n|--no-edit-merge|--no-interactive-rebase] [<branch> [<branch> [<branch> ...]]]
```
PawelLipski, line 9: Makes sense to use `--all-...` prefix indeed to avoid confusion with `--merge`, good spot

PawelLipski:
Thanks for all your work <3 I can see some minor fixes that need to be applied, but I'll take over from now on. I'll get this merged on a separate PR (given that I can't push to your branch directly) but AFAIK your contribution will be included in the stats nonetheless.

git_machete/client.py
PawelLipski: I'm gonna put in in docs as another flavor of slide-out, like

```
    git machete slide-out --all-merged [--delete]
    git machete slide-out [-d|--down-fork-point=<down-fork-point-commit>] [--delete] [-M|--merge] [-n|--no-edit-merge|--no-interactive-rebase] [<branch> [<branch> [<branch> ...]]]
```

similarly to `fork-point` which has a couple flavors as well

PawelLipski:
Merged at #1103.

PR #1116, opened 2023-11-28
Throw an `UnexpectedMacheteException` in cases that are truly unexpected

### LINKED ISSUE #1104: Review all `# pragma: no cover` and `# pragma: no branch` cases, possibly throw exceptions there if they're truly unexpected
Plus add `GITHUB_NEW_ISSUE_MESSAGE` to the error message.

Same for the places that aren't covered by tests (just a handful lines).

PawelLipski:


git_machete/github.py
```diff
@@ -247,10 +250,12 @@ def __fire_github_api_request(self, method: str, path: str, request_body: Option
                 if link_header:
                     url_prefix_regex = re.escape(url_prefix)
                     match = re.search(f'<{url_prefix_regex}(/[^>]+)>; rel="next"', link_header)
-                    if match:  # pragma: no branch; there should always be a match
```
PawelLipski, line 250: Whoops this wasn't a correct assumption, see #1126

Issue #769, opened 2023-01-05
Should qualifiers like `push=no`, `rebase=no`, `slide-out=no` apply beyond `traverse`?


PawelLipski:
In particular:
* `push=no`: `advance`?
* `rebase=no`: `reapply`, `slide-out`, `update`?
* [`slide-out=no` once implemented](https://github.com/VirtusLab/git-machete/issues/749): `advance`, `slide-out`?

PawelLipski:
+1 for applying the qualifiers (i.e. skipping the operation) where the operation is **incidental** to what the command does:
* `push=no`: `advance`
* `rebase=no`: `slide-out`
* `slide-out=no`: `advance`

Not sure what to do when the operation is the **core functionality** of the command:
* `rebase=no`: `reapply`,  `update`?
* `slide-out=no`: `slide-out`?

Probably displaying a warning and still skipping the operation? Or prompting the user whether they really want to proceed?

PR #1209, opened 2024-03-25
Allow setting per-branch update=merge option

gjulianm:
This PR adds support for a `update=merge` annotation. This makes it so that `git machete traverse` or `git machete update` will switch to merging that branch if it has that annotation, without requiring the use of the `-M` option for all the branches.

The use case behind this is the following. As stated in the `git-machete` docs, rebase is usually the best option to maintain a clean history with branches, thus it makes sense to use it as a default. However, when a branch is already being reviewed in PRs, rebasing it can cause Github to lose track of what was reviewed and what changed after the reviews, so it makes sense to merge instead of rebase in those cases. With this PR this can be done automatically. This is something very useful for my workflow and I wanted to send a PR in case it is interesting to include in the upstream. I will probably work in the future on adding the annotation automatically based on the PR state, if you think it's interesting I can send that as a PR too (although maybe that PR is a little bit more opinionated and less useful to the general userbase).

### Notes 

I think I did not miss any place where merges can be optionally done, I searched for the usage of `opt_merge` too just in case. I added a test case to check that this was working, based on the test that uses the `-M` option. I had to remove the restriction on passing `--no-edit-merge` without `-M`, as now it does make sense to pass that option even if merge is not the default behavior.


PawelLipski:
Wow nice 😯 still, you know our stance on merges in general... they just don't work well with stacked branches 😅 but anyway, the options like `-M` are there for the specific cases/workflows where it makes sense.

One thing that I'd reconsider in this PR... I'm worried that calling this qualifier `merge=yes` will be confusing. The semantics of the existing `{push,rebase,slide-out}=no` qualifiers is that the given operation is **not** performed when running `git machete traverse`... which esp. matters in case of `traverse --yes`.

Along these lines... one could then propose a `merge=no` qualifier, which would make `traverse` skip the merges for the specific branch when `--merge` mode is selected.

Now, if we consider hypothetical `...=yes` qualifiers (which don't exist so far)... I'd say it'd be more logical for them to mean sth like _perform rebase/push/slide-out/merge even if this operation is disabled via some other mechanism_ 🤔

But! how about **renaming this qualifier to sth like `update=merge`**? There's `git machete update` command, coming from the very earliest days of git-machete... which just does the simple syncing of a branch to its parent via rebase or merge. Using a `update=merge` qualifier would logically mean that we let `traverse` perform update via merge, rather than the default rebase 🤔

PawelLipski:
(edited the description to fix a typo)

PawelLipski:
>  I will probably work in the future on adding the annotation automatically based on the PR state, if you think it's interesting I can send that as a PR too (although maybe that PR is a little bit more opinionated and less useful to the general userbase).

Wow I'm afraid indeed that this would be too opinionated to place directly in core git-machete 😅 this might however be added to the repo as a sample standalone script (git hook? see [hook_samples directory](https://github.com/VirtusLab/git-machete/tree/master/hook_samples))

git_machete/client.py
```diff
@@ -577,7 +578,8 @@ def slide_out(self,
         self.__git.checkout(new_upstream)
         for new_downstream in new_downstreams:
             self.__git.checkout(new_downstream)
-            if opt_merge:
+            use_merge = opt_merge or (new_downstream in self.annotations and self.annotations[new_downstream].qualifiers.merge)
```
PawelLipski, line 581: Actually... as for now, the qualifiers don't apply outside of `traverse`. There's an issue for that which should be rather quick to do: https://github.com/VirtusLab/git-machete/issues/769. I've even assigned it to the current milestone v3.25.0... there's just always something else that pops up 😅 currently I'm adding GitLab integration alongside GitHub

PawelLipski: Although... this qualifier has a different semantics anyway, so maybe leave as is indeed

gjulianm:
> But! how about **renaming this qualifier to sth like `update=merge`**?

It does make much more sense than `merge=yes` indeed. Changed and updated the PR :)

git_machete/annotation.py
```diff
@@ -9,15 +9,17 @@ def __init__(self, annotation: str):
...
+        self.__merge_text: str = ""
         self.rebase: bool = True
         self.push: bool = True
         self.slide_out: bool = True
+        self.merge: bool = False
 
-        def match_pattern(text: str) -> str:
-            return f'.*\\b{text}=no\\b.*'
+        def match_pattern(text: str, value: str = "no") -> str:
+            return f".*\\b{text}={value}\\b.*"
```
PawelLipski: TBH I'd just move the entire qualifier into the `text` &mdash; so that the calls look like `match_pattern("rebase=no")` or `match_pattern("update=merge")`.

Same for `def sub_pattern`

git_machete/annotation.py
```diff
@@ -9,15 +9,17 @@ def __init__(self, annotation: str):
         self.__push_text: str = ''
```
PawelLipski, line 9: If you could also bump version in `git_machete/__init__.py` and add an entry in `RELEASE_NOTES.md`, crediting yourself (see the previous entries)

git_machete/annotation.py
```diff
@@ -37,11 +39,17 @@ def sub_pattern(text: str) -> str:
...
+            self.merge = True
+            self.__merge_text = "update=merge"
+            self.__annotation_without_qualifiers = re.sub(sub_pattern('update', value='merge'), ' ', self.__annotation_without_qualifiers)
+
     def get_annotation_text_without_qualifiers(self) -> str:
         return self.__annotation_without_qualifiers.strip()
 
     def get_qualifiers_text(self) -> str:
-        return f'{self.__rebase_text} {self.__push_text} {self.__slide_out_text}'.replace('  ', ' ').strip()
+        return f'{self.__rebase_text} {self.__push_text} {self.__slide_out_text} {self.__merge_text}'.replace('  ', ' ').strip()
```
PawelLipski: Note that now you can end up with triple space in the output, so maybe use `re.sub(' +', ' ', ...)`

gjulianm:
Fixed your comments :) I bumped to v3.25.0 as it's a behavior change, not sure if changes in annotations are worth of a minor bump though, if you want a patch increase let me know.

RELEASE_NOTES.md
```diff
@@ -1,5 +1,9 @@
 # Release notes
 
+## New in git-machete 3.25.0
+
+- added: Annotation `update=merge` allows selecting merge strategy per branch (contributed by @gjulianm)
```
PawelLipski: Nit:
```suggestion
- added: qualifier `update=merge` allows selecting merge strategy per branch (contributed by @gjulianm)
```
(lower case + a different term, for consistency)

PawelLipski (approved):

PawelLipski (approved):
Gr8! thanks for your work, I'm gonna release that with 3.25.0 📦

gjulianm:
Nice! Thanks for being so responsive and also for making this tool, it's super useful for my workflow :D

PawelLipski:
Spread the word! btw, how did you learn of git-machete, was it via [Ben Congdon's post](https://benjamincongdon.me/blog/2022/07/17/In-Praise-of-Stacked-PRs/) or some other way?

gjulianm:
It was via a coworker, at my new company it seems there are some people who use it and it's fairly useful :D

PR #1220, opened 2024-04-15
Use pytest-cov over coverage.py to enable pytest-xdist

### LINKED ISSUE #1218: Enable pytest-xdist in coverage mode
It seems that the easiest way to do this would be to switch from vanilla coverage.py to pytest-cov (https://github.com/nedbat/coveragepy/issues/1341#issuecomment-1302863172).

Then, maybe even check if Windows tests can be run in coverage mode in CI?

PawelLipski:


.circleci/config.yml
```diff
@@ -184,7 +184,7 @@ jobs:
       - run: pip install tox
       # TODO (#1005): make zsh completion tests pass on macOS machine on CI
       - run: tox -e test-completions -- -vv -k "not zsh"
-      - run: PYTHON_VERSION=3.12-macos  tox -e coverage -- -vv
+      - run: PYTHON_VERSION=3-macos  tox -e coverage -- -vv
```
PawelLipski, line 187: image

Windows in 3:07, macOS in 2:30, both with coverage. Not bad! ([link](https://app.circleci.com/pipelines/github/VirtusLab/git-machete/4894/workflows/9de63785-5b24-4da4-ab71-54a0d99a5d02))

PR #1229, opened 2024-04-20
Fix incorrect re.sub invocation

zb3:
The fourth argument to `re.sub` is the `count` argument and it specifies the maximum number of pattern occurrences to be replaced.
So in this case, the `re.MULTILINE` flag is not used but instead interpreted as an integer (8) so only 8 replacements can be done.

BTW, `re.MULTILINE` only affects `^` and `$`, so I guess it's not useful in this case.

### DIFF
--- a/git_machete/utils.py
+++ b/git_machete/utils.py
@@ -101,7 +101,7 @@ def find_executable(executable: str) -> Optional[str]:
 
 
 def compact_dict(d: Dict[str, Any]) -> Dict[str, str]:
-    return {k: re.sub('\n +', ' ', str(v), re.MULTILINE) for k, v in d.items()}
+    return {k: re.sub('\n +', ' ', str(v)) for k, v in d.items()}
 
 
 def debug(msg: str) -> None:


PawelLipski:
Whoops... good spot 😯 how did you find that? this is only used in debug logging code... was it via some automated analysis?

PawelLipski:
Btw I've just bumped version on PR #1230, lemme rebase onto fresh `develop`...

PawelLipski (approved):

zb3:
I'd made the same mistake in my script and I remember I couldn't figure out why my code didn't work.. finally GPT4 helped me and then I realized this bug isn't obvious even for some experienced programmers..
So I searched popular repositories on sourcegraph using a simple regular expression, then I noticed the familiar repository :)

This bug can be automatically detected by `flake8` **with** the `flake8-bugbear` plugin (this bug has the symbol `B034` there)
https://github.com/PyCQA/flake8-bugbear

PR #1233, opened 2024-04-23
Set --no-edit-merge flag automatically when using -y in traverse

gjulianm:
When using `git machete traverse -y`, the behavior is to disable interactive rebase. However, if some branches have the `update=merge` annotation, they will show a merge commit edit window because the corresponding option `--no-edit-merge` is not set automatically. This PR changes that behavior.

### DIFF
--- a/tests/test_traverse.py
+++ b/tests/test_traverse.py
@@ -790,6 +790,43 @@ class TestTraverse(BaseTest):
             """,
         )
 
+    def test_traverse_with_merge_annotation_and_yes_option(self, mocker: MockerFixture) -> None:
+        (
+            self.repo_sandbox.remove_remote()
+            .new_branch("develop")
+            .commit()
+            .new_branch("mars")
+            .commit()
+            .new_branch("snickers")
+            .commit()
+            .check_out("mars")
+            .commit()
+            .check_out("develop")
+            .commit()
+        )
+
+        body: str = """
+            develop
+                mars update=merge
+                    snickers
+            """
+        rewrite_branch_layout_file(body)
+
+        self.patch_symbol(mocker, "builtins.input", mock_input_returning("q"))
+        launch_command("traverse")
+        self.patch_symbol(mocker, "builtins.input", mock_input_returning("n", "q"))
+        launch_command("traverse")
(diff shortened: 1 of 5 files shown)


PawelLipski:
Makes sense! could you add a test that would've captured the unintended behavior?

tests/test_update.py
```diff
@@ -72,6 +72,35 @@ def test_update_by_merge(self) -> None:
...
+                level-1-branch
+                    level-2-branch
+            """
+        rewrite_branch_layout_file(body)
+
+        self.repo_sandbox.check_out("level-1-branch")
+        old_level_1_commit_hash = self.repo_sandbox.get_current_commit_hash()
+        # If -y is specified, --no-edit-merge is implied. This test ensures that, if we got a message edit window
+        # the test would fail as the merge is not performed fully without interaction.
+        launch_command("update", "--merge", "-y")
```
PawelLipski: Whoops `update` doesn't accept `-y` ☹️ the behavior will only differ in case of `traverse` (as that's the only command so far where qualifiers like `update=merge` work)

gjulianm: Yeah I saw it, I fixed the test, I think it should be ok now.

PawelLipski:
Oh crap, brownout of Intel mac resources 😅

tests/test_traverse.py
```diff
@@ -790,6 +790,43 @@ def test_traverse_with_merge_annotation(self, mocker: MockerFixture) -> None:
...
+        )
+
+        body: str = """
+            develop
+                mars update=merge
+                    snickers
+            """
+        rewrite_branch_layout_file(body)
+
+        with overridden_environment(GIT_EDITOR='false'):
```
PawelLipski, line 815: I took liberty to patch this test a bit:

1. Remove the unnecessary `traverse` invocations
2. Make the test reproducible on each machine, regardless of what's the default editor and how it behaves when output is a non-terminal. Currently, it was passing on my machine since I've got `vim` set up as the default editor, and `vim` happens to exit immediately with non-zero exit code when executed in a headless context like these tests... but still, I can imagine scenarios when this test would pass even without your fix, or even get stuck indefinitely waiting for user input.

PawelLipski (approved):

PR #1238, opened 2024-04-26
Improve squash merge detection

gjulianm:
This PR improves detection of squash merges. The existing logic only works if there are no commits in between the fork point and the merge, which in busy repositories might not be the case. I've added direct comparison of diffs using `git patch-id`, which should be reasonably stable. 

The change adds some delay in `git machete status` as there are more commands being executed. For my repos this doesn't seem to be above "barely noticeable", but the impact could be worse when there are more commits to compare.

Also, in order to be able to run `patch_id` I've modified `_popen_git` and subsequent functions so they accept standard input, I hope that's ok.


PawelLipski:
Wow 😯  I'm gonna check how it performs on my large commercial repo (~800k commits)

gjulianm:
> Wow 😯 I'm gonna check how it performs on my large commercial repo (~800k commits)

My main worry is [this loop](https://github.com/VirtusLab/git-machete/pull/1238/files#diff-a50d300abea7f8b757dda0e66fc48839620aa8b04384556359a39fa3783f09b3R831-R836). For very long-lived branches that are way behind their fork point it could be a problem, I was thinking that maybe I could set a limit of commits (e.g., if we don't see the squash merge in the first 100 commits, then assume it's not merged).

PawelLipski:
> > Wow 😯 I'm gonna check how it performs on my large commercial repo (~800k commits)
> 
> My main worry is [this loop](https://github.com/VirtusLab/git-machete/pull/1238/files#diff-a50d300abea7f8b757dda0e66fc48839620aa8b04384556359a39fa3783f09b3R831-R836). For very long-lived branches that are way behind their fork point it could be a problem, I was thinking that maybe I could set a limit of commits (e.g., if we don't see the squash merge in the first 100 commits, then assume it's not merged).

Whoops invoking `git diff` in a loop will indeed be troublesome :/ currently the number of invoked git commands is `O(N)` with respect to branch layout size (even though if some individual commands like `git log` might be linear wrt. git log size) ☹️

PawelLipski:
Actually... stepping back a bit, how often did you encounter the case when a squash merge has not been detected (but is now detected after your PR)? Could you provide a quick example how that situation looked like? it's been a long time since I've touched that logic TBH 🤔

gjulianm:
> Actually... stepping back a bit, how often did you encounter the case when a squash merge has not been detected (but is now detected after your PR)? Could you provide a quick example how that situation looked like? it's been a long time since I've touched that logic TBH 🤔

Quite often, in the main repo I use there are always commits between the fork point and the squash merge. I validated with the tests I added that indeed that was the case, if you remove the new logic (e.g., forcing an early exit [here](https://github.com/VirtusLab/git-machete/pull/1238/files#diff-a50d300abea7f8b757dda0e66fc48839620aa8b04384556359a39fa3783f09b3R809) changing `if result` to `if result or True`) the tests that add commits in between fail.

Up to this point I was manually sliding out merged branches because I thought it was a problem on how the merges were done in the repo, but then I noticed it only happened in busy repos, and that's how I found out the problem were those intermediate commits.

We might be able to do some shell magic to at least make git compute the patch id without having the diff for each commit in the Python code, let me check.

PawelLipski:
Huh okay... would you mind putting this logic behind a [git config key](https://git-machete.readthedocs.io/en/stable/#config)? 🤔 flag to `status` can also be considered instead, as there's already `--no-detect-squash-merges`. Maybe sth like `--squash-merge-detection={none,simple,exact}`? And then `--no-detect-squash-merges` to be deprecated.

gjulianm:
> Huh okay... would you mind putting this logic behind a [git config key](https://git-machete.readthedocs.io/en/stable/#config)? 🤔 flag to `status` can also be considered instead, as there's already `--no-detect-squash-merges`. Maybe sth like `--squash-merge-detection={none,simple,exact}`? And then `--no-detect-squash-merges` to be deprecated.

Sounds good, I'll take a look, thx!

git_machete/git_operations.py
```diff
@@ -223,8 +223,8 @@ def _run_git(self, git_cmd: str, *args: str, flush_caches: bool, allow_non_zero:
         return exit_code
 
     def _popen_git(self, git_cmd: str, *args: str,
-                   allow_non_zero: bool = False, env: Optional[Dict[str, str]] = None) -> CommandResult:
-        exit_code, stdout, stderr = utils.popen_cmd("git", git_cmd, *args, env=env)
+                   allow_non_zero: bool = False, env: Optional[Dict[str, str]] = None, input: Optional[str] = None) -> CommandResult:
```
PawelLipski, line 226:
```suggestion
                   allow_non_zero: bool = False, env: Optional[Dict[str, str]] = None, stdin: Optional[str] = None) -> CommandResult:
```

For consistency with `stdout, stderr` below

git_machete/utils.py
```diff
@@ -187,18 +187,21 @@ class PopenResult(NamedTuple):
 
 
 def _popen_cmd(cmd: str, *args: str,
-               cwd: Optional[str] = None, env: Optional[Dict[str, str]] = None) -> PopenResult:
+               cwd: Optional[str] = None, env: Optional[Dict[str, str]] = None, input: Optional[str] = None) -> PopenResult:
```
PawelLipski, line 190:
```suggestion
               cwd: Optional[str] = None, env: Optional[Dict[str, str]] = None, stdin: Optional[str] = None) -> PopenResult:
```

git_machete/git_operations.py
```diff
@@ -804,10 +804,53 @@ def is_equivalent_tree_reachable(
...
         result = earlier_tree_hash in intermediate_tree_hashes
-        debug(f"result = {result}")
+        debug(f"intermediate tree hashes result = {result}")
+
+        if result:
+            self.__is_equivalent_tree_reachable_cached[equivalent_to_commit_hash, reachable_from_commit_hash] = result
+            return result
+
+        # Let's try another way, a little more complex but takes into account the possibility
+        # that there were other commits between the fork point of the two branches and the squashed merge.
```
PawelLipski: Pls note that we strive to keep `git_machete/git_operations.py` as machete-agnostic as possible... and in particular, I wouldn't mention _fork point_ here (which is originally a git term, see `git rebase --fork-point`, but [heavily adapted within git machete](https://medium.com/virtuslab/make-your-way-through-the-git-rebase-jungle-with-git-machete-e2ed4dbacd02#:~:text=Appendix%3A%20fork%20point%20%E2%80%94%20not%20so%20easy%20business%E2%80%A6)).

PawelLipski: Maybe just
```suggestion
        # that there were other commits between the common ancestor of the two branches and the squashed merge.
```
Not sure if that preserves the exact sense 🤔

git_machete/git_operations.py
```diff
@@ -804,10 +804,53 @@ def is_equivalent_tree_reachable(
...
+        fork_point = self.get_merge_base(reachable_from_commit_hash, equivalent_to_commit_hash)
```
PawelLipski: Yeah, looks that `common_ancestor` would fit he context better

git_machete/git_operations.py
```diff
@@ -804,10 +804,53 @@ def is_equivalent_tree_reachable(
...
+        ).stdout
+        equivalent_patch_id = self.get_patch_id(equivalent_changeset)
+
+        if equivalent_patch_id is None and equivalent_changeset.strip() == '':
+            # Empty changeset means the branches are identical, so the tree is equivalent.
+            self.__is_equivalent_tree_reachable_cached[equivalent_to_commit_hash, reachable_from_commit_hash] = True
+            return True
+
+        for logentry in self.get_commits_between(fork_point, reachable_from_commit_hash):
+            changeset = self._popen_git(
```
PawelLipski: Hmmm actually... use `git log --patch` to retrieve all patches in a single `git` invocation, then pipe the entire output to `git patch-id`

PawelLipski: Then no loop necessary 🎉 so still `O(N_managed_branches)` git invocations.

PawelLipski: Still, even if this optimization in place, TBD if flag/config key is necessary

git_machete/git_operations.py
```diff
@@ -804,10 +804,53 @@ def is_equivalent_tree_reachable(
...
+        debug(f"patch id result = {result}")
         self.__is_equivalent_tree_reachable_cached[equivalent_to_commit_hash, reachable_from_commit_hash] = result
         return result
 
+    def get_patch_id(self, patch_contents: str) -> Optional[str]:
+        out = utils.get_non_empty_lines(self._popen_git("patch-id", input=patch_contents).stdout)
+
+        if len(out) == 0:
+            return None
+        return out[0].split(' ')[0]
```
PawelLipski: Pls mention in code comment why the split is necessary (the second part is commit hash, which we don't care about here)

git_machete/git_operations.py
```diff
@@ -804,10 +804,53 @@ def is_equivalent_tree_reachable(
...
+            fork_point,
+            equivalent_to_commit_hash
+        ).stdout
+        equivalent_patch_id = self.get_patch_id(equivalent_changeset)
+
+        if equivalent_patch_id is None and equivalent_changeset.strip() == '':
+            # Empty changeset means the branches are identical, so the tree is equivalent.
+            self.__is_equivalent_tree_reachable_cached[equivalent_to_commit_hash, reachable_from_commit_hash] = True
+            return True
+
```
PawelLipski: Maybe just
```suggestion
        equivalent_changeset = self._popen_git(
            "diff",
            fork_point,
            equivalent_to_commit_hash
        ).stdout
        if equivalent_changeset.strip() == '':
            # Empty changeset means the branches are identical, so the tree is equivalent.
            self.__is_equivalent_tree_reachable_cached[equivalent_to_commit_hash, reachable_from_commit_hash] = True
            return True
        equivalent_patch_id = self.get_patch_id(equivalent_changeset)
        if equivalent_patch_id is None:

```

git_machete/git_operations.py
PawelLipski: But as for now, let's punt flag/config key, I can always add it later

tests/test_git_operations.py
```diff
@@ -45,3 +45,65 @@ def is_commit_present_in_repository(revision: AnyRevision) -> bool:
...
+                .commit("master first commit")
+                .new_branch("feature")
+                .commit("feature commit")
+                .check_out("master")
+        )
+
+        git = GitContext()
+        assert git._run_git("merge", "--squash", "feature", flush_caches=False) == 0
+        assert git._run_git("commit", "-m", "squashed", flush_caches=False) == 0
+        assert git.is_equivalent_tree_reachable(AnyRevision("feature"), AnyRevision("master")) is True
```
PawelLipski: I'd name the params of `is_equivalent_tree_reachable` explicitly in these tests

git_machete/git_operations.py
PawelLipski: Maybe even better (at least that's what speaks better to me, somehow):
```suggestion
        # that before the `equivalent_to` commit has been squashed onto into the history of `reacheable_from`,
        # more commits appeared on history of `reacheable_from`.
        # Hence, there might NOT be a commit reachable from `reacheable_from` that points to the exact same tree as in `equivalent_to`.
```

RELEASE_NOTES.md
```diff
@@ -3,6 +3,7 @@
 ## New in git-machete 3.25.3
 
 - fixed: `-y` option in `git machete traverse` automatically sets `--no-edit-merge` flag, to retain behavior when the `update=merge` qualifier is set (contributed by @gjulianm)
+- fixed: Better detection of squash merges and rebases when there are commits between the fork point and the merge/rebase (contributed by @gjulianm)
```
```suggestion
- fixed: better detection of squash merges and rebases when there are commits between the fork point and the merge/rebase (contributed by @gjulianm)
```

git_machete/git_operations.py
PawelLipski: Ha! even better, git cherry (https://git-scm.com/docs/git-cherry), seems to be created for this particular purpose

PawelLipski: (If only you can align the arguments required by git cherry right with what's required here)

gjulianm:
Added the option, let me know if it looks good. I wasn't sure if some of the docs are automatic or not (e.g. the traverse/status help pages) so I didn't update them. Also, not sure if the deprecation is how you expected.

I'll take a look later regarding the use of `git log -p` to get all the patches in one git call.

docs/source/cli/config.rst
```diff
@@ -49,6 +49,12 @@ Note: ``config`` is not a command as such, just a help topic (there is no ``git
     If you want the worktree to have its own branch layout file (located under ``.git/worktrees/.../machete``),
     set ``git config machete.worktree.useTopLevelMacheteFile false``.
 
+``machete.squashMergeDetection``:
```
PawelLipski: Pls run `tox`! and the man page and Python docs will regenerate

git_machete/cli.py
```diff
@@ -364,6 +365,9 @@ def add_code_hosting_parser(command: str, subcommand_suffix: str, include_sync:
     status_parser.add_argument('-l', '--list-commits', action='store_true')
     status_parser.add_argument('-L', '--list-commits-with-hashes', action='store_true')
     status_parser.add_argument('--no-detect-squash-merges', action='store_true')
+    status_parser.add_argument('--squash-merge-detection',
```
PawelLipski: As you can see, a test for `status` is now failing in CI... but pls refrain from fixing it, maybe once you migrate to `git log --patch` or `git cherry`, the extra option/git config key won't even be necessary and what's now `exact` mode will be made the default 🤔

tests/test_git_operations.py
```diff
@@ -45,3 +46,100 @@ def is_commit_present_in_repository(revision: AnyRevision) -> bool:
                                                       later=LocalBranchShortName('master')) is False
         assert self.repo_sandbox.is_ancestor_or_equal(earlier=LocalBranchShortName('develop'),
                                                       later=LocalBranchShortName('feature')) is True
+
+    def test_is_equivalent_tree_reachable_with_squash_merge(self) -> None:
```
PawelLipski: These tests are failing now as well ☹️

gjulianm: Yeah, sorry about that, pushed too early. They should be working now.

git_machete/git_operations.py
gjulianm: I checked out `git cherry`, but it suffers from the same problem, it can only check if a single commit has been applied, not multiple at the same time.

gjulianm:
Now it's using `git log -p`, we still have to call `git patch-id` for each changeset though, I didn't see any option to get the patch-id from the `git log` call directly.

git_machete/git_operations.py
PawelLipski: Aww indeed, I missed that... from git-cherry manual:

> Determine whether there are commits in <head>..<upstream> that are equivalent to those in the range <limit>..<head>
> Outputs the SHA1 of every commit in <limit>..<head>, prefixed with - for commits that have an equivalent in <upstream>, and + for commits that do not.

There might be a workaround if needed: creating a synthetic squash commit on the go using `git commit-tree`, similarly to what `git machete squash` does. Still, there might still be a problem with aligning args to `git cherry` correctly (even more so).

Let's leave as is for now... I'm gonna test your implementation for some time on my large commercial repo, with `GIT_MACHETE_MEASURE_COMMAND_TIME` env var (undocumented, internal) to see how much time is spent in each underlying `git` command.

git_machete/git_operations.py
```diff
@@ -911,6 +958,22 @@ def get_commits_between(self, earliest_exclusive: AnyRevision, latest_inclusive:
             utils.get_non_empty_lines(self._popen_git("log", "--format=%H:%h:%s", f"^{earliest_exclusive}", latest_inclusive, "--").stdout)
         ))))
 
+    def get_diffs_between(self, earliest_exclusive: AnyRevision, latest_inclusive: AnyRevision) -> Dict[FullCommitHash, str]:
+        # Returns a dictionary of git hashes and their diffs between two revisions
+        marker = "__MACHETE___COMMIT_MARKER__"  # Use a marker so we can split the output into individual diffs for each commit
```
PawelLipski: Hmmm is it possible to just pipe the entire output of `git log --patch` to `git patch-id`? and then you should get `(patch SHA, commit SHA)` pairs 🤔 no parsing of the output of `git log --patch` needed

git_machete/git_operations.py
PawelLipski: And then let's decide if this new mode should be kept behind a flag/config key, or maybe it could be seamlessly made a default

git_machete/git_operations.py
PawelLipski: If possible, this should speed up the entire thing a little bit more (less `git` process invocations)

git_machete/git_operations.py
```diff
@@ -911,6 +958,22 @@ def get_commits_between(self, earliest_exclusive: AnyRevision, latest_inclusive:
...
+        out = self._popen_git("log", "--patch", f"--format=format:{marker}%H", f"^{earliest_exclusive}", latest_inclusive, "--").stdout
```
PawelLipski: And indeed, maybe add some limit on number of commits included in the results... like `-1000` option (1000 extracted to `constants.py`)

git_machete/git_operations.py
PawelLipski: Oh, pls rebase onto the fresh `develop`, I've added some changes useful for time measurement

PawelLipski: (https://github.com/VirtusLab/git-machete/pull/1245)

git_machete/git_operations.py
gjulianm: `git patch-id` seems to be very low-level, I didn't find any way to make it output one patch ID per block of input. We could do something with AWK and some bash magic, the number of `git` process invocations would be the same although they wouldn't be done from Python.

git_machete/git_operations.py
gjulianm: Done! (sorry about the delays in getting to your comments, had a busy week :D)

PawelLipski: Thanks! I'll take a look once I'm done with my commercial work ⚒️

git_machete/git_operations.py
PawelLipski: Hmmm could you check `git log -p | git patch-id` on `master` in this repo?

I'm getting something like
```
3fdf24106a4e04a993e33ab4cf267980e90d1f48 cbd9cf96fe298e3339fd00553d758c5e48347c62
8134efef8d2173a007a246b4afcd952839bd338a 2a8ed18196145b15f2e2816f66f9e76ea77e5f7f
e4f8fe513423413e89920af61c13715b463bf703 ace2593d9d25144d8153acc33700228bcca2ff27
1a2813c326d41d10a041742dbf367ce0fa47540a 33808252399b358d068c63651bb23c9b3ab465c4
0df316d8afe3d3f9b2acd9f8fe33f9fedfb4c386 fe48b5e51299cbfc143b7b3051f6eeb84b4c95f4
6487ae2cc44a597e9122f6e146d90f3654476f46 fba21dfb749f5af7ccd8f76fb41a72ce91903312
75e74c3587ec1cdcc260c538bf5d94f6ccaa57ad 40531ec669fa535c2993a52dbfe8d6db9850b29d
...
```

which seems to be `(patch hash, commit hash)` pairs... would it be applicable for this use? 🤔

gjulianm: It definitely is... I didn't think this was possible from reading the docs, I didn't think of actually trying :D

PawelLipski: Okay! lemme now try it on my large commercial repo. I'll do the final touches (any documentation etc. fixes needed)

PawelLipski: ☝🏻 the repo has ~800k commits and ~400k tracked files... the execution times `git log --patch` are in the order of 2-4 seconds **for each managed branch**. So indeed, let's keep it behind a flag/config key, as you implemented it (and **not** make `exact` mode the default) 😅

gjulianm: Heh, yeah, it's quite a bit. I merged the latest develop onto this branch and also fixed some errors in CI, and checked that indeed simple mode is the default.

PawelLipski:
I'm gonna introduce minor fixes directly onto your branch. Sorry for the lags, I'm busy as heck in my commercial job :/

PawelLipski:
Also, the comments on the PR are for myself from now on &mdash; you've already done more than I could ever ask 😅

RELEASE_NOTES.md
```diff
@@ -1,7 +1,8 @@
 # Release notes
 
-## New in git-machete 3.25.3
+## New in git-machete 3.26.0
 
+- added: better detection of squash merges and rebases, controlled by argument `--squash-merge-detection={none,simple,exact}` and git config key `machete.squashMergeDetection` (contributed by @gjulianm)
```
PawelLipski: New flag added, bumping minor version (as per [versioning policy](https://github.com/VirtusLab/git-machete/blob/master/CONTRIBUTING.md#versioning)).

git_machete/git_operations.py
```diff
@@ -796,7 +798,12 @@ def is_equivalent_tree_reachable(
             return True
 
         if (equivalent_to_commit_hash, reachable_from_commit_hash) in self.__is_equivalent_tree_reachable_cached:
-            return self.__is_equivalent_tree_reachable_cached[equivalent_to_commit_hash, reachable_from_commit_hash]
+            prev_mode, prev_result = self.__is_equivalent_tree_reachable_cached[equivalent_to_commit_hash, reachable_from_commit_hash]
+
+            # Only return cached result if we're using the same mode or if we already checked with the
```
PawelLipski: I don't think this extra caching check makes sense. IIUC within the given git-machete invocation (i.e. within the given Python process lifetime), there's only ever a single set of CLI options, and hence a single squash detection mode. I'll probably revert that.

PawelLipski: And within tests, a new `GitContext` is created in each test method anyway.

git_machete/git_operations.py
```diff
@@ -812,10 +819,48 @@ def is_equivalent_tree_reachable(
...
+        debug(f"patch id result = {result}")
+        cache_value = (opt_squash_merge_detection, result)
+        self.__is_equivalent_tree_reachable_cached[equivalent_to_commit_hash, reachable_from_commit_hash] = cache_value
         return result
 
+    def get_patch_id(self, patch_contents: str) -> Optional[str]:
+        out = utils.get_non_empty_lines(self._popen_git("patch-id", input=patch_contents).stdout)
+
+        if len(out) == 0:
+            return None
```
PawelLipski, line 877: Line uncovered by tests. Probably okay to leave it so, there doesn't seem to be any reasonable path that would cover this case. Also, doesn't feel right to just `# pragma: no cover` it away.

docs/source/cli/config.rst
```diff
@@ -49,6 +49,12 @@ Note: ``config`` is not a command as such, just a help topic (there is no ``git
...
+    Controls the algorithm used to detect squash merges. Possible values are:
+    * ``none``: No squash merge/rebase detection.
+    * ``simple``: Compares the tree state of the merge commit with the tree state of the upstream branch. This detects squash merges/rebases as long as there was not any commit on the upstream branch since the last common commit.
```
PawelLipski:
```suggestion
    * ``simple``: Compares the tree state of the merge commit with the tree state of the upstream branch. This detects squash merges/rebases as long as there was no commit on the upstream branch since the last common commit.
```

docs/source/cli/config.rst
```diff
@@ -49,6 +49,12 @@ Note: ``config`` is not a command as such, just a help topic (there is no ``git
...
+    * ``exact``: Compares the patch that would be applied by the merge commit with the commits that occurred on the upstream branch since the last common commit. This detects squash merges/rebases even if there were commits on the upstream branch since the last common commit. However, it might have a performance impact as it requires listing all the commits in the upstream.
```
PawelLipski:
```suggestion
    * ``exact``: Compares the patch that would be applied by the merge commit with the commits that occurred on the upstream branch since the last common commit. This detects squash merges/rebases even if there were commits on the upstream branch since the last common commit. However, it might have a performance impact as it requires computing patches for commits in the upstream.
```

git_machete/cli.py
```diff
@@ -455,7 +462,10 @@ def update_cli_options_using_parsed_args(
         elif opt == "n":
             cli_opts.opt_n = True
         elif opt == "no_detect_squash_merges":
-            cli_opts.opt_no_detect_squash_merges = True
+            print("--no-detect-squash-merges is deprecated, use --squash-merge-detection=none")
+            cli_opts.opt_squash_merge_detection = SquashMergeDetection.NONE
+        elif opt == "squash_merge_detection" and arg is not None:  # If the value is None it means the user did not provide any value
```
```suggestion
        elif opt == "squash_merge_detection" and arg is not None:  # If the value is None it means the user did not provide any value, which is forbidden by argparse
```

git_machete/git_operations.py
```diff
@@ -786,6 +787,7 @@ def is_equivalent_tree_reachable(
             self,
             equivalent_to: AnyRevision,
             reachable_from: AnyRevision,
+            opt_squash_merge_detection: SquashMergeDetection
```
PawelLipski: This method should no longer be called `is_equivalent_tree_reachable`, since in `EXACT` mode we're checking for some other condition (the method result might be true even though there's **no** equivalent tree reachable).

PawelLipski: Or even better, I'd keep `GitContext` oblivious to `SquashMergeDetection` mode. This is a machete-specific concept, after all, and we strive to keep `GitContext` as machete-agnostic as possible.

So, I'll extract a method like `is_equivalent_patch_reachable` (the exact name TBD), and call it somewhere from `MacheteClient` so that the overall logic is equivalent to the current state.

docs/source/cli/config.rst
```diff
@@ -49,6 +49,12 @@ Note: ``config`` is not a command as such, just a help topic (there is no ``git
...
+
 
 **Environment variables:**
```
PawelLipski, line 56: Docs for `status` and `traverse` need to be updated as well

git_machete/options.py
```diff
@@ -64,6 +66,10 @@ def validate(self) -> None:
         if self.opt_sync_github_prs and self.opt_sync_gitlab_mrs:
             raise MacheteException(
                 "Option `-H/--sync-github-prs` cannot be specified together with `-L/--sync-gitlab-mrs`.")
+        if not isinstance(self.opt_squash_merge_detection, SquashMergeDetection):
+            valid_values = ', '.join(e.value for e in SquashMergeDetection)
```
PawelLipski: To be covered by tests

git_machete/constants.py
```diff
@@ -27,3 +28,19 @@ class GitFormatPatterns(Enum):
...
+class SquashMergeDetection(Enum):
+    NONE = "none"
+    SIMPLE = "simple"
+    EXACT = "exact"
+
+    @staticmethod
+    def from_string(value: str) -> 'SquashMergeDetection':
+        if value == "none":
+            return SquashMergeDetection.NONE
+        if value == "simple":
```
PawelLipski: To be covered by tests. There aren't any end-to-end tests currently that would check for a result of a full invocation of `git machete status` in `EXACT` mode, only lower-level tests of `GitContext`. The former are typically preferable as an actual user interaction is tested.

git_machete/git_operations.py
```diff
@@ -911,6 +956,20 @@ def get_commits_between(self, earliest_exclusive: AnyRevision, latest_inclusive:
             utils.get_non_empty_lines(self._popen_git("log", "--format=%H:%h:%s", f"^{earliest_exclusive}", latest_inclusive, "--").stdout)
         ))))
 
+    def get_patch_ids_for_commits_between(
```
PawelLipski: Move next to `get_patch_id`

tests/test_status.py
```diff
@@ -380,7 +380,7 @@ def test_squashed_branch_recognized_as_merged_with_traverse(self) -> None:
 
         # but under --no-detect-squash-merges, feature is detected as "x" (behind) develop
         assert_success(
-            ["status", "-l", "--no-detect-squash-merges"],
+            ["status", "-l", "--squash-merge-detection=none"],
```
PawelLipski: Lemme add a test that would only pass with `--squash-merge-detection=exact`

RELEASE_NOTES.md
PawelLipski: Also, mention the deprecation of `--no-detect-squash-merges` + which commands are affected by the flag changes

git_machete/git_operations.py
```diff
@@ -812,10 +819,48 @@ def is_equivalent_tree_reachable(
...
+            common_ancestor, reachable_from_commit_hash, MAX_COMMITS_FOR_SQUASH_MERGE_DETECTION)
+        patch_ids = set(patchmap.values())
+        result = equivalent_patch_id in patch_ids
+
+        debug(f"patch id result = {result}")
+        cache_value = (opt_squash_merge_detection, result)
+        self.__is_equivalent_tree_reachable_cached[equivalent_to_commit_hash, reachable_from_commit_hash] = cache_value
         return result
 
+    def get_patch_id(self, patch_contents: str) -> Optional[str]:
```
PawelLipski: Introduce a dedicated type (alias for `str`) for patch ids

docs/source/cli/config.rst
```diff
@@ -49,6 +49,12 @@ Note: ``config`` is not a command as such, just a help topic (there is no ``git
     If you want the worktree to have its own branch layout file (located under ``.git/worktrees/.../machete``),
     set ``git config machete.worktree.useTopLevelMacheteFile false``.
 
+``machete.squashMergeDetection``:
+    Controls the algorithm used to detect squash merges. Possible values are:
+    * ``none``: No squash merge/rebase detection.
+    * ``simple``: Compares the tree state of the merge commit with the tree state of the upstream branch. This detects squash merges/rebases as long as there was no commit on the upstream branch since the last common commit.
+    * ``exact``: Compares the patch that would be applied by the merge commit with the commits that occurred on the upstream branch since the last common commit. This detects squash merges/rebases even if there were commits on the upstream branch since the last common commit. However, it might have a performance impact as it requires computing patches for commits in the upstream.
+
```
PawelLipski: Mention the default + wrap the long lines

tests/test_git_operations.py
```diff
@@ -45,3 +46,122 @@ def is_commit_present_in_repository(revision: AnyRevision) -> bool:
...
```
PawelLipski: Rename the tests adequately once the tested method is split/renamed

git_machete/git_operations.py
PawelLipski: I ended up splitting the caches into one for equivalent-tree-reachable and one for equivalent-patch-reachable

git_machete/git_operations.py
PawelLipski: Done, added `FullPatchId`

git_machete/git_operations.py
```diff
@@ -811,11 +822,70 @@ def is_equivalent_tree_reachable(
...
+
+        if equivalent_to_commit_hash == reachable_from_commit_hash:
+            return True
+
+        if (equivalent_to_commit_hash, reachable_from_commit_hash) in self.__is_equivalent_patch_reachable_cached:
+            return self.__is_equivalent_patch_reachable_cached[equivalent_to_commit_hash, reachable_from_commit_hash]
+
+        common_ancestor = self.get_merge_base(reachable_from_commit_hash, equivalent_to_commit_hash)
+        if not common_ancestor:
+            return False
```
PawelLipski, line 852: Line uncovered. Extremely unlikely in real-life scenarios (requires orphan commits), but I've got coverage OCD :trollface:

git_machete/cli.py
```diff
@@ -529,6 +535,10 @@ def update_cli_options_using_config_keys(
         else:
             cli_opts.opt_push_tracked, cli_opts.opt_push_untracked = False, False
 
+    squash_merge_detection = git.get_config_attr_or_none(key=git_config_keys.SQUASH_MERGE_DETECTION)
+    if squash_merge_detection is not None:
+        cli_opts.opt_squash_merge_detection = SquashMergeDetection.from_string(squash_merge_detection)
```
PawelLipski: Mention in the error message that this is coming from config key

gjulianm:
> I'm gonna introduce minor fixes directly onto your branch. Sorry for the lags, I'm busy as heck in my commercial job :/

No problem, thank you very much for the help and for the tool too :)

git_machete/constants.py
```diff
@@ -27,3 +30,20 @@ class GitFormatPatterns(Enum):
     FULL_MESSAGE = "%B"
     # subject NOT included
     MESSAGE_BODY = "%b"
+
+
+class SquashMergeDetection(Enum):
```
PawelLipski, line 36: Rel: https://github.com/VirtusLab/git-machete/issues/1257

docs/source/cli/slide-out.rst
```diff
@@ -7,7 +7,9 @@ slide-out
 .. code-block:: shell
 
     git machete slide-out --removed-from-remote [--delete]
-    git machete slide-out [-d|--down-fork-point=<down-fork-point-commit>] [--delete] [-M|--merge] [-n|--no-edit-merge|--no-interactive-rebase] [<branch> [<branch> [<branch> ...]]]
+    git machete slide-out [-d|--down-fork-point=<down-fork-point-commit>] [--delete]
```
PawelLipski, line 10: So that it doesn't need to be scrolled horizontally in the browser

PawelLipski (approved):

PawelLipski:
🎉

PawelLipski:
To be released soon as v3.26.0

Thanks for your work! this was one of the most non-trivial changes in the recent years TBH 😅

gjulianm:
> To be released soon as v3.26.0
> 
> Thanks for your work! this was one of the most non-trivial changes in the recent years TBH 😅

Great! Thank you for finishing it and all the help :D

PR #1295, opened 2024-08-06
log: pass through certain `git log` options for `git machete log`

tdyas:
Pass through the `-n` (`--max-count`) and `-p` (`--patch`) options for `git machete log`. These are useful to make use of.

### DIFF
--- a/git_machete/cli.py
+++ b/git_machete/cli.py
@@ -121,6 +121,29 @@ class MacheteHelpAction(argparse.Action):
         parser.exit(status=ExitCode.SUCCESS)
 
 
+# Inspired by an answer at https://stackoverflow.com/questions/10027242/python-argparse-to-handle-arbitrary-numeric-options-like-head1.
+class UpdateCount(argparse.Action):
+    """Helper to support -N style options."""
+
+    def __call__(
+            self,
+            parser: argparse.ArgumentParser,  # noqa: U100
+            namespace: argparse.Namespace,
+            values: Union[str, Sequence[Any], None],  # noqa: F841, U100
+            option_string: Optional[str] = None  # noqa: index, F841, U100
+    ) -> None:
+        if not option_string:
+            return
+        digit = option_string[1]
+        increment = int(digit)
+        previous = getattr(namespace, self.dest)
+        if previous is None:
+            previous = 0
+        else:
+            previous *= 10
+        setattr(namespace, self.dest, previous + increment)
+
+
 def create_cli_parser() -> argparse.ArgumentParser:
     common_args_parser = argparse.ArgumentParser(
         prog='git machete',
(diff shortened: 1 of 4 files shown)


tdyas:
This PR takes the approach of explicitly configuring and limiting the options which can be passed through. Another approach would be to just have an actual `argparse` pass-through argument and basically impose no validation whatsoever on what the user can provide to be passed through. Thoughts?

PawelLipski:
Makes sense! I also thought of passing arbitrary args to `git diff` via `git machete diff` at some point.

Actually... could you cook up a PoC for something like:

```shell
git machete log [<branch>] -- -n1 --patch --other-arbitrary-git-log-options...
```

So `--` separates log-specific arguments 

This would be more robust than just selecting specific args to pass to the underlying command (although I agree that `-n` and `--patch` are probably the most commonly used ones in `git log`) 🤔

tdyas:
> Actually... could you cook up a PoC for something like:
> 
> ```shell
> git machete log [<branch>] -- -n1 --patch --other-arbitrary-git-log-options...
> ```
> 
> So `--` separates log-specific arguments
> 
> This would be more robust than just selecting specific args to pass to the underlying command (although I agree that `-n` and `--patch` are probably the most commonly used ones in `git log`) 🤔

Makes sense. Will do.

tdyas:
`argparse` does not seem to support only capturing arguments which occur only after a `--`. See https://stackoverflow.com/questions/22850332/getting-the-remaining-arguments-in-argparse for what other people have found.

The issue here is that even if the `branch` is missing before a `--` , argparse will still try to capture it even if after a `--`:

```
Python 3.12.4 (main, Jul 17 2024, 13:04:38) [GCC 11.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from argparse import ArgumentParser
>>> p = ArgumentParser()
>>> p.add_argument("branch", action='store')
_StoreAction(option_strings=[], dest='branch', nargs=None, const=None, default=None, type=None, choices=None, required=True, help=None, metavar=None)
>>> p.add_argument("extra", nargs='*')
_StoreAction(option_strings=[], dest='extra', nargs='*', const=None, default=None, type=None, choices=None, required=True, help=None, metavar=None)
>>> p.parse_args(["foo", "bar", "xyzzy"])
Namespace(branch='foo', extra=['bar', 'xyzzy'])
>>> p.parse_args(["--", "foo", "bar", "xyzzy"])
Namespace(branch='foo', extra=['bar', 'xyzzy'])
>>> p.parse_known_args(["--", "foo", "bar", "xyzzy"])
(Namespace(branch='foo', extra=['bar', 'xyzzy']), [])
>>> p.parse_known_args(["foo", "bar", "xyzzy"])
(Namespace(branch='foo', extra=['bar', 'xyzzy']), [])
```

I don't see a way to have argparse only fill an argument if after a `--`. :(

PawelLipski:
Okay! thanks for checking... I'll take over your PR from now on (fix CI, add docs and tests)

PawelLipski:
Ha! actually, I've made pass-through args work with a little hack, see https://github.com/VirtusLab/git-machete/pull/1303

tdyas:
> Ha! actually, I've made pass-through args work with a little hack, see #1303

Cool!

tdyas:
Closing this PR on favor of #1303 then.

PR #1313, opened 2024-08-26
Use the same algorithm to recognize merged branches in `delete-unmanaged` as for `status` and `traverse`

PawelLipski:


git_machete/client.py
```diff
@@ -1283,46 +1284,47 @@ def rebase(self, onto: AnyRevision, from_exclusive: AnyRevision, branch: LocalBr
...
+            return
 
-            branches_to_delete_unmerged_to_head = [branch for branch in branches_to_delete if branch not in branches_merged_to_head]
-            for branch in branches_to_delete_unmerged_to_head:
-                msg_core = f"{bold(branch)} (unmerged to HEAD)"
+        if opt_yes:
+            for branch in branches_to_delete:
+                print(f"Deleting branch {bold(branch)}...")
+                self.__git.delete_branch(branch, force=True)
+        else:
```
PawelLipski, line 1307: @cinnamennen work in progress... note that this will change the message displayed next to the squash-merged branch (from `unmerged` to `merged`), but the confirmation will still be needed (i.e. `--yes` is needed to turn off the need  for confirmation completely)

PR #1326, opened 2024-09-16
Add check for AUR package build

PawelLipski:
# Based on PR #1333

## Chain of upstream PRs as of 2024-09-18

* PR #1333:
  `develop` ← `fix/1329-workaround-git-patch-id`

  * **PR #1326 (THIS ONE)**:
    `fix/1329-workaround-git-patch-id` ← `fix/aur-test-since-3.26.0`


ci/aur/Dockerfile
```diff
@@ -0,0 +1,13 @@
+
+FROM archlinux:base-devel
+RUN pacman -Sy --noconfirm git
```
ilai-deutel, line 3: From https://hub.docker.com/_/archlinux:

> Arch Linux is a rolling release distribution, so a full update is recommended when installing new packages. In other words, we suggest to either execute `RUN pacman -Syu` immediately after your `FROM` statement or as soon as you docker run into a container.

This should ensure you always use the latest version of every package (not just `git`), which is expected in ArchLinux systems

From https://wiki.archlinux.org/title/System_maintenance#Partial_upgrades_are_unsupported:

> Do not use:
> * pacman -Sy package

PawelLipski: Good to know, fixed in #1337

PR #1332, opened 2024-09-18
Use aware datetime instead of deprecated utcfromtimestamp

ilai-deutel:
`datetime.datetime.utcfromtimestamp` is [deprecated](https://docs.python.org/3/library/datetime.html#datetime.datetime.utcfromtimestamp) since Python version 3.12


PawelLipski (approved):

PawelLipski:
Thanks!

PR #1352, opened 2024-10-12
Provide a GitHub PR title from `.git/info/title` file

### LINKED ISSUE #1351: Provide a GitHub PR title from `.git/info/title` file
Similar to the functionality for adding a PR description from `.git/info/description`, it would be useful to be able to set the PR title from a `.git/info/title` file.

This is a suggestion for an alternative method to the `--title` CLI parameter, which should take precedence over this method.

## Expected behavior

Given that
- `.git/info/title` contains a title, e.g. "Add New Feature"
- `.git/info/description` contains a description, e.g. "# Summary\n..."

When we invoke
```
git machete github create-pr
```

Then a pull request is created with the title and description set.

russellyeo:
Solves https://github.com/VirtusLab/git-machete/issues/1351 (not sure if having a linked issue is preferred or not, created one just in case)

Similar to the functionality for adding a PR description from `.git/info/description`, it would be useful to be able to set the PR title from a `.git/info/title` file.

This is a suggestion for an alternative method to the `--title` CLI parameter, which should take precedence over this method.

## Expected behavior

Given that
- `.git/info/title` contains a title, e.g. "Add New Feature"
- `.git/info/description` contains a description, e.g. "# Summary\n..."

When we invoke
```
git machete github create-pr
```

Then a pull request is created with the title and description set.

### DIFF
--- a/docs/source/cli/github.rst
+++ b/docs/source/cli/github.rst
@@ -83,11 +83,7 @@ Creates, checks out and manages GitHub PRs while keeping them reflected in branc
     If ``.git/info/milestone`` file is present, its contents (a single number --- milestone id) are used as milestone.
     If ``.git/info/reviewers`` file is present, its contents (one GitHub login per line) are used to set reviewers.
 
-    The PR title is determined in the following order of priority:
-    1. If the `--title` option is provided, its value is used as the PR title.
-    2. If the `.git/info/title` file is present, its contents (a single line) are used as the PR title.
-    3. If neither the `--title` option is provided nor the `.git/info/title` file is present, the PR title defaults to the subject of the first unique commit on the branch.
-
+    The subject of the first unique commit of the branch is used as PR title.
     If ``.git/info/description`` or ``.github/pull_request_template.md`` template is present, its contents are used as PR description.
     Otherwise (or if ``machete.github.forceDescriptionFromCommitMessage`` is set), PR description is taken from message body of the first unique commit of the branch.
 
--- a/docs/source/cli/github.rst
+++ b/docs/source/cli/github.rst
@@ -83,7 +83,11 @@ Creates, checks out and manages GitHub PRs while keeping them reflected in branc
     If ``.git/info/milestone`` file is present, its contents (a single number --- milestone id) are used as milestone.
     If ``.git/info/reviewers`` file is present, its contents (one GitHub login per line) are used to set reviewers.
 
-    The subject of the first unique commit of the branch is used as PR title.
+    The PR title is determined in the following order of priority:
+    1. If the `--title` option is provided, its value is used as the PR title.
+    2. If the `.git/info/title` file is present, its contents (a single line) are used as the PR title.
+    3. If neither the `--title` option is provided nor the `.git/info/title` file is present, the PR title defaults to the subject of the first unique commit on the branch.
+
     If ``.git/info/description`` or ``.github/pull_request_template.md`` template is present, its contents are used as PR description.
     Otherwise (or if ``machete.github.forceDescriptionFromCommitMessage`` is set), PR description is taken from message body of the first unique commit of the branch.
 
--- a/tests/test_github_create_pr.py
+++ b/tests/test_github_create_pr.py
@@ -986,4 +986,3 @@ class TestGitHubCreatePR(BaseTest):
             o-feature *  PR #1 (some_other_user)
             """,
         )
-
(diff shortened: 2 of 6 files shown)


russellyeo:
This was tested by creating and merging a PR in my fork of the repository here https://github.com/russellyeo/git-machete/pull/2, with a local `.git/info/title` file providing the title

PawelLipski:
Nice! complete with tests and docs 🥰 I'll take a closer look tmrw/Monday

docs/source/cli/github.rst
```diff
@@ -83,7 +83,11 @@ Creates, checks out and manages GitHub PRs while keeping them reflected in branc
     If ``.git/info/milestone`` file is present, its contents (a single number --- milestone id) are used as milestone.
     If ``.git/info/reviewers`` file is present, its contents (one GitHub login per line) are used to set reviewers.
 
-    The subject of the first unique commit of the branch is used as PR title.
+    The PR title is determined in the following order of priority:
+    1. If the `--title` option is provided, its value is used as the PR title.
```
PawelLipski: Good spot, `--title` isn't mentioned here in the current docs

PawelLipski:
Hmmm one general consideration... I'd say that's it's rather rare for people to have a constant PR title that they'd like to reuse across pull requests (and `.git/info/...` or other such files like in `.github/...` are best suited for such repetitive contents). This is unlike with PR description, which is very likely to be reused across PRs. What is your use case/what is the PR title that you're reusing? 🤔

russellyeo:
Hey @PawelLipski, thanks for taking a look at this. I also want to thank you for the excellent tool - I now can't imagine working without it.

> Hmmm one general consideration... I'd say that's it's rather rare for people to have a constant PR title that they'd like to reuse across pull requests (and `.git/info/...` or other such files like in `.github/...` are best suited for such repetitive contents). This is unlike with PR description, which is very likely to be reused across PRs. What is your use case/what is the PR title that you're reusing? 🤔

Good point. Perhaps my workflow is a little unusual - I use a git alias which opens these files one-by-one to gather my input and then create the PR:

```sh
[alias]
    create-pr = !code --wait .git/info/description && code --wait .git/info/reviewers && git machete github create-pr
```
(Invoked with `git create-pr`)

This change would fit nicely into this workflow. It could likely be achieved with some scripting and the `--title` PR, but I thought it would be a nice opportunity to follow the existing pattern. Writing to a file also lends itself well to scripting. For example, I might set up a script to automate converting the branch name into a nicely formatted title.

docs/source/cli/github.rst
russellyeo: I noticed I had neglected to add documentation and tests for gitlab. Just pushed another commit to add.

PawelLipski:
>  I also want to thank you for the excellent tool - I now can't imagine working without it.
Thank you!

> Good point. Perhaps my workflow is a little unusual - I use a git alias which opens these files one-by-one to gather my input and then create the PR

Okay, I see... this still looks like a quite unusual approach 😅 As you can imagine, I'm pretty concerned about bloating the interface of git-machete too much, given that there's a minor release (i.e. new features) pretty much every month and [docs are already huuuge](https://git-machete.readthedocs.io/en/stable/) 🫠

May I suggest one of the following:

1. Keep the feature as-is, but skip it from docs. This way it'll cater for your specific workflow, without the extra perceived complexity to `create-{pr,mr}` (esp. important for fresh users). There're already some undocumented features that only I am likely to ever use, like [`GIT_MACHETE_MEASURE_COMMAND_TIME` env var](https://github.com/VirtusLab/git-machete/blob/v3.30.0/git_machete/utils.py#L23).
2. Alternatively, instead of using `.git/info/title` which suggests a static title to be used for all PRs (which is rather impractical and hence potentially confusing for the other users), allow for passing @-prefixed file name as title: `--title=@foo.txt`. This is a pattern not found in git-machete yet, but otherwise pretty well-known from tools like `curl`. This will be less confusing as it suggests that title isn't static but rather dynamically generated.

WDYT?

russellyeo:
@PawelLipski 

Thanks for the suggestions, both options seem like a good compromise. I'm happy with either!  I've reverted the changes to the docs as the simplest route (option 1). I'm also open to discarding all changes altogether, just let me know what your preference is

tests/test_github_create_pr.py
```diff
@@ -954,3 +954,36 @@ def test_github_create_pr_when_base_branch_appeared_on_remote(self, mocker: Mock
...
+
+        assert_success(
+            ['status'],
+            """
+            develop
+            |
+            o-feature *  PR #1 (some_other_user)
+            """,
+        )
+
```
PawelLipski:
```suggestion
```

PawelLipski (approved):

PawelLipski:
Merging as is. Thanks!

PR #1367, opened 2024-11-04
Allow disabling fetch for certain remotes in traverse

gjulianm:
This PR adds a new configuration option, `machete.traverse.fetch.<remote>`, that allows disabling the fetch on certain remotes when running `git machete traverse --fetch` (or `git machete traverse -W`, which includes `--fetch`). 

The motivation: we might have remotes that are temporarily offline, or take a long time to respond, but are not the main remote we use to synchronize. This config allows continued usage of `git machete traverse -W` for the remotes that are working.

### DIFF
--- a/docs/source/cli/config.rst
+++ b/docs/source/cli/config.rst
@@ -52,7 +52,7 @@ Note: ``config`` is not a command as such, just a help topic (there is no ``git 
     .. include:: git-config-keys/traverse_push.rst
 
 ``machete.traverse.fetch.<remote>``:
-    Configures the behavior of ``git machete traverse`` command for the given remote when --fetch flag is used. If set to ``false``, this remote will not fetched before the traversal. The default value of this configuration is ``true``.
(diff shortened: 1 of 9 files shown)


PawelLipski:
Nice! I'm gonna take a closer look tmrw

docs/source/cli/config.rst
```diff
@@ -51,6 +51,9 @@ Note: ``config`` is not a command as such, just a help topic (there is no ``git
 ``machete.traverse.push``:
     .. include:: git-config-keys/traverse_push.rst
 
+``machete.traverse.fetch.<remote>``:
+    Configures the behavior of ``git machete traverse`` command for the given remote when --fetch flag is used. If set to ``false``, this remote will not fetched before the traversal. The default value of this configuration is ``true``.
```
PawelLipski:
```suggestion
    Configures the behavior of ``git machete traverse`` command for the given remote when ``--fetch`` flag is used. If set to ``false``, this remote will not be fetched before the traversal. The default value of this configuration is ``true``. This is useful for excluding remotes that are temporarily offline, or take a long time to respond.
```

PawelLipski:
Ready after a few patches... to be release soon as v3.31.0

PawelLipski (approved):

PR #1392, opened 2025-01-27
Unify test helpers access

### LINKED ISSUE #1131: Unify access to helpers from tests
Some helpers are imported as top-level symbols from `mockers`, and some are available via `self.repo_sandbox`. Let's unify &mdash; probably better to move stuff to top-level symbols wherever possible. Maybe `repo_sandbox` isn't needed at all?

wolepp:
Resolves: https://github.com/VirtusLab/git-machete/issues/1131

All `GitRepositorySandbox` methods are moved to `mockers_git_repository` if they interact with git or to `mockers.py` otherwise (e.g. creating subprocess)


PawelLipski (approved):
Gr8! way easier to read & write now

tests/test_update.py
```diff
@@ -65,56 +73,65 @@ def test_update_by_merge(self) -> None:
...
+            hash3 = get_current_commit_hash()
+            check_out("level-0-branch")
+            commit("New commit on level-0-branch")
+            hash4 = get_current_commit_hash()
+            execute("git cherry-pick level-1-branch")
+
+        print(hash1)
+        print(hash2)
+        print(hash3)
+        print(hash4)
```
PawelLipski: Remove

PR #1503, opened 2025-10-03
fix(gitlab): update glab CLI token regex to support latest PAT format and auth status output format

max-nicholson:
1. PAT format for new tokens contains 3 "." separated segments e.g. "glpat-somechars.[\d\d].somechars", which doesn't work with the current token regex
2. `glab` auth status output has changed the line prefix for the token as of v1.66, which also doesn't work with the current token regex


PawelLipski (approved):
Nice! thanks for also updating the tests

I'll rebase & merge this PR, and I'll open a quick separate PR just to credit you in the release notes

PawelLipski:
To be released soon as a part of v3.37.0

PR #1524, opened 2025-11-03
Implement worktree support for traverse command

### LINKED ISSUE #1522: traverse cannot work with branches already checked out in worktrees
`git traverse -W` refuses to continue when a managed branch is already checked out in another location using git worktree: 
```
fatal: 'feature-branch' is already checked out at 'path/to/the/worktree/'
git checkout --quiet feature-branch -- returned 128
```

When working with stacked PRs or generally multiple PRs (otherwise why would you need git machete in the first place!) it's actually very common to have them checked out in separate worktrees. So this is entirely blocking using `traverse` when worktrees are involved. 

Am I missing something? 

I have two ideas how to solve this:
1. Quick and potentially unsafe: add the option to traverse with `--ignore-other-worktrees` added to git checkout operations. The problem is when the worktrees are in a dirty state. This wouldn't update the state in the worktree itself and might leave an inconsistent state of the branch. Perhaps, only in case of dirty state traverse would refuse to work. 
2. Instead of checking out, traverse could jump cd into the worktree and perform the necessary updates from there.

PawelLipski:
# Based on PR #1528

## Chain of upstream PRs as of 2025-11-14

* PR #1528:
  `develop` ← `refactor/rename-worktree-git-ops`

  * **PR #1524 (THIS ONE)**:
    `refactor/rename-worktree-git-ops` ← `improve/1522-worktree-switch-traverse`

- Add get_worktrees() and get_worktree_path() methods to GitOperations
- Modify traverse to cd into worktrees where branches are checked out
- Cache worktree information to avoid repeated git worktree list calls
- Add informative messages when changing directories to worktrees
- Add comprehensive tests for worktree handling including detached HEAD cases
- Add integration test for traverse with multiple worktrees

Fixes #1522

**TBD**: how about checkout from other commands like `git machete go`? ➡️ leave as is, there's no easy way to switch shell's wd from inside Python process; just fail

**TODO**: mention in docs ➡️ done

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -2,6 +2,7 @@
 
 ## New in git-machete 3.37.1
 
+- improved: `traverse` now changes directory to the worktree where a branch is checked out, rather than failing (suggested by @lsierant)
 - fixed: use `simple` (rather than `none`) mode of squash merge detection in `git machete discover`, to keep parity with the IntelliJ plugin
 
 ## New in git-machete 3.37.0
--- a/docs/man/git-machete.1
+++ b/docs/man/git-machete.1
@@ -2143,6 +2143,8 @@ Operations like \fBgit machete github anno\-prs\fP (\fBgit machete gitlab anno\-
 and \fBgit machete github checkout\-prs\fP (\fBgit machete gitlab checkout\-mrs\fP) add \fBrebase=no push=no\fP branch qualifiers
 when the current user is NOT the author of the PR/MR associated with that branch.
 .sp
+\fBNote on git worktrees:\fP if a branch is already checked out in another worktree, \fBtraverse\fP will change directory to that worktree rather than failing.
+.sp
 \fBOptions:\fP
 .INDENT 0.0
 .TP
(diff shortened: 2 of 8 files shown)


lsierant:
@PawelLipski I've tested it locally and works flawlessly! Thanks! 

The only thing (which I'm not sure it should addressed) is a slight UX/flow difference when the branch in a worktree is dirty and cannot be manipulated (e.g. rebased). 

Without the worktree, after something fails when traversing, git-machete leaves the current tree in the currently processing branch, so it's easy to fix it and proceed with another traverse. 

With worktrees, it cd into a worktree but when it exists with an error, you're still in the current worktree. Slightly annoying, but not blocking! 

Worth considering having this UX unified.

PawelLipski:
> With worktrees, it cd into a worktree but when it exists with an error, you're still in the current worktree. Slightly annoying, but not blocking!

Yeah, I think it's going to be hard to work around, that's more of an OS limitation (Linux/macOS/Windows) &mdash; a process can't change its parent's working directory ([Reddit](https://www.reddit.com/r/linux_programming/comments/1e3tevm/how_to_change_working_dir_of_parent_process_bash/)).

I'll instead make `traverse` print a warning (as the last thing in the output) which encourages the user to `cd` to the worktree corresponding to the checked out branch. The warning will be always printed if the initial WD doesn't correspond to the checked out branch, regardless of whether `traverse` failed or finished successfully.

PawelLipski:
To be released as v3.37.1 soon

PR #1603, opened 2026-03-07
Rename all Python identifiers that had a trailing underscore

PawelLipski:


git_machete/client/base.py
```diff
@@ -455,32 +455,32 @@ def status(
...
-            prefix_dfs(root, accumulated_path_=[])
+            prefix_dfs(root, accumulated_path=[])
 
         out = io.StringIO()
         sync_to_parent_status: Dict[LocalBranchShortName, SyncToParentStatus] = {}
         fork_point_hash_cached: Dict[LocalBranchShortName, Optional[FullCommitHash]] = {}  # TODO (#110): default dict with None
         fork_point_branches_cached: Dict[LocalBranchShortName, List[BranchPair]] = {}
 
-        def fork_point_hash(branch_: LocalBranchShortName) -> Optional[FullCommitHash]:
-            if branch not in fork_point_hash_cached:
```
PawelLipski, line 475: It was a subtle bug - `branch` was always equal to `branch_`, so it never surfaced

PR #1608, opened 2026-03-09
Simplify the newly-split implementation of `status`

PawelLipski:


git_machete/client/traverse.py
```diff
@@ -299,6 +299,7 @@ def traverse(
                             self._state.down_branches_for[upstream] or [])
                         if branch in self._state.annotations:
                             del self._state.annotations[branch]
+                        self._state.managed_branches.remove(branch)
```
PawelLipski, line 302: Subtle bug, the branch was not removed from `managed_branches`

PR #1614, opened 2026-03-23
fix: say 'root' instead of 'first root' when there is only one root

### LINKED ISSUE #1550: Don't use "first root" in user-facing messages when the root is the only one
Just use "root"

At least observed in `traverse` ("Checking out first root"), TBD if anywhere else

mvanhorn:
## Summary

When the machete file has a single root branch, user-facing messages now say "root" instead of "first root". The "first root" qualifier is only used when multiple roots exist and the distinction matters.

## Changes

- `git_machete/client/traverse.py`: conditionally use "first root" or "root" based on `len(self._state.roots)`
- `git_machete/client/base.py`: same conditional for the `PickRoot.FIRST` warning message
- Updated test expectations in `test_traverse.py`, `test_traverse_github.py`, `test_traverse_gitlab.py`, and `test_traverse_worktrees.py` for single-root scenarios

The multi-root test case in `test_traverse.py` (develop + master) still uses "first root branch" since the qualifier is needed there.

Closes #1550

This contribution was developed with AI assistance (Claude Code).

### DIFF
--- a/git_machete/client/base.py
+++ b/git_machete/client/base.py
@@ -706,7 +706,7 @@ class MacheteClient:
                 if if_unmanaged == PickRoot.FIRST:
                     warn(
                         f"{bold(branch)} is not a managed branch, assuming "
-                        f"{self._state.roots[0]} (the first root) instead as root")
+                        f"{self._state.roots[0]} ({'the first root' if len(self._state.roots) > 1 else 'root'}) instead as root")
                     return self._state.roots[0]
                 else:  # if_unmanaged == PickRoot.LAST
                     warn(
--- a/git_machete/client/base.py
+++ b/git_machete/client/base.py
@@ -706,12 +706,12 @@ class MacheteClient:
                 if if_unmanaged == PickRoot.FIRST:
                     warn(
                         f"{bold(branch)} is not a managed branch, assuming "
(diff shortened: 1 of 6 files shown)


PawelLipski:
Thanks for the contribution! I've updated the base branch to `develop` instead of `master` (the latter is used for the released code)

git_machete/client/base.py
```diff
@@ -706,7 +983,7 @@ def root_branch_for(self, branch: LocalBranchShortName, if_unmanaged: PickRoot)
                 if if_unmanaged == PickRoot.FIRST:
                     warn(
                         f"{bold(branch)} is not a managed branch, assuming "
-                        f"{self._state.roots[0]} (the first root) instead as root")
+                        f"{self._state.roots[0]} ({'the first root' if len(self._state.roots) > 1 else 'root'}) instead as root")
```
PawelLipski: The wording in case of the sole root is awkward: _assuming BRANCH (root) instead as root_

Maybe just _assuming BRANCH instead as root_

Similarly below for `PickRoot.LAST`, while you're here

mvanhorn: Fixed in 245ccd2 - when there's a sole root, the message now reads "assuming BRANCH instead as root" without the redundant "(root)" annotation. Also applied the same fix to the `PickRoot.LAST` path.

PawelLipski (approved):

PR #1627, opened 2026-03-31
Add `cd-into-temporary-worktree` value for `machete.traverse.whenBranchNotCheckedOutInAnyWorktree` git config key

### LINKED ISSUE #1588: A different branch is checked out in the main worktree after the traverse
While working with git-machete on the worktree, I noticed that the main branch of the feature chain changed (it was checked out in a non-current directory) during the execution of the traverse command. However, it was not rolled back after execution finished, which can lead to unexpected side effects.

It would be good if traverse returned to the original branch at the end of the command execution

PawelLipski:


git_machete/client/traverse.py
```diff
@@ -515,6 +545,8 @@ def traverse(
                     f"The initial branch {bold(initial_branch)} has been slid out. "
                     f"Returned to nearest remaining managed branch {bold(nearest_remaining_branch)}")
         finally:
+            self._remove_temporary_worktree()
+
```
PawelLipski, line 554: Whoops this is confusing - what if the branch was checked out in a temporary worktree?

PR #1640, opened 2026-04-16
fix(create-pr): skip head-branch-exists check when sync just pushed the branch

### LINKED ISSUE #1591: Don't print `Checking if head branch ... exists in origin remote` after the branch has just been pushed
This check only makes sense for the cases where the branch lives in `.git/refs/remotes` locally, but has been removed from the remote.
This won't happen when the branch has **just** been pushed 😅

tmchow:
## Summary

Skip the redundant `Checking if head branch <X> exists in <remote> remote...` progress line and its accompanying `ls-remote` call in `create-pr` / `create-mr` when `sync_before_creating_pull_request` just pushed the same branch to the code-hosting remote. Addresses #1591.

## Why this matters

`create-pr` runs `sync_before_creating_pull_request` first, which pushes the current branch to the relevant remote if it is `UNTRACKED`, `AHEAD_OF_REMOTE`, or `DIVERGED_FROM_AND_NEWER_THAN_REMOTE`. The inline comment at `git_machete/client/with_code_hosting.py:323` already acknowledges this:

```python
# (sync_before_creating_pull_request ensures the branch is pushed,
#  so if it's missing here, it must have been removed)
```

But the progress line and `ls-remote` still run unconditionally. That is pure noise right after a fresh push: we know the branch is there because this same process just put it there.

The check still makes sense in the other direction, when the remote-tracking ref exists locally but the branch was deleted on the server since the last fetch/push. Only the just-pushed path is redundant.

## Changes

Two files: `git_machete/client/with_code_hosting.py` for the behavior change, `git_machete/cli.py` for the caller wiring.

- `sync_before_creating_pull_request` now returns `Optional[str]`, the remote it pushed to (or `None` if no push happened). Detection works by snapshotting each remote's tracking-ref OID via `get_commit_hash_by_revision` before the `_handle_*_state` call and diffing afterwards. The first remote whose OID changed is the push destination.
- `create_pull_request` accepts `head_pushed_to_remote: Optional[str] = None` and gates the `Checking if head branch ...` skip on `head_pushed_to_remote == head_org_repo_remote.remote`. The kwarg stays optional because `TraverseMacheteClient` also calls `create_pull_request` without a prior `sync_before_creating_pull_request` pass.
- `cli.py` captures the return value from `sync_before_creating_pull_request` and threads it through to `create_pull_request`.
- The base-branch check is unchanged. The fail-fast "branch removed from remote since last fetch/push" path is unchanged.
- In multi-remote setups where `machete.github.remote` / `machete.gitlab.remote` points at a different remote than the one sync pushed to, the remotes do not match, the gate is false, and the existing preflight runs like before.

### Why OID-diff instead of wrapping `self._git.push`

An earlier revision wrapped `self._git.push` to capture the remote name as it was called. That approach conflated two outcomes that look the same from the push wrapper's point of view but differ in reality:

1. `UNTRACKED` branch where the user answers `y` to "Push `<branch>` to `<remote>`?" (push happens, remote ref gets a new OID).
2. `UNTRACKED` branch whose remote counterpart already exists at the same commit, where the user answers `y` to "Set the remote of `<branch>` to `<remote>` without pushing or pulling?" (tracking gets set, no push, remote ref is untouched).

In case 2 the old wrapper also marked the branch as "pushed" and skipped the existence check, which was wrong because the skip was meant as "we just put it there" rather than "the branch happens to be in sync." Comparing remote-tracking OIDs before and after the `_handle_*_state` call makes this distinction trivially: a real push changes the remote-tracking OID, a set-tracking does not.

It also sidesteps the name-mangled `getattr(self, '_MacheteClientWithCodeHosting__head_branch_pushed_to_remote', None)` read that the wrapper approach needed, which `@PawelLipski` flagged as ugly in review.

## Testing

- `pytest tests/test_github_create_pr.py tests/test_gitlab_create_mr.py tests/test_traverse_github.py` -> 36 passed. Existing tests that set up a branch already on the remote expect both progress lines and still see them. Existing tests whose scenarios push an untracked branch in-line had the now-redundant `Checking if {head_branch_name} branch ...` line removed from their expected output (one removed line per affected test, no assertion semantics changed beyond the string).
- `test_github_create_pr_skips_head_branch_check_after_sync_push` covers the single-remote happy path: branch is `UNTRACKED`, sync pushes it, output contains `Checking if base branch ...` but not `Checking if head branch ...`, and the PR still lands.
- The existing `test_github_create_pr_with_multiple_non_origin_remotes` test covers the set-tracking-without-push case: user answers `y` to "Set remote of feature to origin_1 without pushing or pulling", output still contains `Checking if head branch feature exists in origin_1 remote... YES`. This test would have failed under the old monkey-patch approach.
- New test covering the multi-remote-mismatch case: two remotes configured, `machete.github.remote` points at remote B, sync pushes to remote A, so the gate is false and the `Checking if head branch ...` preflight still runs.
- Full suite (excluding env-specific zsh-completion tests): 379 passed, 1 skipped.

Fixes #1591

This contribution was developed with AI assistance (Codex).

[Compound Engineering](https://github.com/EveryInc/compound-engineering-plugin)

### DIFF
--- a/git_machete/client/with_code_hosting.py
+++ b/git_machete/client/with_code_hosting.py
@@ -117,6 +117,7 @@ class MacheteClientWithCodeHosting(StatusMacheteClient):
 
     def sync_before_creating_pull_request(self, *, opt_yes: bool) -> None:
         spec = self.code_hosting_spec
+        self.__head_branch_pushed_by_sync = False
         self.expect_at_least_one_managed_branch()
         self._set_empty_line_status()
 
@@ -157,32 +158,43 @@ class MacheteClientWithCodeHosting(StatusMacheteClient):
             SyncToRemoteStatus.DIVERGED_FROM_AND_NEWER_THAN_REMOTE)
         if s in statuses_to_push:
             if current_branch not in self.annotations or self.annotations[current_branch].qualifiers.push:
-                if s == SyncToRemoteStatus.AHEAD_OF_REMOTE:
-                    assert remote is not None
-                    self._handle_ahead_state(
-                        current_branch=current_branch,
-                        remote=remote,
-                        is_called_from_traverse=False,
-                        opt_push_tracked=True,
-                        opt_yes=opt_yes)
-                elif s == SyncToRemoteStatus.UNTRACKED:
-                    self._handle_untracked_state(
-                        branch=current_branch,
-                        is_called_from_traverse=False,
-                        is_called_from_code_hosting=True,
-                        opt_push_tracked=True,
-                        opt_push_untracked=True,
-                        opt_yes=opt_yes)
-                elif s == SyncToRemoteStatus.DIVERGED_FROM_AND_NEWER_THAN_REMOTE:
-                    assert remote is not None
-                    self._handle_diverged_and_newer_state(
-                        current_branch=current_branch,
-                        remote=remote,
-                        is_called_from_traverse=False,
-                        opt_push_tracked=True,
-                        opt_yes=opt_yes)
-                else:
-                    raise UnexpectedMacheteException(f"Invalid sync to remote status: `{s}`.")
+                original_push = self._git.push
+
+                def push_and_mark(*args, **kwargs):  # type: ignore[no-untyped-def]
+                    result = original_push(*args, **kwargs)
+                    self.__head_branch_pushed_by_sync = True
+                    return result
+
+                self._git.push = push_and_mark  # type: ignore[method-assign]
+                try:
+                    if s == SyncToRemoteStatus.AHEAD_OF_REMOTE:
+                        assert remote is not None
+                        self._handle_ahead_state(
+                            current_branch=current_branch,
+                            remote=remote,
+                            is_called_from_traverse=False,
+                            opt_push_tracked=True,
+                            opt_yes=opt_yes)
+                    elif s == SyncToRemoteStatus.UNTRACKED:
+                        self._handle_untracked_state(
+                            branch=current_branch,
+                            is_called_from_traverse=False,
+                            is_called_from_code_hosting=True,
+                            opt_push_tracked=True,
+                            opt_push_untracked=True,
+                            opt_yes=opt_yes)
+                    elif s == SyncToRemoteStatus.DIVERGED_FROM_AND_NEWER_THAN_REMOTE:
+                        assert remote is not None
+                        self._handle_diverged_and_newer_state(
+                            current_branch=current_branch,
+                            remote=remote,
+                            is_called_from_traverse=False,
+                            opt_push_tracked=True,
+                            opt_yes=opt_yes)
+                    else:
+                        raise UnexpectedMacheteException(f"Invalid sync to remote status: `{s}`.")
+                finally:
+                    self._git.push = original_push  # type: ignore[method-assign]
 
                 self._print_new_line(False)
                 self.status(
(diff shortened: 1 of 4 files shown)


git_machete/client/with_code_hosting.py
```diff
@@ -157,32 +158,43 @@ def sync_before_creating_pull_request(self, *, opt_yes: bool) -> None:
...
-                else:
-                    raise UnexpectedMacheteException(f"Invalid sync to remote status: `{s}`.")
+                original_push = self._git.push
+
+                def push_and_mark(*args, **kwargs):  # type: ignore[no-untyped-def]
+                    result = original_push(*args, **kwargs)
+                    self.__head_branch_pushed_to_remote = args[0] if args else kwargs.get('remote')
+                    return result
+
+                self._git.push = push_and_mark  # type: ignore[method-assign]
```
PawelLipski: 😳

git_machete/client/with_code_hosting.py
```diff
@@ -231,6 +243,8 @@ def create_pull_request(
             opt_update_related_descriptions: bool,
             opt_yes: bool
     ) -> None:
+        pushed_remote = getattr(self, '_MacheteClientWithCodeHosting__head_branch_pushed_to_remote', None)
```
PawelLipski: Whoops this is ugly :/

tmchow:
@PawelLipski good call - the monkey-patching was the wrong tool here 🤦‍♂️ Pushed a refactor that replaces the `self._git.push` wrapper with a simple OID snapshot: we capture each remote's tracking-ref OID before the `_handle_*` call and diff afterwards to identify which remote (if any) got the push. lmk what you think!

A nice side effect: the OID-diff approach correctly distinguishes a real push from the "set tracking remote without pushing" case (untracked branch whose remote counterpart already exists at the same commit), which the previous monkey-patch treated identically. Also got rid of the `getattr(self, '_MacheteClientWithCodeHosting__head_branch_pushed_to_remote', None)` name-mangling escape hatch - `sync_before_creating_pull_request` now returns `Optional[str]` and `create_pull_request` takes it as a kwarg.

PR description was also updated to reflect the new approach. ran the full test suite (excluding env-specific zsh-completion tests) and it all passed.

git_machete/client/with_code_hosting.py
```diff
@@ -263,30 +292,40 @@ def create_pull_request(
...
+            head_pushed_to_remote is not None and head_pushed_to_remote == head_org_repo_remote.remote
+        )
 
         # Check both base and head branches in a single git ls-remote call if they use the same remote
         if base_org_repo_remote.remote == head_org_repo_remote.remote:
-            print_fmt(f"Checking if {spec.head_branch_name} branch <b>{head}</b> "
-                      f"exists in <b>{head_org_repo_remote.remote}</b> remote... ",
-                      newline=False)
+            if head_branch_was_pushed_to_code_hosting_remote:
+                base_branch_found_on_remote = self._git.does_remote_branch_exist(base_org_repo_remote.remote, base)
```
PawelLipski, line 302: Hmm in this case, the blocking operation of checking if the remote branch exists still needs to be performed 🤔 so the message `print_fmt(f"Checking if {spec.base_branch_name} branch <b>{base}</b> "...` still needs to be printed beforehand 🤔

PawelLipski (commented):
I'm actually starting to think that this optimization might not be worth adding the extra complexity in the already complex codebase 😳 esp. given that determining whether the branch has been pushed isn't trivial

tmchow:
@PawelLipski Fair point. It does add more complexity, and you're right that the blocking check + its print still need to stay (line 302), so the win shrinks further. 

Happy to close this out if you'd rather not carry it. lmk

PawelLipski:
Yeah, sorry about that - I've underestimated how much complexity it adds for a cosmetic gain :/ lemme close the PR and the issue as _won't fix_

PR #1701, opened 2026-05-18
Show actionable error when target branch is locked by another worktree

### LINKED ISSUE #1696: go interactive: cannot go into a checked out worktree
### Description

This is partially a bug and a feature request. When one of the machete-tracked branches is checked out in a worktree, machete should cd into that directory instead of trying to check out in current dir (which results in an error normally):
```
Checking out lsierant/KUBE-26-background-tester... fatal: 'lsierant/KUBE-26-background-tester' is already used by worktree at '/Users/lukasz.sierant/mdb/lsierant_KUBE-26-background-tester'
git -c log.showSignature=false checkout --quiet lsierant/KUBE-26-background-tester -- returned 128
```

We should automatically cd into worktree dir. 

Additionally I have a feature request: please provide a way for defining a post-cd hook. In my case I would like to automatically source venv whenever I jump into a specific worktree. But perhaps this should be handle by external tools like direnv...

Supporting worktrees correctly is more and more important as agents nowadays often works in worktrees. Having a way to switch between worktrees would make git machete go an ultimate worktree switcher.

PS. I discovered interactive mode in git machete go completely by accident (I just forgot the nav commands) and wow there is exactly the feature I wanted to have here for a long time now! Thanks!  

### Steps to reproduce

1. git machete go
2. select a branch that is checked out in another worktree

### Expected behavior

We should cd into that directory instead of erroring out

### Actual behavior

Error as in the description

### git version

git version 2.50.1 (Apple Git-155)

### git machete version

git version 2.50.1 (Apple Git-155)

### Operating system

macOS 26

### Verbose / debug output (optional)

```false

```

### Additional context

_No response_

PawelLipski:
## Summary

`git machete go`, `git machete slide-out` and `git machete {github,gitlab} checkout-{pr,mr}s` used to fail with the cryptic `fatal: '<branch>' is already used by worktree at <path>` (exit 128) from `git checkout` when the target branch was already checked out in another linked worktree.
They now fail with a clean machete-level error that names the locking worktree and points the user at the two reasonable next steps:

> Branch `<branch>` is already checked out in another worktree at `<path>`.
> Run `cd <path>` to work on it there, or `git worktree remove <path>` to drop that worktree.

## Why not just `os.chdir` into the linked worktree (the obvious "fix")?

Because we can't.

`os.chdir` from a Python subprocess (which `git machete` is) mutates the cwd of *that subprocess only*.
The instant the process exits, the user's shell is back where it started - no portable way to "export the cd back to the shell" without involving a shell function wrapper (`eval "$(git machete go --print-cd)"` style), which we don't ship.

So a silent chdir + exit-0 would:

- mislead **interactive users** (`git machete go down` claims success, prints "Changing directory to ...", exits - and the user's shell cwd is unchanged, with no branch newly checked out anywhere they can see),
- silently break **agents / CI** (exit 0 implies "checkout succeeded" - caller then runs `git status` / `git log` against the *wrong* branch in the *wrong* worktree, with no signal that anything went sideways).

Failing with a clear, actionable, non-zero-exit error is strictly safer.

## Factoring

`Git` stays machete-agnostic and exposes two methods:

- `Git.checkout(branch)` - plain `git checkout <branch>`, unchanged from before.
- `Git.checkout_in_current_worktree(branch)` - new; runs `git worktree list --porcelain` itself, raises the actionable `MacheteException` above if `<branch>` is held by another linked worktree, otherwise delegates to `Git.checkout`.

Callers:

- `go`, `slide-out`, `{github,gitlab} checkout-{pr,mr}s` use `checkout_in_current_worktree` - one-shot commands that must not silently land in the wrong worktree.
- `traverse` keeps using plain `Git.checkout`.
  It already maintains its own up-front worktree snapshot (`TraverseMacheteClient.__worktree_root_dir_for_branch`) and pre-routes any branch held by another worktree into its existing chdir path inside `_switch_branch` - so the guard in `checkout_in_current_worktree` would be redundant there.
  Traverse's chdir behavior is intentionally kept: it's a long-running session, subsequent `git rebase` / `git merge` / `git push` calls within the same process do observe the new cwd, and traverse already prints a hint reminding the user to `cd` themselves at the end of the loop.

## Test plan

- [x] New `tests/test_go.py::TestGo::test_go_with_worktree` - asserts the new actionable `MacheteException` text, asserts `os.getcwd()` is unchanged, asserts HEAD is unchanged.
- [x] New `tests/test_slide_out.py::TestSlideOut::test_slide_out_fails_when_new_parent_is_in_another_worktree` - same shape for the slide-out path.
- [x] Existing `tests/test_traverse.py` + `tests/test_traverse_worktrees.py` - traverse's chdir-into-linked-worktree behavior preserved.
- [x] Existing `tests/test_github_checkout_prs.py` + `tests/test_gitlab_checkout_mrs.py` - exercise the new `checkout_in_current_worktree` path; all green.
- [x] `tox -e mypy` clean.
- [x] CI matrix (Python 3.6 + git 1.8.0 through Python 3.14 + git 2.51.0).

Reported by @lsierant.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -27,6 +27,20 @@
 
 - Don't `git commit` or `git push` unless explicitly asked.
 
+## Tests
+
+- Skip tests that require a minimum Git version with `@pytest.mark.skipif(get_git_version() < (X, Y), reason="...")`,
+  not with an `if get_git_version() < (X, Y): return` early-return at the top of the test body.
+  The decorator surfaces the skip in pytest's report (and in the JUnit XML CI uploads); the early-return silently masquerades as a pass.
+
+## Comments
+
+- Don't add code comments that narrate test-harness internals or explain why an assertion's expected value was massaged
+  to match the harness's output (e.g. "backticks are stripped by `_fmt` in ASCII mode, so the expected string omits them",
+  "the harness lower-cases this, so we compare against lower-case", etc.).
+  If the actual and expected values match, the assertion already documents itself; if they don't, fix the production code or the harness, don't annotate the workaround.
+  This generalizes: avoid comments that exist solely to justify a specific literal in a test - the test name and the assertion are the contract.
+
 ## Formatting
 
 - No trailing whitespace on any line.
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -2,6 +2,8 @@
 
 ## New in git-machete 3.41.1
 
+- fixed: `git machete go`, `slide-out` and `{github,gitlab} checkout-{pr,mr}s` now fail with an actionable error when the target branch is already checked out in another worktree, rather than the cryptic `fatal: '<branch>' is already used by worktree at <path>` from `git checkout` (reported by @lsierant)
+
 ## New in git-machete 3.41.0
 
 - added: an [Agent Skill](https://agentskills.io/) at `skills/git-machete/SKILL.md`
--- a/git_machete/client/go_interactive.py
+++ b/git_machete/client/go_interactive.py
@@ -132,7 +132,7 @@ class GoInteractiveMacheteClient(StatusMacheteClient):
         if selected_branch is not None and selected_branch != current_branch:
             print()
             print_fmt(f"Checking out <b>{selected_branch}</b>... ", newline=False)
-            self._git.checkout(selected_branch)
+            self._git.checkout_in_current_worktree(selected_branch)
             print_fmt(green_ok())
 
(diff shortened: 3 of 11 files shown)


lsierant (commented):
Thanks for the explanations and the better error!

PawelLipski:
> without involving a shell function wrapper

Actually, it'd be even worse 😅 we'd need to convert `git` itself to shell function, otherwise invocations like `git machete` (as opposed to `git-machete`) wouldn't work either 🫠

PR #1728, opened 2026-06-18
Add config keys to locate a PR/MR base repository independently of the head

PawelLipski:
# Based on PR #1733

## Chain of upstream PRs as of 2026-06-23

* PR #1734:
  `develop` ← `docs/githublab-access-keys`

  * PR #1733:
    `docs/githublab-access-keys` ← `fix/hublab-remote-logic`

    * **PR #1728 (THIS ONE)**:
      `fix/hublab-remote-logic` ← `feature/base-githublab-config-keys`

Introduce `machete.github.{baseRemote,baseOrganization,baseRepository}` and `machete.gitlab.{baseRemote,baseNamespace,baseProject}` git config keys to locate the base/target repository of a pull/merge request independently of the head/source repository (e.g. when the base lives in an upstream repository and the head in a fork).

Each base key defaults to its non-base counterpart when unset and takes priority over the otherwise-inferred value when set; the organization/namespace and repository/project keys can be overridden independently of each other.


docs/source/git-config-keys/github_access.rst
```diff
@@ -19,6 +19,22 @@
     The GitHub repository (the part after ``/`` in ``organization/repository``); otherwise inferred from the remote URL.
     For example, ``git config machete.github.repository git-machete``
 
+``machete.github.baseRemote``
+    Like ``machete.github.remote``, but used to locate the base repository that the pull request targets,
+    which may differ from the head repository (for example, the base in an upstream repository and the head in a fork).
+    Defaults to ``machete.github.remote`` when unset, and takes priority over the otherwise-inferred value when set.
```
PawelLipski: Discrepancy: `.remote` is used to derive other keys; `.baseRemote` isn't used for anything else 🤔

PawelLipski: `.domain` must be the same for both repos, so we can assume it's the one taken from `.remote`; the same domain should be derived from `.baseRemote` anyway (we can check for that explicitly for super-purity; if they differ, we tell the user to set `.domain` explicitly - this is a super-unlikely case, maybe in case of migration from github.com to self-hosted)

PawelLipski: And `.baseOrg` and `.baseRepo` can be derived from `.baseRemote` just as `.remote` is; with **no** fallback to `.org` and `.repo`. And they'll need to be set both at the same time (identical with `.org`/`.repo`).

The only fallback then would effectively be `.baseRemote` to `.remote`.

"➡️" means "affects" 
`.remote` & `.baseRemote` ➡️ `.domain`
`.remote` ➡️ `.org`, `.repo`; `.baseRemote` (fallback)
`.baseRemote` ➡️ `.baseOrg`, `.baseRepo`

Minor: `.domain` can go to the top of the list in docs in the base PR

PR #1740, opened 2026-07-06
fix: washed-out yellow on light-background terminals

### LINKED ISSUE #380: Does yellow edge really look yellow in white-background terminals?
In IntelliJ, it looks kinda okay:

image

But see GNOME terminal:

image

Let's think of some another color that could still be called yellow &mdash; both in dark- and light-themed terminals (if that's even possible to achieve...)

joshazmy:
## Summary
`FullTerminalAnsiOutputCodes.YELLOW` used plain SGR 33 ("dark yellow"), which renders as a washed-out, barely-visible color on light/white terminal backgrounds. This color was already fixed once, to the brighter 256-color gold1 (`\033[00;38;5;220m`), in 0b43361 ("Improved visibility on white terminal backgrounds"), but got silently overwritten back to plain SGR 33 by 21f9b8a, which was actually only trying to match a *different* yellow use (the hook-ignored-hook hint) to git's own advice color.

This restores the 256-color gold1 for full-fledged terminals. 8-color terminals (`BasicTerminalAnsiOutputCodes`) can't render the 256-color escape, so they keep plain SGR 33 explicitly rather than inheriting the 256-color value.

Fixes #380

## Test plan
- Added `test_yellow_is_not_the_washed_out_dark_yellow` regression test in `tests/test_utils.py`
- Ran `tests/test_utils.py` — 8 passed, 2 skipped (platform-specific, unrelated)
- Ran the full color-related suite (`test_status.py`, `test_discover.py`, `test_advance.py`, `test_fork_point.py`, `test_go_interactive.py`, `test_traverse.py`) — all pass except 18 pre-existing failures unrelated to this change (verified: identical failures reproduce against unmodified `develop`, caused by `shutil.copytree` failing on a `.git/fsmonitor--daemon.ipc` unix socket in this sandbox, not by this fix)

### DIFF
--- a/git_machete/utils/terminal.py
+++ b/git_machete/utils/terminal.py
@@ -86,7 +86,12 @@ class FullTerminalAnsiOutputCodes:
     DIM = '\033[2m'
     UNDERLINE = '\033[4m'
     GREEN = '\033[32m'
-    YELLOW = '\033[33m'
+    # Plain SGR 33 ("dark yellow") renders as a washed-out, barely-visible color on light/white-background
+    # terminals (#380). Gold1 from the 256-color palette reads clearly as "yellow" on both dark and light
+    # backgrounds; it was already used for exactly this reason back in 0b43361 ("Improved visibility on white
+    # terminal backgrounds"), but got silently overwritten by 21f9b8a while that commit was actually only
+    # trying to match a *different* yellow use (the hook-ignored-hook hint below) to git's own advice color.
+    YELLOW = '\033[00;38;5;220m'
     ORANGE = '\033[00;38;5;208m'
     RED = '\033[91m'
     REVERSE_VIDEO = '\033[7m'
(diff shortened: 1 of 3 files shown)


git_machete/utils/terminal.py
```diff
@@ -86,7 +86,12 @@ class FullTerminalAnsiOutputCodes:
     DIM = '\033[2m'
     UNDERLINE = '\033[4m'
     GREEN = '\033[32m'
-    YELLOW = '\033[33m'
+    # Plain SGR 33 ("dark yellow") renders as a washed-out, barely-visible color on light/white-background
```
PawelLipski, line 89: The "SGR" acronym isn't well-known, pls expand (at least in one place)

git_machete/utils/terminal.py
```diff
@@ -86,7 +86,12 @@ class FullTerminalAnsiOutputCodes:
...
+    # terminals (#380). Gold1 from the 256-color palette reads clearly as "yellow" on both dark and light
+    # backgrounds; it was already used for exactly this reason back in 0b43361 ("Improved visibility on white
+    # terminal backgrounds"), but got silently overwritten by 21f9b8a while that commit was actually only
+    # trying to match a *different* yellow use (the hook-ignored-hook hint below) to git's own advice color.
```
PawelLipski, line 93: Pls do not describe the historical state in the comments, just the current one (to reduce confusion/reduce mental load)

PawelLipski:
Thanks for the contribution - could you actually attach the screens of before and after this change on your terminal?

PawelLipski:
No response from the author, closing

PR #1755, opened 2026-07-21
Fix discover showing different branch trees across worktrees

### LINKED ISSUE #1754: git machete `discover`/`status` shows different branch trees across worktrees
### Description

git machete `discover`/`status` shows different branch trees across worktrees

### Steps to reproduce

**1.** In the **main worktree**, checked out on `branchB`, run `git branch`:

```
  branchA
* branchB
  branchC
  branchD
+ branchE
```

(`branchE` is checked out in a secondary worktree, indicated by `+`.)

**2.** Still in the main worktree, run `git machete discover` followed by `git machete status`:

```
Discovered tree of branch dependencies:
  main
  │
  └─branchA (untracked)
  │
  └─branchB [<this worktree>] (diverged from origin)
  │ │
  │ └─branchC
  │   │
  │   ├─branchD (untracked)
```

**3.** Note that `branchE` is not listed at all in the discovered tree.

**4.** Switch to the **secondary worktree** (where `branchE` is checked out).

**5.** Run `git machete discover` and `git machete status` again:

```
  main
  │
  └─branchE [<this worktree>] (untracked)
  │
  └─branchB [<main worktree>] (diverged from origin)
```

### Expected behavior

I'm not sure whether it's intentional that git machete shares the main worktree's `.git/machete` file across all worktrees, so expected behavior depends on the intent:

1. If sharing is **not** intentional, then running `git machete discover` in another worktree should update only *that* worktree's own `.git/machete` file, rather than overwriting the shared one.
2. If sharing **is** intentional, then the discovered tree should show all branches across all worktrees, fully expanded, regardless of which worktree the command is run from.
3. Either way, `git machete discover` run from the main worktree should include branches checked out in other worktrees, even if it collapses them.

### Actual behavior

The discovered dependency tree differs depending on which worktree the command is run from. In the main worktree, `discover` finds `branchA`, `branchB`, `branchC`, and `branchD`, but not `branchE`. In the secondary worktree, it finds `branchE` and `branchB`, but drops `branchA`, `branchC`, and `branchD`.

### git version

git version 2.50.1 (Apple Git-155)

### git machete version

git-machete version 3.44.1

### Operating system

macOS Tahoe 26.5.2 (25F84)

### Verbose / debug output (optional)

```false

```

### Additional context

_No response_

earfman:
Fixes #1754.

## Problem

`git machete discover` produces a different branch tree depending on which worktree it is run from, even though all worktrees share one `.git/machete` file by default (`machete.worktree.useTopLevelMacheteFile=true`). Running `discover` in one worktree can silently drop branches that are visible from another.

## Root cause

`discover` trims its output to the ~10 most-recently-checked-out branches, ranked via `Git.get_latest_checkout_timestamps()`. That method parsed `git reflog show` for `HEAD`, but `HEAD`'s reflog is per-worktree (`<git-dir>/logs/HEAD` for the main worktree, `<git-dir>/worktrees/<id>/logs/HEAD` for each linked one). So the recency ranking only saw the current worktree's checkouts; branches recently used in other worktrees were treated as stale and pruned, differently depending on where the command ran.

## Fix

Aggregate every worktree's `HEAD` reflog (`<git-dir>/logs/HEAD` plus `worktrees/*/logs/HEAD`), keeping the most recent checkout timestamp per branch. The ranking is now worktree-independent, so `discover` produces the same, complete tree regardless of the invoking worktree — keeping the shared-file behaviour (intentional and documented) while making the tree consistent, per the reporter's points 2 and 3.

The reflog files are read directly (a single `git reflog show` can only report the current worktree's HEAD), decoded as strict UTF-8 with `errors="surrogateescape"` so a non-ASCII committer name can't raise mid-read.

## Testing

New regression test `test_discover_is_worktree_independent` asserts that `discover` yields the same managed branches from both the main and a linked worktree (using `--checked-out-since` for a deterministic trigger); it fails without the fix and passes with it. The existing discover/git/worktree suites pass locally. Added a `RELEASE_NOTES.md` entry under 3.44.2.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -2,7 +2,7 @@
 
 ## New in git-machete 3.44.2
 
-- fixed: `git machete discover` no longer produces a different branch tree depending on which worktree it is run from; the fresh-branch recency ranking now aggregates HEAD reflogs across all worktrees rather than only the current one (contributed by @earfman)
+- fixed: `git machete discover` no longer produces a different branch tree depending on which worktree it is run from; the fresh-branch recency ranking now aggregates HEAD reflogs across all worktrees rather than only the current one (reported by @jasonoura, contributed by @earfman)
 
 ## New in git-machete 3.44.1
 
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -2,7 +2,7 @@
 
 ## New in git-machete 3.44.2
 
-- fixed: `git machete discover` (and `status`) no longer produce a different branch tree depending on which worktree they are run from; the fresh-branch recency ranking now aggregates HEAD reflogs across all worktrees rather than only the current one (#1754)
+- fixed: `git machete discover` no longer produces a different branch tree depending on which worktree it is run from; the fresh-branch recency ranking now aggregates HEAD reflogs across all worktrees rather than only the current one (contributed by @earfman)
 
 ## New in git-machete 3.44.1
 
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -2,6 +2,8 @@
 
 ## New in git-machete 3.44.2
 
+- fixed: `git machete discover` (and `status`) no longer produce a different branch tree depending on which worktree they are run from; the fresh-branch recency ranking now aggregates HEAD reflogs across all worktrees rather than only the current one (#1754)
+
 ## New in git-machete 3.44.1
 
 - fixed: `git machete help config` (and `help status`) no longer renders stray underlines past the `machete.status.extraSpaceBeforeBranchName` section
--- a/git_machete/git.py
+++ b/git_machete/git.py
@@ -1460,22 +1460,56 @@ class Git:
             raise UnexpectedMacheteException(f"Cannot parse timespec: `{date}`")
 
     def get_latest_checkout_timestamps(self) -> Dict[str, int]:  # TODO (#110): default dict with 0
-        # Entries are in the format '<branch_name>@{<unix_timestamp> <time-zone>}'
-        result = {}
-        # %gd - reflog selector (HEAD@{<unix-timestamp> <time-zone>} for `--date=raw`;
-        #   `--date=unix` is not available on some older versions of git)
-        # %gs - reflog subject
-        output = self._popen_git("reflog", "show", "--format=%gd:%gs", "--date=raw").stdout
-        for entry in get_non_empty_lines(output):
-            pattern = "^HEAD@\\{([0-9]+) .+\\}:checkout: moving from (.+) to (.+)$"  # noqa: FS003
-            match = re.search(pattern, entry)
-            if match:
-                from_branch = match.group(2)
-                to_branch = match.group(3)
-                # Only the latest occurrence for any given branch is interesting
-                # (i.e. the first one to occur in reflog)
-                if from_branch not in result:
-                    result[from_branch] = int(match.group(1))
-                if to_branch not in result:
-                    result[to_branch] = int(match.group(1))
+        # HEAD's reflog is *per-worktree*: the main worktree records checkouts in `<git-dir>/logs/HEAD`,
+        # while each linked worktree keeps its own `<git-dir>/worktrees/<id>/logs/HEAD`. A plain
+        # `git reflog show` therefore only sees checkouts made in the *current* worktree.
+        # `discover` uses these timestamps to pick the ~N most-recently-checked-out branches, so reading a
+        # single worktree's reflog makes that selection depend on where the command runs - producing a
+        # different discovered tree in each worktree over a shared branch layout file (issue #1754).
+        # Aggregate every worktree's HEAD reflog and keep the most recent checkout timestamp per branch.
+        result: Dict[str, int] = {}
+
+        # All per-worktree HEAD reflogs live under the *common* git dir (`get_main_worktree_git_dir`
+        # collapses a linked worktree's `.git/worktrees/<id>` back to the shared `.git`).
+        reflog_paths = [self.get_main_worktree_git_subpath("logs", "HEAD")]
+        worktrees_dir = self.get_main_worktree_git_subpath("worktrees")
+        if os.path.isdir(worktrees_dir):
+            for worktree_id in sorted(os.listdir(worktrees_dir)):
+                reflog_paths.append(self.get_main_worktree_git_subpath("worktrees", worktree_id, "logs", "HEAD"))
+
+        # Raw reflog line: "<old-sha> <new-sha> <name> <email> <unix-ts> <tz>\t<subject>".
+        # We read the files directly (rather than N `git reflog show` subprocesses, one per worktree) since
+        # `git reflog show` can only report the current worktree's HEAD; the on-disk format is stable.
+        pattern = re.compile("^checkout: moving from (.+) to (.+)$")
+        for reflog_path in reflog_paths:
+            try:
+                # Decode as UTF-8 independently of the locale (matching the old `git reflog show`
+                # subprocess path, whose bytes were `.decode("utf-8")`d), and use surrogateescape so a
+                # committer name with invalid UTF-8 bytes can never raise mid-read - the ASCII timestamp
+                # and `checkout:` subject we actually parse are unaffected.
+                with open(reflog_path, encoding="utf-8", errors="surrogateescape") as reflog_file:
+                    lines = reflog_file.readlines()
+            except OSError:
+                # A worktree may have no HEAD reflog yet (e.g. just created); skip it.
+                continue
+            for line in lines:
+                meta, tab, subject = line.partition("\t")
+                if not tab:
+                    continue
+                match = pattern.match(subject.strip())
+                if not match:
+                    continue
+                meta_fields = meta.split()
+                # `<unix-ts> <tz>` are the last two whitespace-separated fields before the tab,
+                # so `[-2]` is the timestamp regardless of spaces in the committer name.
+                try:
+                    timestamp = int(meta_fields[-2])
+                except (IndexError, ValueError):
+                    continue
+                for branch in (match.group(1), match.group(2)):
+                    # Keep the most recent checkout per branch across *all* worktrees. Comparing timestamps
+                    # (rather than "first entry wins") is order-independent, so it's correct even though the
+                    # reflog files are appended oldest-first - the opposite order to `git reflog show`.
+                    if timestamp > result.get(branch, 0):
+                        result[branch] = timestamp
         return result
--- a/git_machete/git.py
+++ b/git_machete/git.py
@@ -1459,57 +1459,43 @@ class Git:
         except (UnderlyingGitException, ValueError):
             raise UnexpectedMacheteException(f"Cannot parse timespec: `{date}`")
 
-    def get_latest_checkout_timestamps(self) -> Dict[str, int]:  # TODO (#110): default dict with 0
+    def get_latest_checkout_timestamps(self) -> Dict[str, int]:
         # HEAD's reflog is *per-worktree*: the main worktree records checkouts in `<git-dir>/logs/HEAD`,
         # while each linked worktree keeps its own `<git-dir>/worktrees/<id>/logs/HEAD`. A plain
-        # `git reflog show` therefore only sees checkouts made in the *current* worktree.
-        # `discover` uses these timestamps to pick the ~N most-recently-checked-out branches, so reading a
-        # single worktree's reflog makes that selection depend on where the command runs - producing a
-        # different discovered tree in each worktree over a shared branch layout file (issue #1754).
-        # Aggregate every worktree's HEAD reflog and keep the most recent checkout timestamp per branch.
+        # `git reflog show` therefore only sees checkouts made in the *current* worktree, so `discover`
+        # (which uses these timestamps to pick the ~N most-recently-checked-out branches) would produce a
+        # different tree depending on where it runs, over a shared branch layout file (issue #1754).
+        # Run `git reflog show` once per worktree directory and aggregate, keeping the most recent
+        # checkout timestamp per branch.
         result: Dict[str, int] = {}
+        # %gd - reflog selector (HEAD@{<unix-timestamp> <time-zone>} for `--date=raw`;
+        #   `--date=unix` is not available on some older versions of git)
+        # %gs - reflog subject
+        pattern = re.compile(r"^HEAD@\{([0-9]+) .+\}:checkout: moving from (.+) to (.+)$")  # noqa: FS003
 
-        # All per-worktree HEAD reflogs live under the *common* git dir (`get_main_worktree_git_dir`
-        # collapses a linked worktree's `.git/worktrees/<id>` back to the shared `.git`).
-        reflog_paths = [self.get_main_worktree_git_subpath("logs", "HEAD")]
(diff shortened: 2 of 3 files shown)


RELEASE_NOTES.md
```diff
@@ -2,6 +2,8 @@
 
 ## New in git-machete 3.44.2
 
+- fixed: `git machete discover` (and `status`) no longer produce a different branch tree depending on which worktree they are run from; the fresh-branch recency ranking now aggregates HEAD reflogs across all worktrees rather than only the current one (#1754)
```
PawelLipski: Pls credit yourself :) see the earlier entries

PawelLipski: And we don't include issue/PR numbers here

PawelLipski:
Thanks for the contrib! pls fix mypy so that we can also validate test coverage (will be posted automatically by codecov once the tests pass)

And also, why does PR title/description mention status? AFAICS `status` command itself isn't affected - does it meant the status that's printed by `discover` itself? still, I'd remove the reference

git_machete/git.py
```diff
@@ -1460,22 +1460,56 @@ def get_git_timespec_parsed_to_unix_timestamp(self, date: str) -> int:
...
+        result: Dict[str, int] = {}
+
+        # All per-worktree HEAD reflogs live under the *common* git dir (`get_main_worktree_git_dir`
+        # collapses a linked worktree's `.git/worktrees/<id>` back to the shared `.git`).
+        reflog_paths = [self.get_main_worktree_git_subpath("logs", "HEAD")]
+        worktrees_dir = self.get_main_worktree_git_subpath("worktrees")
+        if os.path.isdir(worktrees_dir):
+            for worktree_id in sorted(os.listdir(worktrees_dir)):
+                reflog_paths.append(self.get_main_worktree_git_subpath("worktrees", worktree_id, "logs", "HEAD"))
+
```
PawelLipski: Note that depending on `machete.worktree.useTopLevelMacheteFile` git config key,  machete file can be shared b/w worktrees (default), or be stored per-worktree. Still, I think it's reasonable to assume that discovery can happen based on all the reflogs combined, regardless of what worktree it is run from. Worst case, when `useTopLevelMacheteFile=false`, too many branches will be taken into account in discovery and the user will need to trim down the per-worktree machete file.


Also, when `useTopLevelMacheteFile=false`, does `discover` write to the per-worktree machete file (for the worktree where `discover` is executed)? I'd assume so, but pls verify while you're here

earfman:
Thanks @PawelLipski, and thanks for the collaborator invite! Pushed two commits that address everything:

**mypy** — the failure was a bare `set` return annotation on a helper in the new test; changed it to `Set[str]`, so the suite should go green now (and codecov can post coverage).

**Release note** — credited myself and dropped the issue number, matching the earlier entries.

**`status` reference** — removed from the title and description. You're right that the `status` command itself isn't affected; it was pointing at the tree `discover` prints, but agreed it reads as misleading.

**Per-worktree file question** — confirmed: with `useTopLevelMacheteFile=false`, `discover` writes to the per-worktree machete file of the worktree it's run from. The path comes from `__get_git_machete_branch_layout_file_path()`, which returns the current worktree's git dir in that mode (and the main worktree's in the default shared mode); both the read and the save go through that one resolved `_branch_layout_file_path`, so this PR doesn't touch it — the change is confined to the reflog aggregation used for recency ranking. Verified end-to-end: `discover` in the main worktree saved to `.git/machete`, and in a linked worktree to `.git/worktrees/<id>/machete`. And exactly as you anticipated, because ranking now combines all worktrees' reflogs, the linked worktree's file did pick up a branch checked out only in the main worktree — the "too many branches, trim the per-worktree file" case you described, not a regression.

git_machete/git.py
PawelLipski: (okay responded below)

In future, if we deem that discovery should happen per-worktree when `useTopLevelMacheteFile=false`, we can easily conditionally change the logic to the old one. For now, better not introduce too many moving parts, this is complex enough as it is 😅 worktrees (thanks to wide agent adoption...) only opened another cans of worms, thus adding complexity in multiple places

tests/test_discover.py
```diff
@@ -213,3 +217,46 @@ def test_discover_with_merged_branches(self, mocker: MockerFixture) -> None:
...
+            return {line.strip() for line in read_file(machete_file).splitlines() if line.strip()}
+
+        # ...and feat_wt lives in a linked worktree, checked out only there.
+        worktree = add_worktree("feat_wt")
+
+        from_main = discover_here()
+        os.chdir(worktree)
+        from_worktree = discover_here()
+
+        # The two worktrees must agree, and each cross-worktree branch must survive pruning.
```
PawelLipski, line 259: Also, sanity check - without the change to git.py, are the two discovered layouts different? otherwise the test would be of limited value

RELEASE_NOTES.md
```diff
@@ -2,6 +2,8 @@
 
 ## New in git-machete 3.44.2
 
+- fixed: `git machete discover` no longer produces a different branch tree depending on which worktree it is run from; the fresh-branch recency ranking now aggregates HEAD reflogs across all worktrees rather than only the current one (contributed by @earfman)
```
PawelLipski: Actually, pls also include the reporter:


```suggestion
- fixed: `git machete discover` no longer produces a different branch tree depending on which worktree it is run from; the fresh-branch recency ranking now aggregates HEAD reflogs across all worktrees rather than only the current one (reported by @jasonoura, contributed by @earfman)
```

Probably the first time the given issue has a different reporter & implementer, and neither is me (or someone from my team) 😅  are you guys associated in any way?

tests/test_discover.py
```diff
@@ -213,3 +217,46 @@ def test_discover_with_merged_branches(self, mocker: MockerFixture) -> None:
...
+        assert from_main == from_worktree, (from_main, from_worktree)
+        assert "feat_wt" in from_main    # checked out only in the linked worktree
```
PawelLipski, line 261: Yeah, they'd be different, okay

PawelLipski:
Ouch, there'll be no coverage since it's coming from fork, and secrets are not provided by CI to the builds from forks, lemme think how to resolve it...

git_machete/git.py
```diff
@@ -1460,22 +1460,56 @@ def get_git_timespec_parsed_to_unix_timestamp(self, date: str) -> int:
...
+        # Raw reflog line: "<old-sha> <new-sha> <name> <email> <unix-ts> <tz>\t<subject>".
+        # We read the files directly (rather than N `git reflog show` subprocesses, one per worktree) since
+        # `git reflog show` can only report the current worktree's HEAD; the on-disk format is stable.
```
PawelLipski: Actually yeah, if it comes to format stability - is it declared anywhere in git docs?
I mean, tests pass for a very wide range of versions (1.8 to the recent 2.5x ones), so the format is effectively stable, but curious if git claims it explicitly

git_machete/git.py
```diff
@@ -1460,22 +1460,56 @@ def get_git_timespec_parsed_to_unix_timestamp(self, date: str) -> int:
             raise UnexpectedMacheteException(f"Cannot parse timespec: `{date}`")
 
     def get_latest_checkout_timestamps(self) -> Dict[str, int]:  # TODO (#110): default dict with 0
```
PawelLipski: Remove this TODO while you're here, it's stale ;p

git_machete/git.py
```diff
@@ -1460,22 +1460,56 @@ def get_git_timespec_parsed_to_unix_timestamp(self, date: str) -> int:
...
+        worktrees_dir = self.get_main_worktree_git_subpath("worktrees")
+        if os.path.isdir(worktrees_dir):
+            for worktree_id in sorted(os.listdir(worktrees_dir)):
+                reflog_paths.append(self.get_main_worktree_git_subpath("worktrees", worktree_id, "logs", "HEAD"))
+
+        # Raw reflog line: "<old-sha> <new-sha> <name> <email> <unix-ts> <tz>\t<subject>".
+        # We read the files directly (rather than N `git reflog show` subprocesses, one per worktree) since
+        # `git reflog show` can only report the current worktree's HEAD; the on-disk format is stable.
+        pattern = re.compile("^checkout: moving from (.+) to (.+)$")
+        for reflog_path in reflog_paths:
```
PawelLipski: I'd extract reading of reflogs to a lower-level method which would return the reflog entries as some named tuples, and then `get_latest_checkout_timestamps` would just use these entries without dealing with file format

git_machete/git.py
```diff
@@ -1460,22 +1460,56 @@ def get_git_timespec_parsed_to_unix_timestamp(self, date: str) -> int:
...
+                try:
+                    timestamp = int(meta_fields[-2])
+                except (IndexError, ValueError):
+                    continue
+                for branch in (match.group(1), match.group(2)):
+                    # Keep the most recent checkout per branch across *all* worktrees. Comparing timestamps
+                    # (rather than "first entry wins") is order-independent, so it's correct even though the
+                    # reflog files are appended oldest-first - the opposite order to `git reflog show`.
+                    if timestamp > result.get(branch, 0):
+                        result[branch] = timestamp
```
PawelLipski, line 1495: Also, in fact, would it be very difficult to instead of all this parsing, just run `git reflog` in each individual directory, and aggregate the results at the end? This is run once per `discover` (which is a relatively rare operation), and `reflog` is fast compared to e.g. `log` (mostly does what you do here anyway)

git_machete/git.py
PawelLipski: Nah, no, see below

git_machete/git.py
PawelLipski: This way we won't need to take care of parsing raw reflog data, but instead `git reflog` output which is easier to control

git_machete/git.py
PawelLipski: (maybe irrelevant, see comment at the end of file)

earfman:
Thanks @PawelLipski — addressed the review:

**Reflog approach (your suggestion)** — switched from parsing the raw reflog files to running `git reflog show` per worktree via `git -C <dir>` and aggregating the most recent checkout timestamp per branch. It reuses the same version-safe `--format=%gd:%gs --date=raw` the code already relied on, so the on-disk-format-stability question goes away entirely — we're back to reading `git reflog` output, just once per worktree instead of once total. The `git < 2.5` path reads the sole worktree's reflog directly (no `git -C`, which predates worktrees anyway).

**Stale TODO** — removed.

**Release note** — took your suggestion and credited the reporter: `(reported by @jasonoura, contributed by @earfman)`.

**Sanity check on the test** — confirmed: without the git.py change, `discover` from the two worktrees produces different layouts (test fails); with it, they match (test passes). So the test does exercise the fix.

**On the reporter** — no relation; @jasonoura filed #1754 and I picked it up to fix it. First time reporter and implementer are different and unrelated, as you spotted.

git_machete/git.py
```diff
@@ -1459,23 +1459,43 @@ def get_git_timespec_parsed_to_unix_timestamp(self, date: str) -> int:
...
+
+        if self.get_git_version() < WORKTREE_COMMAND:
+            # No linked worktrees (and no `git -C`) on this old-git path: read the sole worktree's reflog directly.
+            collect(self._popen_git("reflog", "show", "--format=%gd:%gs", "--date=raw", allow_non_zero=True).stdout)
+        else:
+            # `git -C <dir>` runs reflog as if invoked from that worktree, so `HEAD` resolves to *that*
+            # worktree's per-worktree HEAD reflog - the only way a single `git reflog` can see another
+            # worktree's checkouts. `load_branch_by_worktree_root_dir` already lists every live worktree
+            # (main + linked, stale ones pruned).
+            for worktree_root_dir in self.load_branch_by_worktree_root_dir():
```
PawelLipski: Since only the directories (keys of the dict) and not branches (optional values of the dict) matter, let's add a helper that would return just worktree paths as a list (unless it already exists)

git_machete/git.py
```diff
@@ -1459,23 +1459,43 @@ def get_git_timespec_parsed_to_unix_timestamp(self, date: str) -> int:
...
+                for branch in (match.group(2), match.group(3)):
+                    # Keep the most recent checkout per branch across *all* worktrees. Comparing timestamps
+                    # (rather than "first entry wins") is order-independent, so aggregating several worktrees'
+                    # reflogs stays correct regardless of the order they're visited.
+                    if timestamp > result.get(branch, 0):
+                        result[branch] = timestamp
+
+        if self.get_git_version() < WORKTREE_COMMAND:
+            # No linked worktrees (and no `git -C`) on this old-git path: read the sole worktree's reflog directly.
+            collect(self._popen_git("reflog", "show", "--format=%gd:%gs", "--date=raw", allow_non_zero=True).stdout)
```
PawelLipski, line 1503: Actually, why `allow_non_zero`? it wasn't there before 🤔

PawelLipski (commented):
Minor comments left, mostly ready to go

earfman:
Both addressed in the latest commit:

**Worktree-paths helper** — extracted `get_worktree_root_dirs()` (paths of all live worktrees as a list — just the keys of `load_branch_by_worktree_root_dir`, for callers that don't care which branch is checked out where) and switched the reflog loop to use it.

**`allow_non_zero`** — deliberate, and now documented with a comment at the call site. `git reflog show` exits with 128 ("your current branch ... does not have any commits yet") against an unborn HEAD — which is exactly the state of a worktree added with `git worktree add --orphan` (or a just-`git init`-ed repo). Since the loop now visits *all* worktrees, a single orphan worktree anywhere would otherwise crash `discover` run from any worktree. Verified live: with a linked `--orphan` worktree present, `reflog show` inside it exits 128, and with `allow_non_zero=True` `discover` completes normally, treating that worktree as contributing no checkouts. (The pre-PR code got away without it only because it never looked beyond the current worktree.)

mypy, flake8 and the discover suite are green locally.

git_machete/git.py
PawelLipski: (explained in a comment, thanks)

PawelLipski (approved):

earfman:
Thanks for merging, and for the four rounds — the reflog approach and the worktree-paths helper were both your calls and both made the patch better.

One thing I'd value your read on, if you have a minute. I should say plainly: this patch was written with heavy AI assistance, run through a verification step of my own before it went out, and then corrected four times by you. I'm building tooling around that last part — how a maintainer separates a well-formed pull request from a merely plausible-looking one — and you're one of very few people I've watched do it carefully on a real project.

So: are you getting many machine-written PRs on git-machete yet, and when one shows up, what do you actually look at first to decide whether it's worth your review time?

Nothing to sell. I'd rather learn this from someone doing it than guess.

tests/test_discover.py
jasonoura: Can confirm I don't know @earfman

PawelLipski:
> are you getting many machine-written PRs on git-machete yet

I've got a few (<10, probably), and of course I've already merged probably close to 100 my own AI-generated PRs 😅
Generally it was pretty easy so far to separate chaff from the wheat - the former were typically PRs for old issues marked with _good first issue_. While these issues exist precisely to be picked up by 3rd party contributors, in many cases the solution missed the point completely, and/or there was no further response from the author (they just hoped for an instant merge?).

PRs fixing the bugs/inconsistencies (like this one) tend to be higher quality (and contributors more responsive) than the code quality/good first issue ones 😅

PR #1766, opened 2026-08-31
feat: add `-o`/`--push-option` flag to `traverse`

HWiese1980:
Forward one or more `--push-option` values to every `git push` run
during a `traverse`, e.g. `git machete traverse --push-option=ci.skip`.
The flag is repeatable, since push-option values routinely contain `=`
and `,` themselves (`ci.variable=SKIPPED_SUITES=unit,e2e`), which a
single comma-delimited flag couldn't carry.
Requires git >= 2.10.0 and a remote with `receive.advertisePushOptions`
enabled; otherwise `git push` fails outright.

`ParsedCmd` gains a `repeated_opts: Dict[str, List[str]]` field
alongside the existing `opts: Dict[str, str]`, since a repeatable
option can't be represented by the latter without breaking every
existing reader of `opts`. `OptSpec.multiple` marks which options are
collected there instead.

Also: bump version to 3.45.0, update RELEASE_NOTES, docs, man page,
bash/zsh/fish completions, and the `git-machete` agent skill.

## Test plan

- [x] New `tests/test_traverse.py::test_traverse_push_options`: drives real pushes through a `pre-receive` hook on a bare remote and asserts the exact `GIT_PUSH_OPTION_*` environment the receiving end sees, across three traversals (two options, one option, none). Gated by `@pytest.mark.skipif(get_git_version() < PUSH_OPTIONS, ...)`.
- [x] Full suite: 482 passed, 1 skipped.
- [x] `tox -e flake8-check,mypy,isort-check,typos-check,vulture-check,cyclic-import-check,py-docs-check,sphinx-man-check` all green.
- [x] `ci/checks/run-all-checks.sh` (all 47 checks) passes.
- [x] Shell completion E2E: bash 113/113, zsh 112/113 (one unrelated test timing out per run, consistent with the known macOS zsh flakiness).
- [ ] fish completion E2E not verified locally (fish isn't installed on this machine); covered by CI.


PawelLipski:
Hey, to better understand your use case - is it specifically `ci.skip` that's needed for you? or did you also find yourself in need of using other push options?

Also, do these options tend to be more "global" in the sense of applying to all invocations within the given timestamp (like, you want to pass the same push-option to every `git machete traverse` in the given day)? Maybe a git config key would be more adequate then?

HWiese1980:
In my particular case it was indeed `ci.skip` primarily, giving the impulse to implement this feature. I'm sure I'm going to need other flags too at some point in my workflow. I have machete in my partially AI based workflow too. Giving me, as well as the AI agents the capability to set/omit certain push options during an entire machete run would be extremely useful. I also considered configuring the flags dynamically through temporary static config changes (modifying the defaults). It is indeed another way to achieve this behavior. However, it's way more cumbersome and also error-prone.

PawelLipski:
Hmmm actually - AFAICS there's already `push.pushOption` git config key, which affects the underlying `git push` calls from `git machete traverse` (and also `advance`, `github` etc.).

Still, to avoid overloading CLI (which is already pretty bloated in case of `traverse`), and also for consistency with  `GIT_MACHETE_REBASE_OPTS` (which passes-through flags to the underlying `git rebase` invocations):

How about adding this feature as `GIT_MACHETE_PUSH_OPTS` instead? and to make it more robust, let it accept **any** push flags (not just `--push-options`)?

This is also easier to set on a per-command basis with `GIT_MACHETE_PUSH_OPTS='-o ci.skip' git machete traverse`, as compared to git config keys

HWiese1980:
Fair point, and a good alternative approach. I just wanted to avoid the modify git config -> traverse -> revert git config detour, and your approach with GIT_MACHETE_PUSH_OPTS would serve this purpose well enough. Cool! I'm gonna implement it and create a new PR.

PR #1785, opened 2026-09-18
Release v3.46.0

PawelLipski:


docs/source/cli/config.rst
PawelLipski, line 10: Whoops docs on `advice.*` keys missing

